PR #1, opened 2024-01-17
Initial project setup

evansd:
This creates a new skeleton Django project (without an app) and sets up the basic infrastructure around dependency management and testing.

This is cribbed in large part from the repo-template, but I want to make sure that:
 1. everything here is stuff we actually want/need, not just cargo-culted;
 2. I'm happy with the way all the various pieces work.

Of course I want you, dear reader, to be happy too: hence this PR.


.python-version
```diff
@@ -0,0 +1 @@
+3.11.2
```
madwort, line 1: was the upshot from this morning's discussion that this should now be `3.11`?

justfile
```diff
@@ -0,0 +1,147 @@
...
+    $BIN/python manage.py {{ ARGS }}
+
+
+# run tests
+test *ARGS: devenv
+    $BIN/python -m pytest {{ ARGS }}
+
+
+# run tests as they will be in run CI (checking code coverage etc)
+@test-all: devenv
```
bloodearnest, line 139: Huh, TIL about `@command` Never had need of it before.

I think I may be less sensitive to excess verbiage that others.

tests/test_python_version.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+import re
+from pathlib import Path
+
+import pytest
+
+
+BASE_DIR = Path(__file__).parents[1]
+
+PYTHON_VERSION = "3.11"
```
madwort, line 13: what if we made `.python-version` the canonical version, and then we reuse this script across a few different repos?

madwort: (I mean actually reuse, i.e. put the script in a shared place)

bloodearnest (approved):
Regards venv tooling, you've chosen to implement things differently than our other projects in a few ways.

Whilst I think there's some good ideas in your approach, I think in needs some wider discussion.

Am giving my +1, as I don't want to block the 1st PR, but it would be good to have some discussion as to why you've change things.

tests/test_python_version.py
evansd: Ah yeah, making `.python-version` canonical is a good idea. I'm not sure we've got a good mechanism for sharing this kind of thing yet though. And I suspect the details of which files need to be checked and how will end up being quite project specific. I'm definitely in favour of spreading the pattern around though.

justfile
evansd: This _creates_ more verbiage though, doesn't it? In that it echos the script contents before running it. I didn't come up with the idea (it's borrowed from the ehrQL justfile) but I think it's nice to have the script echoed in the CI output.

dotenv-sample
```diff
@@ -1 +1,5 @@
+# Default environment variables for local development
 
+DJANGO_DEBUG=True
+
+DJANGO_SECRET_KEY="django-insecure-q^d8$ak3811$r^b*@i4tqld6c@5co5in*l=9h@8c!nh4!c$v@l"
```
inglesp, line 5: I have a mild preference for something even more obviously insecure here.

justfile
madwort: an `@` in this position stops just from outputting the commandline just before running it

```
test-all:
  @echo "thing"
```

you're saying that an `@` in this position outputs the whole action before running?

```
@test-all:
  echo "thing"
```

tests/test_python_version.py
evansd: Have updated to use `.python-version` as the source-of-truth.

inglesp (approved):

justfile
madwort: ah yes https://github.com/casey/just#quiet-recipes

tests/test_python_version.py
madwort: thanks

justfile
evansd: Yeah, it's _super_ confusing. A `@` prefix on the recipe makes it "quiet" which inverts the usual echoing behaviour so lines which would normally be echoed are suppressed and lines prefixed with `@` are echoed.

But the default behaviour for shebang recipes is _not_ to echo anything. So by making it "quiet" it ends up echoing the script. Easy, right? :joy: 
https://github.com/casey/just?tab=readme-ov-file#quiet-recipes

dotenv-sample
evansd: Fine yes, this was just the default Django generated.

justfile
bloodearnest: Huh, I totally read that wrong!

bloodearnest:
> @bloodearnest In terms of the venv management, this is basically implementing the ideas we discussed (and I think generally agreed on) here but never got round to doing: https://bennettoxford.slack.com/archives/C63UXGB8E/p1693556976925389

Ah, yes, thanks for the context

Not compiling does introduce a small risk that you add something to requirements.in and forget to run to compile. But this is low impact risk, as its effectively a noop.

But I think it would be good to implement this change in job-server/repo-template et all also, as the goal is the same tooling behaviour.  Might be a good use case for just modules, i.e. a `venv` module, so you do `just venv devenv` or `just venv compile`, and the module, being a separate file, can be more easily kept in sync across projects. 

 
> There are a few other "innovations" in there which I'd be happy to discuss, but they're mostly small I think.

Yes, I think those are good improvements. I think they should work if we add a parallel prod requirements file.

evansd:
> But I think it would be good to implement this change in job-server/repo-template et al also, as the goal is consistent tooling behaviour.

Yeah, that would be good. But I think it's worth letting this stuff bed in for a bit. We've already got plenty of (inevitable) drift between projects so I think it's OK to diverge for a bit. But yes, we should be feeding this stuff back upstream and then out to other projects when we're confident in it – and have the time! 

>  I think they should work if we add a parallel prod requirements file.

So, there is already a `requirements.prod.txt` file (which `just devenv` installs alongside the dev requirements) but there's no equivalent `just prodenv` command. The reason for that is (as I understand it) that building the production environment is the responsibility of the Docker file, not of Just.

bloodearnest:
> > I think they should work if we add a parallel prod requirements file.
> 
> So, there is already a `requirements.prod.txt` file (which `just devenv` installs alongside the dev requirements) but there's no equivalent `just prodenv` command. The reason for that is (as I understand it) that building the production environment is the responsibility of the Docker file, not of Just.


Ah, I missed the bash loop :facepalm: It makes me wonder if we only needed `prodenv` to manage the automatic compiling. 

Ideally, I would prefer for a `just prodenv` to be the command that the Dockerfile uses to build the prodenv, for DRY, but in practice that's awkward, and not what we do elsewhere.

However, there is a different potential issue here, related to using pip-compile with separate dev/prod deps generally.

The prod and dev deps may share transitive dependencies. But because they're pip-compiled independently of each other, that can lead to version conflicts when installing both.

We fix this in job-server by adding `--constraint requirements.prod.txt` to the top of `requirements.dev.in`, which should ensure that prod versions are respected when managing dev versions.

evansd:
> We fix this in job-server by adding `--constraint requirements.prod.txt` to the top of `requirements.dev.in`, which should ensure that prod versions are respected when managing dev versions.

Ah, nifty. I shall implement this now!

PR #16, opened 2024-01-25
Add a very basic file browser app

evansd:
Most of this should be pretty obvious. A few high-level notes:

 * You'll see I've used type annotations in a few places, but not consistently. This is a practice we developed in ehrQL where we abandoned the attempt at full type checking (either it wasn't ready for something like ehrQL, or we weren't ready for it), but still found ourselves wanting to mention types in comments where they weren't immediately clear. So we ended up using type annotations as a form of terse, standardised comment. And possibly this will inch us towards a world where we do end up using types for real.

 * My "unit" tests touch the filesystem, which will cause the purists to retch in horror. I'm taking the ehrQL stance that "unit test" just means "fast and focused" and not worrying too much about the precise semantics.

 * I've resorted to inline styles in a couple of places. It turns out that the way Tailwind works is quite different from other CSS frameworks, and the only classes we have defined in our CSS are the ones Job Server happens to use. (Note, this is not dead code elimination: the classes are constructed dynamically when they're referenced so it's not like there's some "full Tailwind" stylesheet we can include.) There's a broader question we'll need to tackle about this, but for now I've added things inline.

### DIFF
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -159,8 +159,10 @@ USE_TZ = True
 
 STATIC_URL = "static/"
 
-ASSETS_DIST = Path(os.environ.get("ASSETS_DIST", BASE_DIR / "assets/dist"))
-STATICFILES_DIRS = [str(ASSETS_DIST)]
+ASSETS_DIST = BASE_DIR / "assets/dist"
+STATICFILES_DIRS = [
+    ASSETS_DIST,
+]
 
 # Serve files from static dirs directly. This removes the need to run collectstatic
 # https://whitenoise.readthedocs.io/en/latest/django.html#WHITENOISE_USE_FINDERS
@@ -169,7 +171,7 @@ WHITENOISE_USE_FINDERS = True
 DJANGO_VITE = {
     "default": {
         # vite assumes collectstatic, so tell it where the manifest is directly
-        "manifest_path": "assets/dist/.vite/manifest.json",
+        "manifest_path": ASSETS_DIST / ".vite/manifest.json",
     },
 }
 
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -180,3 +180,7 @@ DJANGO_VITE = {
 # https://docs.djangoproject.com/en/5.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+# In production we'd expect AIRLOCK_WORKSPACE_DIR to be an absolute path pointing
+# somewhere outside of WORK_DIR
+WORKSPACE_DIR = WORK_DIR / get_env_var("AIRLOCK_WORKSPACE_DIR")
--- a/airlock/workspace_api.py
+++ b/airlock/workspace_api.py
@@ -1,0 +1,65 @@
+import dataclasses
+import pathlib
+
+from django.conf import settings
+from django.urls import reverse
+
+
+@dataclasses.dataclass(frozen=True)
+class PathItem:
+    """
+    This provides a thin abstraction over `pathlib.Path` objects with two goals:
+
+        1. Paths should be enforced as being relative to a certain "container" directory
+           and it should not be possible to traverse outside of this directory or to
+           construct one which points outside this directory (using the designated
+           constructor classmethods).
+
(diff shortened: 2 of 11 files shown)


airlock/settings.py
```diff
@@ -159,8 +159,10 @@ def get_env_var(name):
 
 STATIC_URL = "static/"
 
-ASSETS_DIST = Path(os.environ.get("ASSETS_DIST", BASE_DIR / "assets/dist"))
-STATICFILES_DIRS = [str(ASSETS_DIST)]
+ASSETS_DIST = BASE_DIR / "assets/dist"
```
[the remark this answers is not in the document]
evansd: Yeah, I checked with him about this:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1706006312272769

It turns out we don't need it here because we're trying not to have any build step at all.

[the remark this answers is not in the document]
bloodearnest: Yes, but we have no npm build pipeline in airlock (atm) - we import prebuilt static assets from jobs server. So there's no need to configure this at the moment. We can always add it back if needed.

airlock/settings.py
```diff
@@ -178,3 +180,7 @@ def get_env_var(name):
 # https://docs.djangoproject.com/en/5.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+# In production we'd expect AIRLOCK_WORKSPACE_DIR to be an absolute path pointing
+# somewhere outside of WORK_DIR
+WORKSPACE_DIR = WORK_DIR / get_env_var("AIRLOCK_WORKSPACE_DIR")
```
bloodearnest, line 186: Yeah, this confused me when testing locally.

I have a mild preference to have the default WORKSPACE_DIR to not be in WORK_DIR, but not a big issue.

airlock/workspace_api.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def url(self):
+        suffix = "/" if self.is_directory() else ""
+        return reverse("file_browser", kwargs={"path": f"{self.relpath}{suffix}"})
+
+    def parent(self):
+        if self.relpath.parents:
+            return PathItem(self.relpath.parent)
+
+    def children(self):
```
inglesp, line 40: Makes me slightly uneasy to see this on something that might represent a file.

airlock/workspace_api.py
```diff
@@ -0,0 +1,53 @@
+import dataclasses
+import pathlib
+
+from django.conf import settings
+from django.urls import reverse
+
+
+@dataclasses.dataclass(frozen=True)
+class PathItem:
```
inglesp, line 9: It'd be nice to have a docstring explaining what this is for and why it's worth wrapping `pathlib.Path`.

inglesp (approved):

bloodearnest (approved):

airlock/workspace_api.py
evansd: Yeah, I felt similarly. And also having a `contents()` method on a class which might represent a directory. But having an abstract base class with two subclasses felt a bit heavy-handedly OO (at least at this stage). And I couldn't think of what concrete advantages it would give us given that we're not actually type checking. Obviously both methods raise errors if you call them on the wrong sort of thing, which in dynamic terms isn't much different from them just not being there.

airlock/settings.py
evansd: Yeah I do get that line of thinking. My idea was that locally _all_ the state should get written to `workdir` so it's in one place. Bear in mind these are development settings rather than defaults. You have to set all these explicitly in production.

I'd be open to changing it if it does turn to be confusing though.

airlock/workspace_api.py
inglesp: Fine.

assets/views.py
```diff
@@ -3,7 +3,7 @@
 from django.template.response import TemplateResponse
 
 
-def components(request):
+def components(request):  # pragma: no cover
```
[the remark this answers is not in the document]
evansd: We could, but that really would feel to me like chasing coverage for the sake of coverage given that this is just a demo page Simon threw up.

PR #31, opened 2024-01-30
Refactor PathItem to have plugable root directory.

bloodearnest:
This way, it can be used for viewing both workspaces and output
requests. A PathItem now takes a reference to container object, which
must have a root path, and a way to generate urls.

Added a top level Workspace object, as the first implementation of such
a Container, will add one for an OutputRequest later.

A workspace is now explict, rather than embedded in that path. As such,
I took the opportunity to explicitly add it as url parameter. The
PathItem.relpath is now always relative to the workspace directory, not
WORKSPACE_DIR.

In anticipation of adding a request file browser view, I renamed the
urls/views to be "workspace".

### DIFF
--- a/airlock/views.py
+++ b/airlock/views.py
@@ -2,15 +2,20 @@ from django.http import Http404
 from django.shortcuts import redirect
 from django.template.response import TemplateResponse
 
-from airlock.workspace_api import PathItem
+from airlock.workspace_api import Workspace
 
 
 def index(request):
     return TemplateResponse(request, "index.html")
 
 
-def file_browser(request, path: str = ""):
-    path_item = PathItem.from_relative_path(path)
+def workspace_view(request, workspace_name: str, path: str = ""):
+    workspace = Workspace(workspace_name)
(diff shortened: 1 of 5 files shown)


airlock/views.py
```diff
@@ -2,15 +2,20 @@
...
     return TemplateResponse(request, "index.html")
 
 
-def file_browser(request, path: str = ""):
-    path_item = PathItem.from_relative_path(path)
+def workspace_view(request, workspace_name: str, path: str = ""):
+    workspace = Workspace(workspace_name)
+    # TODO workspace authorization
+    if not workspace.exists():
+        raise Http404()
```
evansd, line 16: Coverage is complaining about this line.

evansd (approved):
Looks good. Having path objects hold a reference to a container object was exactly how I imagined we'd generalise this so nice to see it all seems to work out.

airlock/views.py
bloodearnest: Ah, thanks

PR #32, opened 2024-01-30
Support token based login

### LINKED ISSUE #19: Add login page
Users can generate a token in job-server that can be used for authentication for a limited amount of time.

We need a login page that:

* lets a user enter their token
* sends the token to the job-server endpoint under development in opensafely-core/job-server#4025
* if successful, populates the user's session with data from the endpoint, and redirects the user to the homepage
* if unsuccessful, displays a helpful error message

The session will need to contain:

* whether the user is an output checker
* the list of workspaces that the user has access to

evansd:
I've confirmed that this works against the real API using a real token, but the tests uses mocked `requests` objects for obvious reasons.

For working locally we support a `dev_users.json` file giving a static set of users and token with which to log in. A basic users file is included in the example data.

The existing code expects users to have an `id` attribute but the API doesn't return this. I've worked around this for now but we should decide one way or the other. 

Closes #19


airlock/login_api.py
```diff
@@ -0,0 +1,42 @@
...
+        return get_user_data_prod(user, token)
+
+
+def get_user_data_prod(user: str, token: str):
+    response = requests.post(
+        f"{settings.AIRLOCK_API_ENDPOINT}/releases/auth",
+        headers={"Authorization": settings.AIRLOCK_API_TOKEN},
+        json={"user": user, "token": token},
+    )
+    # We don't currently get any more detail about failures than this
```
bloodearnest, line 24: Hmm, yes, we could expose more info here, given its an authenticated by backend token.

I'm not sure the original decision to avoid exposing extra information was a good one any way, given the costs.

bloodearnest (approved):
This looks good to me.

I'm inclined to remove `User.id`. I'm hoping we won't need a local user db table.

I also think we should probably expose more granular auth failure reasons in the job-server API, to aid debugging.

PR #33, opened 2024-01-30
basic request view

bloodearnest:
- Add basic request file browser UI
- Basic UI buttons, and tweaks


bloodearnest:
Hmm, I get 100% coverage locally

madwort:
fwiw I get the same coverage as CI if I run it locally (96% of `airlock/views.py`)

bloodearnest:
> fwiw I get the same coverage as CI if I run it locally (96% of `airlock/views.py`)

Yeah, it seems non-deterministic, give its a) not always happened in CI b) it some times reports different statements as uncovered.

I sanity check that the lines reported missing are actually executed in tests.

airlock/templates/file_browser/index.html
```diff
@@ -1,9 +1,34 @@
...
+    <input type=hidden name="workspace" value="{{ workspace.name }}"/>
+    <input type=hidden name="output_request" value="{{ output_request.request_id }}"/>
+    {% #button type="submit" tooltip="This output request is ready to be reviewed" variant="success" %}Submit Output Request For Review{% /button %}
+    </form>
+  {% elif is_output_checker %}
+    <form action="" method="POST">
+    {% csrf_token %}
+    <input type=hidden name="workspace" value="{{ workspace.name }}"/>
+    <input type=hidden name="output_request" value="{{ output_request.request_id }}"/>
+    {% #button type="submit" tooltip="Release these to jobs.opensafely.org" variant="danger" %}Release Files{% /button %}
```
rebkwok, line 18: nit (for later): hovering over the tooltip makes a scrollbar appear, and the page shift to the right which is a bit weird

airlock/workspace_api.py
```diff
@@ -63,6 +63,30 @@ def get_url(self, relpath):
         )
 
 
+@dataclasses.dataclass(frozen=True)
+class OutputRequest(Container):
```
rebkwok, line 67: Is it an OutputRequest or a ReleaseRequest?  I may just be still slightly confused about terminology

justfile
```diff
@@ -176,3 +176,14 @@ load-example-data: devenv
       curl -s https://jobs.opensafely.org/opensafely-internal/tpp-database-schema/outputs/85/download/ --output "$tmp"
       unzip "$tmp" -d "$workspace"
     done
+
+    mkdir -p "$workspace/sub_dir_empty"
+
+    tmp=$(mktemp)
+    # grab published database outputs for example csv and html data
+    curl -s https://jobs.opensafely.org/opensafely-internal/tpp-database-schema/outputs/85/download/ --output "$tmp"
+    unzip -u "$tmp" -d "$workspace"
```
rebkwok, line 185: I think something's got mangled in the rebase here.  This bit is duplicated in the loop above - it should be in one or the other. Probably we don't need these files and the request duplicated in every workspace, so outside of the loop is enough for sample data

rebkwok (changes requested):
Mostly looks good; couple of minor changes needed

tests/integration/test_views.py
```diff
@@ -133,3 +142,72 @@ def test_workspaces_index_user_permitted_workspaces(
...
+    response = client_with_permission.get(f"/requests/{workspace_name}/{request_id}/")
+    assert "file.txt" in response.rendered_content
+
+
+def test_request_workspace_does_not_exist(client_with_permission):
+    response = client_with_permission.get(f"/requests/bad/{request_id}/")
+    assert response.status_code == 404
+
+
+def test_request_id_does_not_exist(client_with_permission):
```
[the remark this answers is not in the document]
bloodearnest: great catch, thanks

airlock/templates/file_browser/index.html
bloodearnest: interesting. One for @tomodwyer input, perhaps

airlock/workspace_api.py
bloodearnest: Fair point.  I used OutputRequest because ReleaseRequest sounds worse to say out load :shrug:

justfile
bloodearnest: arg, yes, thanks

airlock/workspace_api.py
bloodearnest: I have renamed.

rebkwok (approved):

airlock/templates/file_browser/index.html
tomodwyer: Yeah this is a really annoying bug, the only solution I've found is to add custom classes to move the tooltip in a different direction 😢

PR #37, opened 2024-01-31
Add a logout view

### LINKED ISSUE #20: Add logout page
For the time being, we store information about whether a user in their session, which gets set when they log in.

If a user's workspaces are updated, or if they become (or stop being) an output checker, the session becomes out of sync.

We should add a logout page so that users can log out and then log back in again, to pick up the latest data from job-server.

rebkwok:
Fixes #20 
- Adds a tiny middleware to add user to the request
- Adds username and is_authenticated to User for use in templates
- A minimal temporary login url/view so that we can login as an output-checker, in absence of actual login flow
- Logout view, with associated template buttons

### DIFF
--- a/airlock/middleware.py
+++ b/airlock/middleware.py
@@ -1,0 +1,12 @@
+from airlock.users import User
+
+
+def user_middleware(get_response):
+    """Add the session user to the request"""
+
+    def middleware(request):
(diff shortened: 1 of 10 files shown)


bloodearnest (approved):
great, it's all starting to come together!

PR #42, opened 2024-02-01
chore: refactor yaml

madwort:
* fix the permissions issue with workdir while we're here

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
         run: |
           just test-all
 
-  docker-test:
+  docker-build:
     runs-on: ubuntu-22.04
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -53,29 +53,89 @@ jobs:
...
+            path: /tmp/image
+
+      - name: Import docker image
+        run: gunzip -c /tmp/image/airlock.tar.gz | docker load
+
+      - name: create workdir
+        run: |
+            sudo touch workdir/db.sqlite3
+            sudo chmod -R ugo+rw workdir/
+            ls -lh workdir/db.sqlite3
```
madwort, line 128: previously docker/smoke-test was passing in CI without this, but only because `docker/test` was touching these files with different permissions.

madwort: nb. I don't love this, open to alternative suggestions

.github/workflows/main.yml
```diff
@@ -53,29 +53,90 @@ jobs:
...
+
+      - name: Import docker image
+        run: gunzip -c /tmp/image/airlock.tar.gz | docker load
+
+      - name: create workdir
+        run: |
+            # the prod container is forcibly running as user 10000
+            # so make this world-readable for now so we can unblock next steps
+            sudo touch workdir/db.sqlite3
+            sudo chmod -R ugo+rw workdir/
```
madwort, line 129: previously docker/smoke-test was passing in CI without this, but only because docker/test was touching these files with different permissions.
nb. I don't love this, open to alternative suggestions

madwort:
closing in favour of #43

PR #52, opened 2024-02-02
Universal login_requires, with exemptions.

bloodearnest:
Now, all views will redirect if the user is not logged in, except for
login and index (at least for now)

We are not using django auth currently, so this is a DIY version, but
it's pretty simple.

it removes the risk of us forgetting to protect a view.


evansd (approved):
Nice

PR #53, opened 2024-02-02
Fix missed reactor because to PRs merged in parallel

bloodearnest:
Stupid gihub.

### DIFF
--- a/airlock/services/__init__.py
+++ b/airlock/services/__init__.py
@@ -1,0 +1,4 @@
+import svcs
+
+
+registry = svcs.Registry()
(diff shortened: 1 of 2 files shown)


airlock/services/__init__.py
```diff
@@ -0,0 +1,4 @@
+import svcs
+
+
+registry = svcs.Registry()
```
evansd, line 4: What's this?

evansd (approved):
Assuming you mean:
> Fix missed refactor because two PRs merged in parallel
> Stupid github

Then looks good!

PR #55, opened 2024-02-02
Add a functional tests helper for a logged in page

rebkwok:
We don't need to test the full login flow again, but we will need logged in users for future functional tests. This adds a helper fixture, `logged_in_page`, to handle the logging in and setting expected session data for a specific user.

The workspace_index_page tests don't test anything special, functional- test wise, they just test that the fixture works to load a page as an expected logged in user.


tests/functional/conftest.py
```diff
@@ -25,3 +26,56 @@ def playwright_install(request):
...
+            "details": {
+                "username": "test_researcher",
+                "fullname": "Researcher",
+                "output_checker": False,
+                "staff": False,
+                "workspaces": ["test-dir1"],
+            },
+        },
+    }
+    with open(settings.AIRLOCK_DEV_USERS_FILE, "w") as outf:
```
bloodearnest, line 56:
```suggestion
    settings.AIRLOCK_DEV_USERS_FILE.write_text(json.dump(user_data))
```

bloodearnest (approved):

evansd:
This feels a bit heavyweight to me as a way of getting an authenticated user. We already have a functional test which exercises the full login flow so it's not like we need to make sure the views work here. I wonder if it would be simpler to create a session with the required user and set the cookie on the Playwright client?

rebkwok:
> This feels a bit heavyweight to me as a way of getting an authenticated user. We already have a functional test which exercises the full login flow so it's not like we need to make sure the views work here. I wonder if it would be simpler to create a session with the required user and set the cookie on the Playwright client?

I did try something like that first. I can have another look at it.

rebkwok:
Closed in favour of #59

PR #59, opened 2024-02-05
Add functional tests fixtures for a logged in users

rebkwok:
We don't need to test the full login flow again, but we will need logged in users for future functional tests. This adds a helper fixture, `login_as_user`, and 2 fixtures using it to mock an output-checker and a researcher user, to create session data and set expected cookies in the browser context.

The workspace_index_page tests don't test anything special, functional- test wise, they just test that the fixture works to load a page as an expected logged in user.


tests/functional/conftest.py
```diff
@@ -25,3 +27,53 @@ def playwright_install(request):
...
+        "workspaces": [],
+        "output_checker": True,
+    },
+    "test_researcher": {
+        "id": "test_researcher",
+        "username": "test_researcher",
+        "workspaces": ["test-dir1"],
+        "output_checker": False,
+    },
+}
```
evansd, line 45: Minor stylistic thing, but given that you've got separate `output_checker_user` and `researcher_user` fixtures which wrap `login_as_user` you could lose this global config entirely and have each of those fixtures pass the relevant config e.g:
```py
@pytest.fixture
def output_checker_user(login_as_user):
    yield login_as_user(
        {
            "id": "test_output_checker",
            "username": "test_output_checker",
            "workspaces": [],
            "output_checker": True,
        }
    )
```

It feels like a unnecessary layer of indirection otherwise.

evansd (approved):
Looks great. I've made one (very ignorable) suggestion.

tests/functional/conftest.py
evansd: Yeah, I could see either approach working in different contexts. Feel free to do it however you think makes most sense.

PR #60, opened 2024-02-06
release files

bloodearnest:
- Add client for existing releases API
- Add release_files function to workspace_api
- Add view for releasing files.

I have manually tested this against a local job-server.  I have some tooling to automate setting that up, which I will add in another PR.


old_api/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+from datetime import datetime, timezone
+
+import requests
+from django.conf import settings
+
+from old_api.schema import FileList, FileMetadata
+
+
+session = requests.Session()
+session.headers["Authorization"] = settings.AIRLOCK_API_TOKEN
```
evansd, line 11: I've found that setting things like this at import time can make testing awkward sometimes. Might not be an issue if this is short-lived code though.

old_api/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        stat = abspath.stat()
+        mtime = datetime.fromtimestamp(stat.st_mtime, tz=timezone.utc).isoformat()
+        files.append(
+            FileMetadata(
+                name=str(relpath),
+                size=stat.st_size,
+                sha256=hashlib.sha256(abspath.read_bytes()).hexdigest(),
+                date=mtime,
+                url=str(relpath),  # not needed, but has to be set
+                metadata={},  # not needed *yet*, but can't be None
```
evansd, line 29: `osrelease` sets `tool` in the metadata for the file as well, and that's what my dashboard SQL currently relies on. I think it would be quite reasonable to omit it, but I should update the SQL to join on the Release object if so.

old_api/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+from django.conf import settings
+
+from old_api.schema import FileList, FileMetadata
+
+
+session = requests.Session()
+session.headers["Authorization"] = settings.AIRLOCK_API_TOKEN
+
+
+def create_filelist(request):
```
evansd, line 14: This is maybe worth a comment or a type annotation just because `request` is such an overloaded term here.

evansd (approved):

old_api/__init__.py
bloodearnest: That's a good point, I'll move it to the two places its used.

old_api/__init__.py
bloodearnest: Yep, I'll call it release_request

PR #63, opened 2024-02-07
chore: split docker dev/prod tests into two GHA jobs

madwort:
* curious as to whether this will be faster overall


evansd (approved):
This seems sensible to me.

PR #65, opened 2024-02-07
local job server for local people

### LINKED ISSUE #66: Tooling for automated local job-server for integration testing.
We need to away to easily run a local configured job-server instance, and point airlock to it, so that we can easily integration test against it.

bloodearnest:
- Add tooling to run a local instance of job-server for testing.
- Better debugging of job-server API failures.

Fixes #66


DEVELOPERS.md
```diff
@@ -98,3 +98,66 @@ Each category lives in its own directory (for example `tests/unit`) and can be r
...
+# Local job-server for integration.
+
+## First time set up
+
+This needs some first time setup, but after that is fairly simple to use. You
+will need the Bitwarden cli `bw` installed to pull the dev Github auth
+credentials. You need to run the following command with your github username.
+
+```
+just job-server/configure GHUSERNAME
```
rebkwok, line 112: In addition to installing the bitwarden client, you also need to be logged in, otherwise `configure` asks for and accepts the master password but fails and says you're not logged in (but it's not entirely clear what you're not logged in to).

An extra note to also make sure you're logged in should be enough, I think? I guess the script could check if you're not already logged in and run this for you
```bw login --sso```

File: DEVELOPERS.md [hunk 2]
```diff
@@ -98,3 +98,66 @@ Each category lives in its own directory (for example `tests/unit`) and can be r
...
+just job-server/run
+```
+
+## Create workspace
+
+You will need at least one workspace set up in job-server and locally in airlock to test integration:
+
+
+```
+just job-server/create-workspace NAME  # defaults to "airlock-test-workspace"
```
rebkwok, line 134: I ran this with the default, so it created me an `airlock-test-workspace` on job-server, but then I don't have one on airlock because the example data setup creates workspaces called `example-workspace`, `example-workspace_1` etc. Would it be worth checking the airlock workspaces dir and also creating an empty workspace folder if there isn't a matching one already?

DEVELOPERS.md
```diff
@@ -98,3 +98,66 @@ Each category lives in its own directory (for example `tests/unit`) and can be r
...
+
+By default, the local job-server maintains db and file on a couple of volumes.
+To reset back to a clean slate, you can kill and re-configure, and then add
+workspaces again.
+
+```
+just job-server/clean
+just job-server/configure GHUSERNAME
+just job-server/create-workspace 
+```
```
rebkwok, line 163: I think it would be worth adding the steps you put in the [slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1707327771069959?thread_ts=1707327243.804019&cid=C069YDR4NCA) to the docs

job-server/justfile
```diff
@@ -0,0 +1,86 @@
...
+    error() { 
+        echo "${1:-}"
+        echo "You probably to run just job-runner/configure"
+        exit 1
+    }
+
+    test -f .env.jobserver || error
+
+    . .env.jobserver
+    test "SOCIAL_AUTH_GITHUB_KEY" != "test" || error "SOCIAL_AUTH_GITHUB_KEY not set properly"
```
rebkwok, line 18: Can we make this check that it's set non-empty, as well as not "test"? In my first failed attempt, the `.env.jobserver` file was created and had `SOCIAL_AUTH_GITHUB_SECRET=` in it, so it didn't ask me for the bw password again.

DEVELOPERS.md
```diff
@@ -98,3 +98,66 @@ Each category lives in its own directory (for example `tests/unit`) and can be r
...
+
+```
+just job-server/remove-releases workspace
+```
+
+## Undoing
+
+To use the dev logins, comment out the `AIRLOCK_API_*` lines in .env. To point
+to a different job-server instance, edit the values in .env. To reset it to
+point to local job-server instance, you can re-run the setup command above.
```
rebkwok, line 150:
```suggestion
point to local job-server instance, you can re-run the configure command above.
```

job-server/justfile
bloodearnest: Ah, yes, that's a bit brittle, will do

DEVELOPERS.md
bloodearnest: done

DEVELOPERS.md
bloodearnest: Hmm, it should have created the airlock directory, will double check.

rebkwok (approved):
Nice! After the initial bitwarden login issues, everything ran nicely.
A couple of suggestions, nothing essential

PR #67, opened 2024-02-08
Add shellcheck to linting

bloodearnest:


justfile
```diff
@@ -103,6 +103,7 @@ check: devenv
     check "$BIN/ruff format --diff --quiet ."
     check "$BIN/ruff check --show-source ."
     check "docker run --rm -i ghcr.io/hadolint/hadolint:v2.12.0-alpine < docker/Dockerfile"
+    check "find job-server/ docker/ airlock/ -name \*.sh -print0 | xargs -0 shellcheck"
```
madwort, line 106:
```
=> find job-server/ docker/ airlock/ -name \*.sh -print0 | xargs -0 shellcheck
find: ‘job-server/’: No such file or directory
```
this is fine?

evansd: This also fails for me because I don't have `shellcheck` installed. It seems to work OK in CI, but I think the dev setup should be responsible for installing the correct versions of all the tooling we need.

evansd: The default `shellcheck` [installation instructions](https://github.com/koalaman/shellcheck?tab=readme-ov-file#installing) push you towards system-wide installation which kinds of rubs me the wrong way. It's not terrible for a tool which probably doesn't update much. But ideally we'd be able to say that our linting steps runs with exactly version X of tool Y.

Docker is an option, as we do for Hadolint, but I don't know what performance looks like.

They also have pre-compiled statically linked binaries which we could shove in the virtualenv `bin/` directory. But then we'd need to write code to do that, and select the appropriate version for the current platform.

evansd: This runs OK for me:
```
    check "find docker/ airlock/ -name \*.sh -print0 | xargs -0 docker run --rm -v \"$PWD:/mnt\" koalaman/shellcheck:v0.9.0"
```

bloodearnest: Huh, I didn't conceive of a world where someone would not have shellcheck installed, so there shell scripts are linted properly.

I can use the docker image.  

The job-server directory just landed (and was the reason for this change), I will rebase.

evansd: I simply write correct shell scripts, Simon:grinning:

evansd (approved):

PR #68, opened 2024-02-09
chore: document where to get the token at the point of need

madwort:


dotenv-sample
```diff
@@ -13,6 +13,10 @@ AIRLOCK_REQUEST_DIR=releases/
 AIRLOCK_DEV_USERS_FILE="dev_users.json"
 
 
-# Note: these settings will be automatically updated by `just job-server/{run,stop}
+# Note: in development, these settings will be automatically updated by
+# `just job-server/{run,stop}
+# In production you can find this token for an existing backend in the
+# job-server database at jobserver_backends.auth_token, and you can use
+# the default value for the endpoint.
```
evansd, line 20: You can also get the token from the staff area (this was documented here before but seems to have been removed):
https://github.com/opensafely-core/airlock/blob/3d74a6b8d8ccc49f3e344b2b0f76104733ffcaa8/dotenv-sample#L11-L13

evansd (approved):

dotenv-sample
madwort: ah, that's much more ergonomic, thanks

madwort, line 20:
```suggestion
# Note: in development, these settings will be automatically updated by
# `just job-server/{run,stop}
# In production you can find this token for an existing backend in the
# Job Server staff area at https://jobs.opensafely.org/staff/backends/ ,
# and you can use the default value for the endpoint.
```

PR #71, opened 2024-02-10
Refactor the test factories

bloodearnest:
The initial class based factories were awkward to use. This change refactores them into top level functions.

They now return the core objects directly, rather than an odd Factory object that looked like but wasn't a real object.

Also removed a couple of one line fixtures that didn't add much beyond calling the factory.

This is deliberate move away from a FactoryBoy style factory api to more customisable function based one.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -22,7 +22,7 @@ def create_workspace(name):
     return Workspace(name)
 
 
-def create_request(workspace, user=default_user, request_id=None):
+def create_release_request(workspace, user=default_user, request_id=None):
     workspace = ensure_workspace(workspace)
     if request_id is None:
         request_id = generate_request_id(workspace.name, user)
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -1,63 +1,44 @@
-from dataclasses import dataclass
-
 from django.conf import settings
 
 from airlock.users import User
 from airlock.workspace_api import ReleaseRequest, Workspace, generate_request_id
 
 
-default_user = User(1, "test", [], True)
+default_user = User(1, "testuser")
 
 
-class FileWriterMixin:
-    def path(self):  # pragma: no cover
-        return NotImplemented()
+def ensure_workspace(workspace_or_name):
+    if isinstance(workspace_or_name, str):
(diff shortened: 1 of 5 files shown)


tests/factories.py
```diff
@@ -1,63 +1,44 @@
...
-        return ReleaseRequestFactory(request_id, self.name)
+def create_workspace(name):
+    workspace_dir = settings.WORKSPACE_DIR / name
+    workspace_dir.mkdir(exist_ok=True, parents=True)
+    return Workspace(name)
 
-    def get(self):
-        return Workspace(self.name)
 
+def create_request(workspace, user=default_user, request_id=None):
```
rebkwok, line 25: Can we name this `create_release_request`? I know it's more verbose, but `request` is so easily confused with http request and I had to check back to be sure of which one it was

tests/integration/test_old_api.py
```diff
@@ -22,20 +24,18 @@ def test_old_api_create_release(responses):
 
 
 def test_old_api_upload_file(responses):
-    rf = WorkspaceFactory("workspace").create_request("request-id")
-    rf.write_file("test/file.txt", "test")
-    item = rf.get().get_path("test/file.txt")
+    request = factories.create_request("workspace", request_id="request-id")
```
rebkwok, line 27: Similarly here - can we call it `release_request`, not just `request`?  This one was particularly confusing because of the removed variable called `rf` (pytest-django has a fixture `rf`, for an http `RequestFactory`)

rebkwok (changes requested):
I prefer this approach.  Just a request (hah) to disambiguate `release_request` and `request` (haven't commented on every one, just the first).

tests/integration/test_old_api.py
bloodearnest: Good call - I'll do this as a commit in a later PR, to avoid excessive rebasing on the current stack of PRs

bloodearnest: never mind, will do it now!

bloodearnest: done

rebkwok (approved):

PR #85, opened 2024-02-15
Split API for authored requests vs requests to review

bloodearnest:
Previously, the same function retrieved both sets.  Now, there are
separate API functions, and the Eequests page lists them separatedly.

Specifically, as now an author who is also and output checker cannot
review their own request, then those requests are exclude from the "to
review" list.

### DIFF
--- a/local_db/api.py
+++ b/local_db/api.py
@@ -60,17 +60,30 @@ class LocalDBProvider(ProviderAPI):
                 author=user.username,
             )
 
-    def get_requests_for_user(self, user: User):
+    def get_requests_authored_by_user(self, user: User):
         requests = []
-        filter_args = {}
 
-        # if not output checker, can only see your own requests
-        if not user.output_checker:
-            filter_args = {"author": user.username}
-
-        for metadata in RequestMetadata.objects.filter(**filter_args):
-            if user.output_checker or metadata.workspace in user.workspaces:
+        for metadata in RequestMetadata.objects.filter(author=user.username).order_by(
+            "status"
+        ):
+            # to create a request, user *must* have explicit workspace
+            # permissions - being an output checker is not enough
+            if metadata.workspace in user.workspaces:
                 requests.append(self._request(metadata))
+
+        return requests
+
+    def get_outstanding_requests_for_review(self, user: User):
+        requests = []
+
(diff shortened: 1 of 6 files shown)


tests/unit/test_api.py
```diff
@@ -101,17 +101,11 @@ def test_provider_request_release_files(mock_old_api):
...
-        (
-            [],
-            True,
-            [
-                "r1",
-                "r2",
-                "r3",
-            ],
-        ),
+        # output checkers can't create requests unless explit permission for workspace
```
inglesp, line 104: I don't understand what creating requests has to do with getting requests

local_db/api.py
```diff
@@ -62,15 +62,28 @@ def get_current_request(self, workspace: str, user: User, create=False):
 
     def get_requests_for_user(self, user: User):
```
inglesp, line 63: Maybe `get_requests_for_researcher`?

local_db/api.py
```diff
@@ -62,15 +62,28 @@ def get_current_request(self, workspace: str, user: User, create=False):
...
+            "status"
+        ):
+            # to create a request, user *must* have explicit workspace
+            # permissions - being an output checker is not enough
+            if metadata.workspace in user.workspaces:
+                requests.append(self._request(metadata))
+
+        return requests
+
+    def get_outstanding_requests(self, user: User):
```
inglesp, line 76: Maybe `get_outstanding_requests_for_output_checker`?

inglesp (approved):
Looks good

tests/unit/test_api.py
bloodearnest: We create a request on demand if one doesn't exist. For that one case, we pass create=True.

local_db/api.py
bloodearnest: Gone with `get_requests_authored_by_user`

local_db/api.py
bloodearnest: Gone with `get_outstanding_requests_for_review`

`..._for_output_checker` implies that they are only for a specific output checker, were as is actually *all* such requests. User is only passed to verify they are indeed an output checker

PR #89, opened 2024-02-16
Add initial support for file groups

rebkwok:
Add initial support for file groups. 

The UI for selecting a group or providing a group name for a new one isn't implemented yet, so currently every file added to a request will be added to a single default group.


airlock/api.py
```diff
@@ -315,12 +336,21 @@ def set_status(
...
         """Add a file to a request.
 
         Subclasses should call super().add_file_to_request(...) to do the
-        copying, then record the file metadata as needed.
+        copying, then create/retrieve the file group and record the file
+        metadata as needed.
+
+        After calling add_file_to_request, the current release_request's
+        filegroups will be out of date. Subclasses should return a new
+        ReleaseRequest to ensure filegroups are current.
```
rebkwok, line 353: I'm not sure about this. I know the `set_state` method changes the ReleaseRequest object, but that doesn't feel like a thing that should be done unless it has to.  I've chosen to make `add_file_to_request` return the release_request here, and a new one in the provider api, but the provider api could modify the existing release_request. In the views it doesn't actually need to, because adding a request redirects to the request file view, where the request will be instantiated again and will have the updated groups.

bloodearnest: Yeah, I agree this is an awkward case. I originally wrote `set_status()` to return a new instance in a similar manner for similar reasons.  But the old instance can still be used, so feels like a foot gun (and indeed it did bite me). Unlike Closure, Python programmers are not used to immutable data types, except for, famously, strings. But if we document this aspect of the API, its probably ok?

And like you say, its unlikely to matter in the current usage, but in future that might change.

Also, lists are mutable members of frozen dataclasses, so you *can* actually mutate them as normal, rather than having to assign to __dict__ as I did with `.status`. As you've added a list, ReleaseRequest is no longer hashable, but I but assume that's not been an issue. I don't its avoidable anyway.

In that case, all `frozen=True` does is prevent mutation of normal attributes.

So I actually think mutating the list is probably fine?  But probably worth discussing with rest of RAP (separate to this PR)?

Whatever we decide, there's probably value in doing the same thing across the board for all the APIs that mutate - either mutate passed object, or return a new object.  If we do return new, it would be nice to have some way to force the caller to not use that instance again (aka tainted objects).

local_db/api.py
```diff
@@ -95,3 +129,36 @@ def set_status(self, request: ReleaseRequest, status: Status, user: User):
...
+
+    def add_file_to_request(
+        self,
+        release_request: ReleaseRequest,
+        relpath: Path,
+        user: User,
+        group_name: Optional[str] = "default",
+    ) -> ReleaseRequest:
+        # call super() to copy the file
+        super().add_file_to_request(release_request, relpath, user, group_name)
```
bloodearnest, line 141: So, thinking this through, if the copy succeeds but the db fails, we are left with inconsistent state.

So it might be better to write to the db first, and then call super() to copy *within the transaction block*.

That way, if the copy fails, the db rolls back. And if the db fails, the copy never happens. In either case, we are left with consistent state.

Might need to update the doc string on ProviderAPI.add_file_to_request to match that if so, I think I got it wrong by saying it should call it first.

local_db/models.py
```diff
@@ -45,3 +45,25 @@ class RequestMetadata(models.Model):
     status = StatusField(default=Status.PENDING)
     author = models.TextField()  # just username, as we have no User model
     created_at = models.DateTimeField(default=timezone.now)
+
+
+class FileGroupMetadata(models.Model):
+    """A group of files that share context and controls"""
+
+    request_id = models.TextField()
```
bloodearnest, line 53: Could this be an FK to ReleaseMetadata?

local_db/models.py
```diff
@@ -45,3 +45,25 @@ class RequestMetadata(models.Model):
...
+    name = models.TextField(default="default")
+
+    class Meta:
+        unique_together = ("request_id", "name")
```
bloodearnest, line 57: nice!

bloodearnest (approved):
This looks great, thanks.

Have made a comment about moving the file copy inside the db transaction - I do think that's probably important to do?

Left a few other comments that more trivial, feel free to ignore.

I think the questions of mutate or return new is worth getting an answer to, but can wait till a later PR I think?

local_db/api.py
rebkwok: Yes, that makes sense, and now that I've started poking things from the frontend I've come across that inconsistent state already

local_db/models.py
rebkwok: Yes, I think so

airlock/api.py
rebkwok: OK, I've left it as is (creating the new object). Something to discuss at the next code walkthrough, maybe. Do we want ReleaseRequest to be hashable? It's not been an issue yet, but a jobserver provider API might want it? I'm going to leave that for later too, though.

airlock/api.py
```diff
@@ -102,6 +103,25 @@ def abspath(self, relpath):
...
+
+
+@dataclass(frozen=True)
+class FileGroup:
+    """
+    Represents a group of one or more files within a release request
+    """
+
+    name: str
+    files: [RequestFile]
```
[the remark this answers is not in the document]
rebkwok: Actually, I don't think it is, and was just an error on my part...at least I can't find any evidence that it is, even though `just check` didn't complain.

PR #94, opened 2024-02-19
Add tree base navigation UI

bloodearnest:
- minimal functional implementation, not pretty
 - uses <details>, so no js needed
 - can switch at run time to test, but defaults to new UI

In order for the template to be able to know if the current item in the
tree is, or is a parent of, the currently selected path, I added the
ability to designate a currently selected path to PathItem, as we can't
express that logic cleanly in the DTL.

Note: supporting both UIs behind a feature flag intended to be short
lived, so that code should be deleted soon.

### DIFF
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -21,6 +21,11 @@ class PathItem:
     container: AirlockContainer
     relpath: pathlib.Path
 
+    # The currently selected view path in this container.  Used to calculate if
+    # how relpath relates to it, i.e. if relpath *is* the currently selected
+    # path, or is one of its parents.
+    selected: pathlib.Path = None
+
     def __post_init__(self):
         # ensure relpath is a Path
         object.__setattr__(self, "relpath", pathlib.Path(self.relpath))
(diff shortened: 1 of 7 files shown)


bloodearnest:
image

airlock/file_browser_api.py
```diff
@@ -21,6 +21,10 @@ class PathItem:
     container: AirlockContainer
     relpath: pathlib.Path
 
+    # The currently selected path in this container is current the selected.
```
rebkwok, line 24: Something's mangled in this comment

airlock/file_browser_api.py
```diff
@@ -21,6 +21,10 @@ class PathItem:
...
+    # Used to calculate if how relpath relates to it.
```
rebkwok, line 25: And this one :)

airlock/file_browser_api.py
```diff
@@ -21,6 +21,10 @@ class PathItem:
...
+    selected: pathlib.Path = None
```
rebkwok, line 26: Should/can selected be an attribute of the container rather than this PathItem? It doesn't feel like a PathItem should be holding information about the selected path in its container.

airlock/views.py
```diff
@@ -210,8 +227,9 @@ def request_view(request, request_id: str, path: str = ""):
     )
 
     context = {
-        "workspace": release_request.workspace,
+        "workspace": api.get_workspace(release_request.workspace),
```
rebkwok, line 230: I fixed this too, in a PR on Friday. I guess it can be closed in favour of this one

rebkwok (approved):
LGTM. Some comments need updating and a question about where the selected path should live but I don't have very strong opinions either way

airlock/file_browser_api.py
bloodearnest: blargl, mondays, will fix

airlock/views.py
bloodearnest: ah, yeah, I approved it I think. This was just a drive by, lets merge yours

airlock/file_browser_api.py
bloodearnest: I did wonder that, but what path is currently selected is not something the Workspace or ReleaseRequest cares about - its temporary UI state, its never persisted in any way.

And PathItem is designed to be that UI focussed object, so I do think it is in the right place.
I

PR #95, opened 2024-02-19
Add support for adding a file to a new or existing file group on a request

### LINKED ISSUE #77: Add support for grouping files together
Users need to be able group files in a release request together.

When a user adds a file to a release request, they should see a modal dialog box with the option to add a file to an existing group, or to create a new group.

There should be a way to add a file to a release request, so that it is not in a group.  Perhaps there could be a special "default" group?

An initial implementation can be done entirely using server-side rendering.

rebkwok:
Fixes #77 

Adds a modal to the "Add file to request" button, with an option to select a file group, or to add a name to create a new one. 
Screenshot from 2024-02-19 11-44-29

For now, in the release request view, I've added a button and modal to view the file groups. I expect this will change with the new tree browser UI, so this is intended to be just a temporary (and easily removable) way of seeing which files have been added to which groups. 
Screenshot from 2024-02-19 11-45-56
Screenshot from 2024-02-19 11-48-25

### DIFF
--- a/airlock/forms.py
+++ b/airlock/forms.py
@@ -1,0 +1,30 @@
+from django import forms
+
+
+class AddFileForm(forms.Form):
+    filegroup = forms.ChoiceField(required=False)
+    new_filegroup = forms.CharField(required=False)
+
+    def __init__(self, *args, **kwargs):
+        release_request = kwargs.pop("release_request")
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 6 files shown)


bloodearnest (approved):
This is great, nice to see how a form and modal look using the components.

I have a local branch for how to view file groups in the new tree veiw.

PR #99, opened 2024-02-21
Unfreeze core objects.

bloodearnest:
Update the two places we mutate state to mutate the object.

Manually made Workspace hashable, for convenience in tests, but also it
is just one bit of immutable data.

### DIFF
--- a/airlock/api.py
+++ b/airlock/api.py
@@ -44,7 +44,7 @@ class AirlockContainer:
         raise NotImplementedError()
 
 
-@dataclass(frozen=True)
+@dataclass
 class Workspace(AirlockContainer):
     """Simple wrapper around a workspace directory on disk.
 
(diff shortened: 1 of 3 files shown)


airlock/api.py
```diff
@@ -316,30 +319,22 @@ def set_status(
...
         release_request: ReleaseRequest,
         relpath: Path,
         user: User,
         group_name: Optional[str] = "default",
-    ) -> ReleaseRequest:
+    ):
         """Add a file to a request.
 
         Subclasses should do what they need to create the filegroup and
         record the file metadata as needed and THEN
```
rebkwok, line 334: Should there be a comment that subclasses should also update the filegroups on this release request?

rebkwok (approved):

airlock/api.py
bloodearnest: Sorry, automerge stuck again.

There should, I will add in next PR

PR #102, opened 2024-02-23
group trees

bloodearnest:
- **Make test factories use the api to create requests**
- **Refactor PathItem to support groups.**

### DIFF
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -1,9 +1,9 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from airlock.api import ROOT_PATH, AirlockContainer, UrlPath
 
 
-@dataclass(frozen=True)
+@dataclass
 class PathItem:
     """
     This provides a thin abstraction over url paths and related filesystem objects with two goals:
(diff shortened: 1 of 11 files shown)


airlock/file_browser_api.py
```diff
@@ -17,58 +18,72 @@ class PathItem:
...
     """
 
+    class PathNotFound(Exception):
+        pass
+
     container: AirlockContainer
-    relpath: UrlPath = ROOT_PATH
+    relpath: UrlPath
+
+    type: str = None
```
rebkwok, line 27: I'm never 100% sure when an enum is a good thing, but would it make sense here? PathItem type should only ever be dir or file, shouldn't it?

rebkwok: Oh, no, I see it can also be workspace/request - maybe an enum is useful then, so it's clear what the possible values are

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+        def walk_selected(node):
+            """Traverse tree to find selected node."""
+            for child in node.children:
+                if child.selected:
+                    return child
+
+                if child.on_selected_path:
+                    return walk_selected(child)
+
+            raise self.PathNotFound("Not selected path found")
```
rebkwok, line 167:
```suggestion
            raise self.PathNotFound("No selected path found")
```

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+    # ensure selected_path is UrlPath
+    selected_path = UrlPath(selected_path)
+    return build_workspace_tree(ROOT_PATH, parent=None)
+
+
+def get_request_tree(release_request, selected_path=ROOT_PATH):
+    """Build a tree recursively for a ReleaseRequest
+
+    For each group, we create a node for that group, and then build a sub-tree
+    its files group.
```
rebkwok, line 222:
```suggestion
    for its file groups.
```

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+
+        root_node.children.append(group_node)
+
+    return root_node
+
+
+def get_filegroup_tree(container, selected_path, group_data, group_path, parent):
+    """Get the tree for a filegroup's files.
+
+    This is more than just a walk the disk. The FileGroup.files is flat list of
```
rebkwok, line 261:
```suggestion
    This is more than just a walk the disk. The FileGroup.files is a flat list of
```

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+    relative paths. So we need group those by common prefex and descend down
```
rebkwok, line 262:
```suggestion
    relative paths. So we need to group those by common prefix and descend down
```

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+    the tree.
+    """
+
+    def build_filegroup_tree(file_parts, path, parent):
+        """Walk a flat list of paths and create a diretcories tree for them."""
```
rebkwok, line 267:
```suggestion
        """Walk a flat list of paths and create a directories tree for them."""
```

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+        grouped = defaultdict(list)
+        for child, *descendants in file_parts:
+            if descendants:
+                grouped[child].append(descendants)
+            else:
+                grouped[child] = []
+
+        tree = []
+
+        # now we have them grouped by first path element, we can great a node
```
rebkwok, line 279:
```suggestion
        # now we have them grouped by first path element, we can create a node
```

airlock/file_browser_api.py
```diff
@@ -95,24 +110,218 @@ def html_classes(self):
...
+
+        # sort directories first then files
+        tree.sort(key=children_sort_key)
+        return tree
+
+    file_parts = [f.relpath.parts for f in group_data.files]
+    return build_filegroup_tree(file_parts, group_path, parent)
+
+
+def is_selected(path, selected_path):
```
rebkwok, line 317: This was confusing because a PathItem also has an `is_selected` method, and because I was expecting it to return a boolean (so unpacking it was odd).

rebkwok (commented):
Have left some minor comments, but it'd be good to talk through the complicated parts

tests/factories.py
```diff
@@ -22,33 +26,46 @@ def ensure_workspace(workspace_or_name):
...
+        user = get_user(username=request.author, workspaces=[request.workspace])
 
-def write_request_file(request, path, contents=""):
-    path = request.root() / path
-    path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(contents)
+    api.add_file_to_request(request, relpath=path, user=user, group_name=group)
 
 
 def create_filegroup(release_request, group_name, filepaths=None):
```
rebkwok, line 67: Is this helper method needed anymore? I added it when we weren't using the api to create request files for tests, so they weren't automatically getting groups created

airlock/file_browser_api.py
bloodearnest: Yes, I think you're right, I will add an enum :+1:

airlock/file_browser_api.py
bloodearnest: yeah, it is not clean.

But there are basically two attributes to set based on the selected_path, and I didn't want to have to repeat the logic each time we create a PathItem, and this is what I came up with.

The alternative as a post-instiation `set_selected(item, selected_path)` function I guess? I thought this was neater.

But I also wrote it when PathItem was still a frozen dataclass. I had to make it mutable in the end, so that the children could have a reference to their parent, before the parent's children had been set.

I'll have another look and see if there's a better way to do this.

tests/factories.py
bloodearnest: There's a couple of other tests that still use it I think?

I will try rework them to use the write_release_request_files changes

airlock/file_browser_api.py
bloodearnest: Have moved to a `set_selected(node, selected_path)` pattern.

rebkwok: tbh I think just naming it differently would have unconfused me, but this works :+1:

rebkwok (approved):

PR #103, opened 2024-02-23
End to end tests for creating, releasing and rejecting requests

### LINKED ISSUE #79: Write an end-to-end test with playwright
This should cover:

- [ ] logging in as a researcher
- [ ] viewing files
- [ ] adding files to a request
- [ ] submitting the request
- [ ] logging out
- [ ] logging in as an output checker
- [ ] viewing files
- [ ] releasing the files

rebkwok:
Locally these two new e2e tests take 6.5s to run; that will be better when we've settled on a UI, but for now it seems worth running them with the ui_options fixture.

Fixes #79

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -99,6 +99,17 @@ Additional arguments passed to `just test` are passed on to pytest. For example,
 run all the tests _except_ the functional tests, run `just test -k 'not functional'`,
 or to run a single test, run e.g. `just test tests/unit/test_urls.py::test_urls`.
 
+### Debugging functional tests
+
+Functional tests run headless by default. To see what's going on, they
+can be run in headed mode. The follwoing command will run just the
+functional tests, in headed mode, slowed down by 500ms. See the
+[playwright docs](https://playwright.dev/python/docs/test-runners#cli-arguments) for additional cli arguments that may be
+useful.
+
+```
+just test -k functional --headed --slowmo 500
+```
 
 # Local job-server for integration.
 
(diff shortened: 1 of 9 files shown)


rebkwok:
> It might be good to do the work now to add id, data- or aria- attributes to our existing html and use those in the tests?

Yes, I had considered that - but wasn't sure how we felt about adding ids just for use in tests.  I think it's probably worth it for not breaking on the text content, and i don't think it's a very big job to do now. Will have a look.

Annoyingly, playwright has a `get_by_test_id` method so you can add a data-test-id attribute, but I can't add it on slippers components. We can add ids though.

bloodearnest:
> > It might be good to do the work now to add id, data- or aria- attributes to our existing html and use those in the tests?
> 
> Yes, I had considered that - but wasn't sure how we felt about adding ids just for use in tests. 

I think ids by default is good idea in general. You never know when they're also useful for css/js targetting
>I think it's probably worth it for not breaking on the text content, and i don't think it's a very big job to do now. Will have a look.

Great, thanks
 
> Annoyingly, playwright has a `get_by_test_id` method so you can add a data-test-id attribute, but I can't add it on slippers components. We can add ids though.

Ah, that's an interesting limitation. I wonder if we can update the base slippers stuff to always take a test_id parameter?

rebkwok:
@bloodearnest Done - I've added testids and ids where I can. I did make one change to make a slippers component pass on an id to a link-type button, but I avoided changing the components otherwise since they're coming from job-server. It would be pretty simple to let any component take an optional `data-testid` attribute though.

PR #112, opened 2024-02-27
Display CSV files as simple tables

rebkwok:
Makes file content scrollable independent of the rest of the page.
Displays CSV files as a simple table, horizontally and vertically scrollable.
No pagination, sorting or filtering as yet.


airlock/templatetags/airlocktags.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@register.simple_tag
+def as_csv_data(file_content):
+    rows = file_content.strip().split("\n")
+    data = {
+        "headers": _split_row(rows[0]),
+        "rows": [_split_row(row) for row in rows[1:]],
+    }
+    return data
```
evansd, line 14: Is there any reason not to use the `csv` module here? It's less code and will handle quoted values correctly.
```py
reader = csv.reader(file_content.splitlines(keepends=True))
return {
    "headers": next(reader),
    "rows": list(reader),
}
```

evansd (approved):

airlock/templatetags/airlocktags.py
rebkwok: For some reason the fact that it wasn't a file at this point meant I didn't - will do that

PR #115, opened 2024-02-28
Add local_db to coverage

rebkwok:
The local_db app was missing from coverage; this also removes a bit of inaccessible code and adds a couple more tests for code that wasn't fully covered.

### DIFF
--- a/local_db/models.py
+++ b/local_db/models.py
@@ -20,7 +20,7 @@ class StatusField(models.TextField):
     choices = [(i.value, i.name) for i in Status]
 
     def from_db_value(self, value, expression, connection):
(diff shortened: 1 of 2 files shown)


local_db/models.py
```diff
@@ -20,7 +20,7 @@ class StatusField(models.TextField):
     choices = [(i.value, i.name) for i in Status]
 
     def from_db_value(self, value, expression, connection):
-        if value is None:
+        if value is None:  # pragma: no cover
```
rebkwok, line 23: skipping coverage for this because it's only used in the RequestMetadata model, where it's non-nullable

PR #117, opened 2024-02-28
Refactor views

rebkwok:
Split up views.py and their corresponding tests.

Nothing has changed in this PR, just rearranged.


bloodearnest (approved):
Thanks for doing this.

I am not keen on the `_views` suffix being on all the separate files (and thus all the tests filenames too). They are already in the `views` directory, I don't think we need to repeat it in the filename?

Partly because of the repetition and longer paths, but also because I am not keen on importable modules names having underscores in them if it is avoidable?

That said, this is a minor stylistic quibble.  I'd be happy to do the renaming myself in a follow on PR, since you've done all the legwork here already?

rebkwok:
> I am not keen on the `_views` suffix being on all the separate files (and thus all the tests filenames too). They are already in the `views` directory, I don't think we need to repeat it in the filename?

I don't have a very strong preference, so happy to change it if you do. I tend to use it just because it makes it easier to see which file I'm looking at by the tab names in vscode (I dislike the standard ansible structure where I end up with multiple `main.yaml`s open, for example).

PR #130, opened 2024-03-04
Faster csv rendering

bloodearnest:
- **Render csv using plain html templates**
- **Make csv table headers sticky**

Using components, a 3k line csv took ~6s to render. After this change, about 0.25s


evansd (approved):
Changes look good but could do with a slightly more descriptive PR title and some mention in the PR description of the performance difference (as described in the thread).

PR #131, opened 2024-03-04
Bring back django debug toolbar

bloodearnest:
But only in dev.

Also, it is not on by default, but setting DJANGO_DEBUG_TOOLBAR env var will do
all the the things needed to enable it.

Also included the template profile extension, as that was my original
motivation.

### DIFF
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -63,6 +63,11 @@ SECRET_KEY = get_env_var("DJANGO_SECRET_KEY")
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = get_env_var("DJANGO_DEBUG") == "True"
 
+if DEBUG:  # pragma: no cover
+    DJANGO_DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", "false") == "True"
+else:  # pragma: no cover
+    DJANGO_DEBUG_TOOLBAR = False
+
 ALLOWED_HOSTS = get_env_var("DJANGO_ALLOWED_HOSTS").split(",")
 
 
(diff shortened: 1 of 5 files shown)


airlock/settings.py
```diff
@@ -63,6 +63,15 @@ def get_env_var(name):
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = get_env_var("DJANGO_DEBUG") == "True"
 
+if DEBUG:  # pragma: no cover
+    DEBUG_TOOLBAR = os.environ.get("DEBUG_TOOLBAR", "false").lower() in [
```
evansd, line 67: For consistency with existing env vars (`DJANGO_DEBUG`, `DJANGO_ALLOWED_HOSTS`) the env var here should probably be `DJANGO_DEBUG_TOOLBAR`. Although admittedly this is a dev only setting gated behind `DJANGO_DEBUG` so I feel a bit less strongly about the namespacing here.

airlock/settings.py
```diff
@@ -63,6 +63,15 @@ def get_env_var(name):
...
+        "1",
+        "true",
+        "yes",
```
evansd, line 70: I'm not wild about having an ad-hoc list of truthy alternatives for this one specific value. If we want to have multiple ways to spell booleans then I think we should have a function somewhere which handles these consistently, and rejects unrecognised values.

Again, I realise this is a dev only setting and so it doesn't matter that much how we handle things here. But as soon as code gets merged it starts to gain the status of an established pattern that then gets copied elsewhere.

evansd (approved):
Very much yes to the principle here, although a bit of bikeshedding over the implementation.

Also, formatting on the PR description looks  bit wonky?

bloodearnest:
> Very much yes to the principle here, although a bit of bikeshedding over the implementation.
> 
> Also, formatting on the PR description looks bit wonky?

Yeah, I think `gh pr create` mangled it

PR #132, opened 2024-03-05
Run playwright tests with chrome v108

rebkwok:
tl;dr There were 2 fiddly things involved in this:
1) getting playwright to run with an old browser version
2) getting the old browser version (google don't make old chrome versions easily available)

1) Playwright by default only runs with the latest stable browsers.
An [executable_path](https://playwright.dev/python/docs/api/class-browsertype#browser-type-executable-path) can be specified when launching a playwright browser, however, the pytest-playwright plugin doesn't provide a way to pass this option.
(Note that the PLAYWRIGHT_BROWSERS_PATH env var seemed promising initially, but is a way to install the browser binaries that playwright expects to a custom path, and/or to specify that path. Playwright only runs with the latest stable browsers by default, so we can't just point it at a path to an old chrome version.)
So I've monkeypatched the fixture that passes the browser launch config, so it now lets us specify a custom browser executable using the PLAYWRIGHT_BROWSER_EXECUTABLE_PATH environment variable. It also
skips playwright browser installation if this variable is set. If the env var is not set, it'll use the default chromium browser as before.

2) Google does not offer old builds of Chrome, but it is possible to get old builds of chromium that should match the version we need (currently 108 on TPP) for testing.
I followed the [instructions here](https://www.chromium.org/getting-involved/download-chromium/#downloading-old-builds-of-chrome-chromium) to find a v108 chromium.
The just command for `docker/test` now fetches the linux download from the [nearest build](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Linux_x64/1058929/), unzips it, and puts it it a folder that's mounted into the docker image. We still need to run `playwright install chromium` in the docker image in order ensure the image has the system dependencies chromium needs.

In CI, the non-docker tests will run with the latest chromium, and the docker tests will run with v108.  I didn't want to make the non-docker ones download the old chromium and try to run with it because it'd need to get a different version depending on the OS.

### DIFF
--- a/.test.env
+++ b/.test.env
@@ -9,3 +9,4 @@ AIRLOCK_WORK_DIR=workdir/
 # Ensure playwright browser binaries are installed to a path in the
 # mounted volume that the non-root docker user has access to
 PLAYWRIGHT_BROWSERS_PATH=/app/.playwright-browsers/
+PLAYWRIGHT_BROWSER_EXECUTABLE_PATH=/app/.playwright-browsers/chrome-linux/chrome
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -117,7 +117,10 @@ Additional arguments passed to `just test` are passed on to pytest. For example,
 run all the tests _except_ the functional tests, run `just test -k 'not functional'`,
 or to run a single test, run e.g. `just test tests/unit/test_urls.py::test_urls`.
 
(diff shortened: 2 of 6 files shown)


bloodearnest (approved):
LGTM, this is indeed a bit more fiddly than you'd think!

PR #148, opened 2024-03-06
File downloading

### LINKED ISSUE #126: Support downloading files to remote desktop in secure environment
To allow users to explore their outputs (especially tabular data), and for output checkers to perform disclosivity checks on tabular data, we need to support users downloading files through Airlock onto their remote desktop in the secure environment.  This will allow them to open files in eg a spreadsheet program.

Each page showing a file should have a "Download" button and the downloaded file should have a suitable filename.

The documentation should be updated to explain this.

UPDATE:
Downloads will only be available for files in release requests, and only for output checkers. We won't add restrictions on the file type (tabular data is both the most likely use case for downloaded, but also the most risky).  All downloads will be logged (full access log functionality will be part of another ticket).

rebkwok:
Uses the existing `contents_url`, and the default for the download filename (i.e. uses the existing filename). I don't think there's a need to include group names or paths in the downloaded filename.

As discussed, only output-checkers can download release request files, and only for requests that they did not author. We don't restrict the types of files that can be downloaded by output-checkers. No workspace files can be downloaded by anyone. 

Fixes #126

### DIFF
--- a/airlock/api.py
+++ b/airlock/api.py
@@ -52,8 +52,10 @@ class AirlockContainer(Protocol):
     def get_url(self, path: UrlPath = ROOT_PATH) -> str:
         """Get the url for the container object with path"""
 
-    def get_contents_url(self, path: UrlPath = ROOT_PATH) -> str:
-        """Get the url for the container object with path"""
+    def get_contents_url(
+        self, path: UrlPath = ROOT_PATH, download: bool = False
+    ) -> str:
+        """Get the url for the contents of the container object with path"""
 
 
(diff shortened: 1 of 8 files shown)


bloodearnest (approved):
This looks good.

Where did we land on the "only downloading csv files" question?

rebkwok:
> Where did we land on the "only downloading csv files" question?

We said we'd allow any files to be downloaded, given that CSV is the most high-risk anyway, and we're limiting it to output-checkers and release request files only

PR #150, opened 2024-03-06
Disable add-file-to-request button for output-checkers

rebkwok:
if they do not have explicit permission for the workspace.


bloodearnest (commented):
Should we also check for the permission in the add_file_to_request POST view as well?

bloodearnest (approved):

Issue #121, opened 2024-03-01
Explore storing all versions of files that have been requested for release


inglesp:
At present, we use the filesystem and the database to store mutable state. This makes maintaining atomicity of operations difficult, a problem which will only become worse if we switch to using a remote API instead of the local db.

One solution to this is switch to a content-addressed storage model where we use the filesystem as, effectively, a dumb immutable blob store and all the interesting state, and more importantly state _changes_, happen in the database.

PR #155, opened 2024-03-08
Refactor ProviderAPI to Business Logic and Data Access Layers

### LINKED ISSUE #125: Refactor ProviderAPI to Business Logic and Data Access Layers
[As discussed in Slack](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1709208384098359).

evansd:
As it stands, the refactored version of `add_file_to_request` isn't safe because we could update the filesystem and then the database update could fail. I think the correct solution here is move to an immutable content-addressed file storage system as described in #121. I'll tackle this in a follow-up PR, but I think we can live with it for now.

(I don't actually think the previous implementation was safe anyway because we could update the filesystem and then the transaction could fail to commit for some other reason like the database being too busy.)

Closes #125


bloodearnest (approved):
This makes sense to me, and is probably the design we should have gone for from the start.  

Am not in love with the `from_dict` approach, but it does provide a flexible semantic barrier between the two APIs. And I don't expect the dicts to get especially complex, so I think its not an issue.

PR #162, opened 2024-03-12
Update to new login api

### LINKED ISSUE #146: Workspace index improvements
On TPP, there are a lot of workspaces.

On the workspace index, the user is presented with a seemingly unordered list. If they are output checker it is all workspaces.

We should add a sort order, probably alphabetical is simplest, athough sorting by "most recently had updated files" might be nice.

For output-checkers, they should probably just see their explicit workspaces here, rather than all workspaces.  When performing a review, they can navigate to that release request's workspace using the link in the release-request page - they'll still have permission to access it.

bloodearnest:
Use the new API/url, and consume the new metadata it gives us

Fixes #146

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -60,7 +60,7 @@ class AirlockContainer(Protocol):
         """Get the url for the contents of the container object with path"""
 
 
-@dataclass
+@dataclass(order=True)
 class Workspace:
     """Simple wrapper around a workspace directory on disk.
 
@@ -69,6 +69,7 @@ class Workspace:
     """
 
     name: str
+    metadata: dict = field(default_factory=dict)
 
     # can be set to mark the currently selected path in this workspace
     selected_path: UrlPath = ROOT_PATH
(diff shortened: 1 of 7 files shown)


airlock/business_logic.py
```diff
@@ -77,6 +78,12 @@ def __post_init__(self):
         if not self.root().exists():
             raise BusinessLogicLayer.WorkspaceNotFound(self.name)
 
+    def project(self):
+        return self.metadata.get("project", None)
+
+    def __lt__(self, other):
```
rebkwok, line 84: I think `@dataclass(order=True)` would mean we don't need this; it compares using the attributes in order, so should still sort properly by name

airlock/business_logic.py
```diff
@@ -77,6 +78,12 @@ def __post_init__(self):
         if not self.root().exists():
             raise BusinessLogicLayer.WorkspaceNotFound(self.name)
 
+    def project(self):
```
rebkwok, line 81: Would this be better as a property (or renamed to `get_project`)?

rebkwok (approved):
Couple of questions, nothing major

airlock/business_logic.py
bloodearnest: oh, nice will add

airlock/business_logic.py
bloodearnest: That would be my instinct, but I was following @evansd's lead on the original design for PathItems and such. `root` could also be a property

rebkwok: I would probably also make `root` a property, but no very strong feelings either way

PR #164, opened 2024-03-13
Switch to content-addressed storage model for files

### LINKED ISSUE #121: Explore storing all versions of files that have been requested for release

evansd:
This attempts to be the minimal set of changes needed to move to this model, but does not introduce any major refactoring or new features. In particular, it doesn't yet close #121 although it gives us the machinery to be able to do that.

The primary benefit this does give us is that all (mutable) state is now unambiguously in the database rather than a mixture of the database and the filesystem. This makes the system safer and, I think, easier to reason about.

Deploying this will involve a database migration and also running a script to migrate existing files into the new system:

    ./manage.py backpopulate_file_id

I've made a couple of decisions here worth noting:

1. The files are still stored in separate directories unique to each ReleaseRequest. In theory, we could just have them all in one big bucket. But it seemed like it might be operationally easier to partition them like this, and it doesn't really add anything in the way of complexity.

2. The files are stored simply as the hex string of the hash (like git does). We did discuss appending some kind of debugging information on the end to make it easier to identify files by eye. I think on reflection though, we'd be better off investing in good debug tooling and keeping the storage model as simple as possible. (We went through a similar process with container names on job-runner and have ended up removing the debug information.)


airlock/business_logic.py
```diff
@@ -212,24 +213,19 @@ def abspath(self, relpath):
         The first part of the relpath is the group, so we parse and validate that first.
         """
         relpath = UrlPath(relpath)
-        root = self.root()
-
         group = relpath.parts[0]
+        file_relpath = UrlPath(*relpath.parts[1:])
```
bloodearnest, line 217: Huh, I've been doing `relpath.relative_to(relpath.parts[0])` to do this.

I think this is probably cleaner. Shame there isn't a slice API on Path!

bloodearnest (approved):
This looks good to me.

Agreed on decision 1.

Less sure about decision 2, but not sure what kind of debugging tool you have in mind.

There's a potential small issue post back population, in that we'll maybe have some empty directory trees in the request directory, as only the files themselves are removed AIUI.

But that's not a problem really.

airlock/business_logic.py
evansd: I wasn't quite clear whether you were saying your version is cleaner or mine is? Happy with either, as neither is really a thing of beauty.

evansd:
On debugging tooling, the example you just gave is exactly the sort of thing I had in mind.

> There's a potential small issue post back population, in that we'll maybe have some empty directory trees in the request directory, as only the files themselves are removed AIUI.

That's a good point. I might as well fix that now for the sake of tidiness.

airlock/business_logic.py
bloodearnest: I meant yours. Sliceing the tail of something with `[1:]` is fairly idomatic, and whilst its not quite as simply as that, I think its probably still more readable.

PR #167, opened 2024-03-14
Refactor how we create and use test users

bloodearnest:
In preparation for pushing more user validation checks down into the BLL, this change refactors test user creation to be easier and more consistant, now that User.workspaces must be a dict.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,16 +1,17 @@
 import pytest
 
-from airlock.users import User
+from tests import factories
 
 
 @pytest.fixture
 def client_with_user(client):
     def _client(session_user):
-        session_user = {"id": 1, "username": "test", **session_user}
+        session_user.setdefault("username", "test")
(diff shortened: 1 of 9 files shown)


tests/integration/conftest.py
```diff
@@ -1,23 +1,21 @@
...
 
 
-@pytest.fixture
-def client_with_user(client):
-    def _client(session_user):
-        session_user.setdefault("username", "test")
-        user = factories.create_user(**session_user)
-        session = client.session
+class AirlockClient(Client):
+    def login(self, username="testuser", workspaces={}, output_checker=False):
```
inglesp, line 8: Seeing mutable default arguments makes me nervous.  Could you change to `workspaces=None`?  `factories.create_user` will then do the right thing.

inglesp (approved):
A really nice set of changes.

evansd:
Yeah, seconded! An improvement along many dimensions and really nicely explained along the way.

PR #171, opened 2024-03-14
permissions checks bll

bloodearnest:
- **Move the permissions checks from validate_workspace into the BLL**
- **Add request permissions check into BLL**
- **Rename validate_foo to get_foo_or_raise**

### DIFF
--- a/airlock/views/helpers.py
+++ b/airlock/views/helpers.py
@@ -11,7 +11,7 @@ def login_exempt(view):
     return view
 
 
-def validate_workspace(user, workspace_name):
+def get_workspace_or_raise(user, workspace_name):
     """Ensure the workspace exists and the current user has permissions to access it."""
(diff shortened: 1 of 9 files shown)


airlock/views/helpers.py
```diff
@@ -11,7 +11,7 @@ def login_exempt(view):
     return view
 
 
-def validate_workspace(user, workspace_name):
+def get_workspace_or_raise(user, workspace_name):
     """Ensure the workspace exists and the current user has permissions to access it."""
```
inglesp, line 15: Not sure this needs a comment now.

inglesp (approved):

airlock/views/helpers.py
bloodearnest: fixed!

PR #172, opened 2024-03-15
Show all file content in iframes

### LINKED ISSUE #142: Lazily load csv and text file content from direct url.
Currently, for rendering file contents, we have two different techniques.

For html and images, we use `<iframe>` or `` respectively with `src=url...` tag. This means the content is loaded separately by the browser 

This is useful because the url responses with appropriate cache headers, so browser can cache and do conditional GETs and avoid reloading as a user switches between files.

For csv and text, we render the content directly inline, as part of the entire page html.  This makes caching the file contents not really possible, as they are mixed in with all the rest. It also makes the template rendering slower and more complex, as its embedding a lot of html in the rendering.

Instead, we could split out the generation of the html representation of csv and plain text into a separate urls and link to that, just like we do for html iframes and image tags.  This would give us the benefits of cacheability for this content too, and make the main view html simpler and faster.

Given the renderable content of csv's and plain text is actually html, there are two ways to do this:

1. Use iframe. This is simple, but iframes are weird beasts to control. We use them for html outputs mainly because we want to isolate from them from the css/js in the container page.  Doing this would mean we would need to bundle the css for tables and plain text (e.g. syntax highlighting) in the embedable html content too.

2. Use HTMX's `hx-get` and `hx-trigger=load` to lazily load them, and just inject the html into the page at the right place.  We control the html for these, so we don't need to worry about isolation as much.

bloodearnest:
Fixes #142


tests/integration/views/test_helpers.py
```diff
@@ -1,16 +1,70 @@
...
+    "suffix,mimetype",
+    [
+        (".html", "text/html"),
+        (".csv", "text/html"),
+        (".png", "image/png"),
+        (".txt", "text/html"),
+    ],
+)
+def test_serve_file_rendered(tmp_path, suffix, mimetype):
+    rf = RequestFactory()
```
rebkwok, line 19: pytest-django has a `rf` fixture that could be used here instead (and in the other tests in this file)

tests/integration/views/test_helpers.py
```diff
@@ -1,16 +1,70 @@
...
+@pytest.mark.parametrize(
+    "filename,mimetype",
+    [
+        ("foo.html", "text/html"),
+        ("foo.csv", "text/html"),
+        ("foo.png", "image/png"),
+        ("foo.txt", "text/html"),
+    ],
+)
+def test_serve_file_filename(tmp_path, filename, mimetype):
```
rebkwok: very minor nit; the test is testing the headers, not the filename

airlock/views/helpers.py
```diff
@@ -37,40 +44,69 @@ def get_release_request_or_raise(user, request_id):
...
+        @functools.wraps(func)
+        def wrapper(abspath, suffix):
+            context = func(abspath, suffix)
+            return SimpleTemplateResponse(template, context)
+
+        wrapper.template_path = template_path
+
+        return wrapper
+
+    return decorator
```
rebkwok, line 74: Neat

rebkwok (approved):

tests/integration/views/test_helpers.py
bloodearnest: Ah, right, yes. It's testing it with the filename parameter though.

tests/integration/views/test_helpers.py
bloodearnest: Ah, yes, I'd forgotten about that, will add.

tests/integration/views/test_helpers.py
bloodearnest: Also, this didn't make it into the final commit, was intermediate, so already gone!

PR #173, opened 2024-03-15
SPIKE: Add mypy type checking

evansd:
This is spike to get a sense of what would be involved in adding type checking to Airlock and how it might benefit us.

This involves installing `mypy` and also type stubs and a mypy plugin to get it to play nicely with Django.

Worth noting is that the latest release of the Django type stubs supports neither the latest version of mypy nor the latest version Django:
https://github.com/typeddjango/django-stubs?tab=readme-ov-file#version-compatibility

However, in practice it seems to have worked without issue in the sense that I don't think any of the problems I've hit have been the result of django-stubs.

I started by getting `mypy` passing on its default settings. This involved a little bit of faff to correct some of our type signatures and a few other bits and pieces. It wasn't too much work, but then neither did it expose any interesting bugs.

I then started enabling all the flags involved in `mypy --strict` mode. Enabling them one-by-one in rough order of how many issues they seemed to find and making the necessary fixes. This did expose some small genuine bugs, but nothing earth shattering.

At first this work progressed quite quickly and I got as far as these three final flags:

    --disallow-untyped-calls
    --disallow-incomplete-defs
    --disallow-untyped-defs

However, what I discovered is that `mypy` had been merrily ignoring all the untyped parts of the codebase and the more of it I added types to the more issues it found.

Some of the work here is pretty straightforward and mechanical (e.g. adding `-> str:` to the end of a lot of function defs) but some of it, in particular the recursive tree walking code where nested structures are getting passed around, is difficult. I think _probably_ the pressure from `mypy` here is good in the sense that easier to type code would also be easier to understand. But it's not quite as simple as just slapping a few annotations on and calling it a day.

### DIFF
--- a/assets/base_views.py
+++ b/assets/base_views.py
@@ -1,6 +1,7 @@
 from datetime import UTC, datetime
 
 from django import forms
+from django.http import HttpRequest
 from django.template.response import TemplateResponse
 
 
@@ -40,7 +41,7 @@ class ExampleForm(forms.Form):
         )
 
 
-def components(request):
+def components(request: HttpRequest) -> TemplateResponse:
     example_date = datetime.fromtimestamp(1667317153, tz=UTC)
     form = ExampleForm({"example_select": "french", "example_email": "you@example.com"})
     form.is_valid()
--- a/assets/base_views.py
+++ b/assets/base_views.py
@@ -8,11 +8,11 @@ class ExampleForm(forms.Form):
 
     example_select = forms.ChoiceField(
         choices=[
-            ["", "Please select a language"],
-            ["english", "English"],
-            ["french", "French"],
-            ["german", "German"],
-            ["spanish", "Spanish"],
+            ("", "Please select a language"),
+            ("english", "English"),
+            ("french", "French"),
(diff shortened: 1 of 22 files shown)


assets/base_views.py
```diff
@@ -8,11 +8,11 @@ class ExampleForm(forms.Form):
 
     example_select = forms.ChoiceField(
         choices=[
-            ["", "Please select a language"],
-            ["english", "English"],
-            ["french", "French"],
-            ["german", "German"],
-            ["spanish", "Spanish"],
+            ("", "Please select a language"),
```
bloodearnest, line 12: ugh, we pull this in from job-server, and we don't need it

Can we ignore assets/*? we don't ahve any code we author in there?

We could perhaps remove this and some other python files

bloodearnest (commented):
Interesting. Some of those things seem worth being strict about.

I propose ignoring old_api and assets

old_api because its a) temporary and b) is in part imported directly from release-hatch.

assets/ because any python code in there is 100% imported from job-server, which is not type checked.

Issue #137, opened 2024-03-06
Support approving/rejecting individual files


inglesp:
To ensure that every file has been reviewed by two output checkers, output checkers need to be able to record, against each file in a release request, whether they have approved or rejected the file for release.  Output checkers need to be able to change their mind, so an approved file may later be rejected (or vice versa), as long as the file has not yet been released.

#136 describes how this fits in to the full release request lifecycle.

Depends on #125 (although maybe the UI could be tackled independently).

inglesp:
Implementation notes:

To implement this, we'll need a new kind of object: a file approval belongs to a file, and records who approved/rejected a file and when.  There should only be one file approval per (file, reviewer) pair, so if a reviewer changes their mind, we should update the existing file approval.  It should only be possible to approve/reject a file of a request in the `SUBMITTED` state.

In detail...

* Rename `Status` enum to `RequestStatus`
* Create new `FileApprovalStatus` enum with members: `APPROVED` and `REJECTED`
* Update `StatusField` so that it is a base class, with two subclasses `RequestStatusField` and `FileApprovalStatusField`
* Create a `FileApproval` model in `local_db/models.py`
    * it needs the following fields:
        * `file`: a foreign key to `RequestFileMetadata`
        * `reviewer`: a text field with the reviewer's username
        * `status`: a `FileApprovalStatusField`
        * `created_at` and `updated_at`: see examples elsewhere
    * and a `unique_together` constraint on `file` and `reviewer`
* Add `approve_file` and `reject_file` methods to `LocalDBDataAccessLayer`
    * these should take a file and a user
    * they should either create a `FileApproval` for the given file and user, or update the existing one
* Add `approve_file` and `reject_file` methods to `BusinessLogicLayer` class
    * these should also take a file and a user, as well as a release request
    * they should check that the user is an output checker, and that the request's state is `SUBMITTED`
    * if the checks pass, they should call through to the corresponding methods on `LocalDBDataAccessLayer`
* Add `file_approve` and `file_reject` functions to `airlock/views/request.py`
    * these should only accept HTTP POSTs
    * these should call through to `approve_file` and `reject_file` in the business logic layer
* Add URLs for the new view functions
* Add "Approve" and "Reject" buttons to the release viewer page, but show only to output checkers and only when the request's state is `SUBMITTED`
    * If a user has already approved/rejected a file, only the button to reject/approve should be shown
* Add `get_file_approvals` method to `LocalDBDataAccessLayer`
[...]

PR #174, opened 2024-03-15
Implement file approval backend

madwort:
as per https://github.com/opensafely-core/airlock/issues/137


local_db/data_access.py
```diff
@@ -120,3 +143,37 @@ def add_file_to_request(
...
+    def reject_file(self, request_id, user, relpath):
+        with transaction.atomic():
+            request_file = RequestFileMetadata.objects.get(
+                filegroup__request_id=request_id, relpath=relpath
+            )
+
+            review, _ = FileReview.objects.get_or_create(
+                file=request_file, reviewer=user
+            )
+            review.status = FileApprovalStatus.REJECTED
```
bloodearnest, line 178: I think you also need to udpate `review.updated_at` for both these too?

bloodearnest (commented):
Look good, all the right pieces in place.

Left a few comments.

However, I am not a fan of get_file_approvals design in the ticket, and deserialisation between the DAL and the BLL.

From my comment there:

> I would much prefer it if business_logic.RequestFile grew an 
> .approvals attr, which is a list of the new 
> business_logic.FileApproval objects for that file. And that when
> get_release_request will build up the full tree of objects via its
> existing from_dict mechanism, and we can just iterate the one 
> object, rather than integrating two data sources.

When we call `bll.get_release_request(...)`, we want to get back a single object representation of *everything*, like so:

```
ReleaseRequest
  .filegroups[]
     .files[]
         .reviews[]
```

Rather than having to call .get_release_request and also then .get_file_approvals and pass both into wherever they are used, its all in the one object tree.

I cannot think of a use case for getting the approvals alone, when we don't also need everything else, but maybe I've missed that?

I think if we do this, it should solve a few issues with this PR atm:
 - removes the need for the circular FileReview.release_request reference
 - makes the from_dict implementation simpler.

airlock/business_logic.py
```diff
@@ -261,6 +266,29 @@ def store_file(release_request: ReleaseRequest, abspath: Path) -> str:
...
+    updated_at: datetime
+    release_request: ReleaseRequest
+
+    @classmethod
+    def from_dict(cls, attrs):
+        # TODO implement
+        return cls(
+            **{k: v for k, v in attrs.items() if k != "release_request"},
+            release_request=ReleaseRequest.from_dict(attrs.get("release_request", ())),
+            # files=[RequestFile.from_dict(value) for value in attrs.get("files", ())],
```
madwort, line 288: this seems unnecessary

local_db/data_access.py
```diff
@@ -120,3 +143,37 @@ def add_file_to_request(
         # Return updated FileGroups data
         metadata = self._find_metadata(request_id)
         return self._get_filegroups(metadata)
+
+    def get_file_approvals(self, request_id):
+        return [
+            self._filereview(r)
+            for r in FileReview.objects.filter(
+                file__filegroup__request_id=request_id
```
madwort, line 151: looks like I've used the implicit release_request here too

airlock/business_logic.py
```diff
@@ -266,6 +266,29 @@ def store_file(release_request: ReleaseRequest, abspath: Path) -> str:
...
+    """
+    Represents a review of a file in the context of a release request
+    """
+
+    file: RequestFile
+    reviewer: User
+    status: FileApprovalStatus
+    created_at: datetime
+    updated_at: datetime
+    release_request: ReleaseRequest
```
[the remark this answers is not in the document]
madwort: I think you're right...!

local_db/data_access.py
madwort: good spot, have set [`auto_now=True`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.DateField.auto_now) instead

bloodearnest: oh! TIL about auto_now, nice. I bet REX would like to know about this!

madwort:
> I cannot think of a use case for getting the approvals alone, when we don't also need everything else, but maybe I've missed that?

I started on this path because I wanted exactly that for the unit tests & then decided to proceed with the implementation in the original ticket. Having looked at it again, though, I think you're absolutely right so have re-worked this in the way that you suggest.

airlock/business_logic.py
```diff
@@ -41,6 +41,11 @@ class RequestFileType(Enum):
     SUPPORTING = "supporting"
 
 
+class FileApprovalStatus(Enum):
+    APPROVED = "APPROVED"
+    REJECTED = "REJECTED"
```
bloodearnest, line 46: I think we may also need an UNKNOWN status here.

When the review is created, it will be one of APPROVED or REJECTED.

But if the file is later updated, I think we want to *reset* the Review back to something neutral (we probably do not want to delete it).

Don't think we should implement that logic here, but wdyt about adding that status now?

airlock/business_logic.py
```diff
@@ -41,6 +41,11 @@ class RequestFileType(Enum):
     SUPPORTING = "supporting"
 
 
+class FileApprovalStatus(Enum):
```
bloodearnest, line 44: minor naming suggestion

```suggestion
class FileReviewStatus(Enum):
```

This reads a bit better to me? Makes the type of `FileReview.status` to be `FileReviewStatus`. 

Could even just be ReviewStatus?

bloodearnest (approved):
This is nice, thanks for reworking those bits. I think its worked out as I'd hoped :pray: 

Left a couple of suggestions, but happy to land this!

airlock/business_logic.py
madwort: Hmmm, in the `local_db` implementation, the `FileReview` is attached to `RequestFileMetadata`, which includes the `file_id` which identifies a specific version of a file. So I /think/ if the file is later updated, a review will exist for an earlier version of the file but not for the new version of the file. It might be worth pushing this logic out of the DAL & into the BLL, but I don't think that's for this PR?

airlock/business_logic.py
madwort: yeah fair point, going with `FileReviewStatus`.

airlock/business_logic.py
bloodearnest: oh right, yes, good point. yeah problem for later.

It does mean we will want to keep the old FileRequestMetadata around when a file is updated, perhaps.

PR #179, opened 2024-03-18
Far futures cacheable urls

bloodearnest:
- **Refactor the just added rendering framework.**
- **Add immutable cache ids to contents urls**

### DIFF
--- a/airlock/renderers.py
+++ b/airlock/renderers.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import csv
 from dataclasses import dataclass
 from email.utils import formatdate
@@ -7,8 +9,6 @@ from pathlib import Path
 from django.http import FileResponse
 from django.template import Template, loader
 from django.template.response import SimpleTemplateResponse
-
-from airlock.business_logic import RequestFile
 
 
 @dataclass
@@ -27,6 +27,7 @@ class RendererTemplate:
 
 @dataclass
 class Renderer:
+    MAX_AGE = 365 * 24 * 60 * 60  # 1 year
     template = None
 
     abspath: Path
@@ -66,6 +67,7 @@ class Renderer:
         return {
             "ETag": self.etag,
             "Last-Modified": self.last_modified,
+            "Cache-Control": f"max-age={self.MAX_AGE}",
         }
 
 
--- a/airlock/renderers.py
+++ b/airlock/renderers.py
@@ -1,0 +1,123 @@
+import csv
+from dataclasses import dataclass
+from email.utils import formatdate
+from functools import cached_property
+from pathlib import Path
+
+from django.http import FileResponse
+from django.template import Template, loader
+from django.template.response import SimpleTemplateResponse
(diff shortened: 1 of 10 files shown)


airlock/renderers.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @cached_property
+    def cache_id(self):
+        if self.request_file:
+            cache_id = self.request_file.file_id
+        else:
+            cache_id = filesystem_key(self.stat)
+
+        if self.template:
+            cache_id += "-" + filesystem_key(self.template_path.stat())
```
evansd, line 54: Given that this is cached, I'd be tempted to make this the hash of the template rather than its mtime. The hashing will only need to be done once per template per process, and given template files are relatively small anyway the overhead should be minimal. I recently discovered that `hashlib` has learned a new `file_digest` method which does the as efficiently as possible.

I'm always a little bit nervous of mtimes which come out of git/docker build chains as I've seen some odd things happen.

airlock/renderers.py
```diff
@@ -0,0 +1,119 @@
...
+
+from airlock.business_logic import RequestFile
+
+
+@dataclass
+class Renderer:
+    template = None
+
+    abspath: Path
+    request_file: RequestFile = None
```
evansd: My instinct here would be to factor this slightly differently and have something:
```py
abspath: Path
relpath: Path
file_cache_id:
```
i.e. pull the conditionals out into one place inside `get_renderer` and avoid the `Renderer` class knowing about `RequestFile`.

I don't have any objection to the way you've done it; only mentioning it for the sake of comparing styles.

evansd (approved):
Nice! I ended up looking at the predecessor PR to this as well, just so I could make sure I had my head round what was going on.

Made a couple of optional comments but happy to see this land.

airlock/renderers.py
bloodearnest: Hmm, git mtimes is a thing, good point. Although that's really only going to be a thing in dev, were a bit of unnecessary cache invalidation is probably ok?

I went with mtime as I wanted it to invalidate caches in dev on when we edit templates (as I don't think django will auto-reload on template changes?)

But we could use stat() mtime to invalidate the cached file_digest?  That way if git changes mtimes, we'll re-calculate the hash, but not invalidate the HTTP cache?

airlock/renderers.py
bloodearnest: Have reworked it as per your suggestion, including caching the template lookup at import time, and re-using the already resolved template for the render.

airlock/renderers.py
bloodearnest: I'm going to punt on this for now. 

The files we serve and cache locally in dev are *not* in git and should be unmodified by git commands.  Also, it fails open, so I doubt we'd even notice.

We can revisit later if there's a problem.

bloodearnest: I'm dumb, the templates are affected by git of course.

Issue #181, opened 2024-03-19
Add telemetry to airlock


bloodearnest:
Add basic django otel requests, we can even emit them in prod via our gateway on dokku4.

Be worth adding some manual measurements to the trace too, like how long it takes the tree to build, whether its htmx request or not

 * [x] - Install/configure django/requests opentelemetry in airlock
 * [x] - Add some custom measurements (tree construction, rendering)
 * [x] - Configure backend-server to emit this data to the right place in honeycomb.

bloodearnest:
[ ] - Install/configure django/requests opentelemetry in airlock
[ ] - Add some custom measurements (tree construction, rendering)
[ ] - Configure backend-server to emit this data to the right place in honeycomb.

rebkwok:
@bloodearnest  Are we done with this ticket (for now), do you think? We have custom measurements on all workspace and request views, and the tree building, but not specifically the file content rendering

bloodearnest:
Yeah, I think so.

I had a look at measuring rendering, but its messy.  Manually decorating works, like `Templater.render = instrument(Template.render)`, but it gives us a *lot* of spans, as every included/recursive template is emitted.


Ideally, we just want to know the total time spend rendering, but I'm not sure how to do that

Happy to close for now, we've got a baseline, and it'll be easy for us to add stuff

PR #182, opened 2024-03-19
Telemetry setup

rebkwok:
First part of #181

### DIFF
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -24,8 +24,6 @@ AIRLOCK_API_ENDPOINT="https://localhost:9000/api/v2"
 export OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
 # key will be for specific prod/test/dev honeycomb environment
 export OTEL_EXPORTER_OTLP_HEADERS="x-honeycomb-team=VALIDKEY1234"
-# this is the dataset that telemetry will go into in honeycomb
-export OTEL_SERVICE_NAME="airlock"
 
 # Set to True to log opentelemetry traces to the console in local env
 export OTEL_EXPORTER_CONSOLE=True
--- a/manage.py
+++ b/manage.py
@@ -4,7 +4,7 @@
 import os
 import sys
 
-import tracing
+import services.tracing as tracing
 
 
 def main():
--- a/services/tracing.py
+++ b/services/tracing.py
@@ -11,13 +11,11 @@ def get_provider():
     # https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#service
     resource = Resource.create(
         attributes={
-            "service.name": "airlock",
+            "service.name": os.environ.get("OTEL_SERVICE_NAME", "airlock"),
             "service.namespace": os.environ.get("BACKEND", "unknown"),
         }
     )
-    provider = TracerProvider(resource=resource)
-    trace.set_tracer_provider(provider)
-    return provider
+    return TracerProvider(resource=resource)
 
 
 def add_exporter(provider, exporter, processor=BatchSpanProcessor):
@@ -34,13 +32,11 @@ def add_exporter(provider, exporter, processor=BatchSpanProcessor):
     provider.add_span_processor(processor(exporter))
 
 
-def setup_default_tracing():
+def setup_default_tracing(set_global=True):
     provider = get_provider()
 
     """Inspect environment variables and set up exporters accordingly."""
     if "OTEL_EXPORTER_OTLP_HEADERS" in os.environ:
-        if "OTEL_SERVICE_NAME" not in os.environ:
-            os.environ["OTEL_SERVICE_NAME"] = "airlock"
         if "OTEL_EXPORTER_OTLP_ENDPOINT" not in os.environ:
             os.environ["OTEL_EXPORTER_OTLP_ENDPOINT"] = "https://api.honeycomb.io"
 
@@ -49,6 +45,11 @@ def setup_default_tracing():
     if "OTEL_EXPORTER_CONSOLE" in os.environ:
         add_exporter(provider, ConsoleSpanExporter())
 
+    if set_global:  # pragma: nocover
+        trace.set_tracer_provider(provider)
+
     from opentelemetry.instrumentation.auto_instrumentation import (  # noqa: F401
         sitecustomize,
     )
+
+    return provider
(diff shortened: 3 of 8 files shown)


airlock/views/workspace.py
```diff
@@ -15,6 +16,9 @@
 from .helpers import get_path_item_from_tree_or_404, get_workspace_or_raise, serve_file
 
 
+tracer = trace.get_tracer_provider().get_tracer(__name__)
```
bloodearnest, line 19: Yeah, thinking about tracers ~= logs is a reasonable default.

It basically boils down to an attribute we can filter on with the span data.

So what would be a useful filter? I am not really sure here.  But `__name__` is probably not that useful?

In jobrunner, we have 3 explicit tracers - jobs, ticks, and loops - for 3 different traces of activity.  Here, we only really have one activity, pattern, HTTP requests.  You could make an argument for this being the "airlock" tracer, perhaps?

Anyway, this is just a comment to say, I'm not 100% clear on best practice for naming your tracer objects :shrug:

airlock/views/workspace.py
```diff
@@ -43,7 +47,8 @@ def workspace_view(request, workspace_name: str, path: str = ""):
         template = "file_browser/contents.html"
         selected_only = True
 
-    tree = get_workspace_tree(workspace, path, selected_only)
+    with tracer.start_as_current_span("build_workspace_tree") as _span:
+        tree = get_workspace_tree(workspace, path, selected_only)
```
bloodearnest, line 51: What about moving this span collection *inside* get_workspace_tree? Then we get it every time it's called, rather than just this call site?

`start_as_span` can be used as a decorator on the function: https://opentelemetry.io/docs/languages/python/instrumentation/#creating-spans-with-decorators

Also, we can add some more useful information to this span too, like workspace and user

e.g. something like this

```
@tracer.start_as_current_span("get_workspace_tree")
def get_workspace_tree(workspace, path, selected_only):
    span = tracer.get_current_span()
    span.attributes.set("workspace", workspace)
    span.attributes.set("path", path)
    span.attributes.set("selected_only", selected_only)
    ...
```

Admittedly, collecting the attributes is a bit messy when using a decorator, so it may be nice to stick with the context manager inside the function. We could make a better decorator if we wanted to: https://betterprogramming.pub/using-decorators-to-instrument-python-code-with-opentelemetry-traces-d7f1c7d6f632

manage.py
```diff
@@ -4,10 +4,14 @@
 import os
 import sys
 
+import tracing
+
 
 def main():
     """Run administrative tasks."""
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "airlock.settings")
+    tracing.setup_default_tracing()
```
bloodearnest, line 13: oh, this is nice, being able to trace management commands is really nice.

tracing.py
```diff
@@ -0,0 +1,48 @@
+import os
+
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
+
+
+def get_provider():
+    provider = TracerProvider()
```
bloodearnest, line 10: I just landed [a PR](https://github.com/opensafely-core/job-runner/pull/717) to jobrunner to use a top level Resource on the provider to add some context to every span. Might be worth doing the same here.

tracing.py
```diff
@@ -0,0 +1,48 @@
...
+    provider.add_span_processor(processor(exporter))
+
+
+def setup_default_tracing():
+    provider = get_provider()
+
+    """Inspect environment variables and set up exporters accordingly."""
+    if "OTEL_EXPORTER_OTLP_HEADERS" in os.environ:
+        if "OTEL_SERVICE_NAME" not in os.environ:
+            raise Exception(
```
bloodearnest, line 35: I just changed job-runners behaviour here to default to "jobrunner" - I think we should do the same here, default to "airlock" if not set, rather than raise.

That makes it easier to share configuration in backend-server, as their default OTEL_SERVICE_NAME is correct for production.

bloodearnest (approved):

airlock/views/workspace.py
rebkwok: I basically just copied it from somewhere in job-server; originally I did just call it airlock. Maybe that's better, then

airlock/views/workspace.py
rebkwok: Sure - I only added this one so that I could check the test setup worked; hadn't really thought about where it was best to go yet.  I'll maybe leave that for a separate PR.

tracing.py
rebkwok: Yep, I'm just changing it to match your jobrunner changes

dotenv-sample
```diff
@@ -12,11 +12,24 @@ AIRLOCK_REQUEST_DIR=releases/
...
 # and you can use the default value for the endpoint.
 AIRLOCK_API_TOKEN=
 AIRLOCK_API_ENDPOINT="https://localhost:9000/api/v2"
+
+# Opentelemetry
+export OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
+# key will be for specific prod/test/dev honeycomb environment
+export OTEL_EXPORTER_OTLP_HEADERS="x-honeycomb-team=VALIDKEY1234"
+# this is the dataset that telemetry will go into in honeycomb
+export OTEL_SERVICE_NAME="airlock"
```
bloodearnest, line 28: This is now the default, so we don't need it in dotenv-sample?

services/tracing.py
```diff
@@ -0,0 +1,57 @@
...
+    # https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.export.html#opentelemetry.sdk.trace.export.BatchSpanProcessor
+    provider.add_span_processor(processor(exporter))
+
+
+def setup_default_tracing(set_global=True):
+    provider = get_provider()
+
+    """Inspect environment variables and set up exporters accordingly."""
+    if "OTEL_EXPORTER_OTLP_HEADERS" in os.environ:
+        if "OTEL_SERVICE_NAME" not in os.environ:
```
bloodearnest, line 40: This `if` is redundant now, as we do the lookup in get_provider

bloodearnest (approved):

PR #188, opened 2024-03-20
Add mypy type checking

evansd:
This is very much experimental and subject to review. The aim is to overall improve developer experience by catching more problems that it creates. We expect there to be some short term friction as we collectively get accustomed to mypy and its ways. But if longer term the balance doesn't seem to be paying off then we should be ruthless about ditching it, or relaxing the config.

In particular, Django and its metaclass magic is sometimes an awkward fit here. There are already a couple of places where we just `# type: ignore` the error and we should feel sad but not bad about adding more if we have to.

See also the discussion on an earlier spike of this work in #173

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -8,7 +8,7 @@ from pathlib import Path
 
 # we use PurePosixPath as a convenient url path representation
 from pathlib import PurePosixPath as UrlPath
-from typing import Protocol
+from typing import Protocol, cast
 
 from django.conf import settings
 from django.urls import reverse
@@ -717,5 +717,7 @@ def _get_configured_bll():
 
 
 # We follow the Django pattern of using a lazy object which configures itself on first
-# access so as to avoid reading `settings` during import
-bll = SimpleLazyObject(_get_configured_bll)
+# access so as to avoid reading `settings` during import. The `cast` here is a runtime
+# no-op, but indicates to the type-checker that this should be treated as an instance of
+# BusinessLogicLayer not SimpleLazyObject.
+bll = cast(BusinessLogicLayer, SimpleLazyObject(_get_configured_bll))
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -474,38 +474,47 @@ class BusinessLogicLayer:
         return release_request
 
     def get_current_request(
-        self, workspace_name: str, user: User, create: bool = False
+        self, workspace_name: str, user: User
     ) -> ReleaseRequest | None:
-        """Get the current request for the a workspace/user.
-
-        If create is True, create one.
-        """
+        """Get the current request for a workspace/user."""
         active_requests = self._dal.get_active_requests_for_workspace_by_user(
             workspace=workspace_name,
             username=user.username,
         )
 
         n = len(active_requests)
-        if n > 1:
-            raise Exception(
-                f"Multiple active release requests for user {user.username} in workspace {workspace_name}"
-            )
+        if n == 0:
+            return None
         elif n == 1:
             return ReleaseRequest.from_dict(active_requests[0])
-        elif create:
-            # To create a request, you must have explicit workspace permissions.
-            # Output checkers can view all workspaces, but are not allowed to
-            # create requests for all workspaces.
-            if workspace_name not in user.workspaces:
-                raise BusinessLogicLayer.RequestPermissionDenied(workspace_name)
-
-            new_request = self._dal.create_release_request(
-                workspace=workspace_name,
-                author=user.username,
-            )
-            return ReleaseRequest.from_dict(new_request)
         else:
-            return None
+            raise Exception(
+                f"Multiple active release requests for user {user.username} in "
+                f"workspace {workspace_name}"
+            )
+
+    def get_or_create_current_request(
+        self, workspace_name: str, user: User
+    ) -> ReleaseRequest:
+        """
+        Get the current request for a workspace/user, or create a new one if there is
+        none.
+        """
+        request = self.get_current_request(workspace_name, user)
+        if request is not None:
+            return request
+
+        # To create a request, you must have explicit workspace permissions.  Output
+        # checkers can view all workspaces, but are not allowed to create requests for
+        # all workspaces.
+        if workspace_name not in user.workspaces:
+            raise BusinessLogicLayer.RequestPermissionDenied(workspace_name)
+
+        new_request = self._dal.create_release_request(
+            workspace=workspace_name,
+            author=user.username,
+        )
+        return ReleaseRequest.from_dict(new_request)
 
     def get_requests_authored_by_user(self, user: User) -> list[ReleaseRequest]:
         """Get all current requests authored by user."""
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -4,10 +4,7 @@ import shutil
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum
-from pathlib import Path
-
-# we use PurePosixPath as a convenient url path representation
-from pathlib import PurePosixPath as UrlPath
+from pathlib import Path, PurePosixPath
 from typing import Protocol, cast
 
 from django.conf import settings
@@ -19,6 +16,10 @@ import old_api
 from airlock.renderers import get_renderer
 from airlock.users import User
 
+
+# We use PurePosixPath as a convenient URL path representation (we reassign rather than
+# use `import as` to satisfy mypy that we intend to export this name)
+UrlPath = PurePosixPath
 
 ROOT_PATH = UrlPath()  # empty path
 
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -8,7 +8,7 @@ from pathlib import Path
 
 # we use PurePosixPath as a convenient url path representation
 from pathlib import PurePosixPath as UrlPath
-from typing import Optional, Protocol
+from typing import Protocol
 
 from django.conf import settings
 from django.shortcuts import reverse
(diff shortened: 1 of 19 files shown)


airlock/renderers.py
```diff
@@ -14,12 +15,13 @@
...
+    path: Path
+    template: Template
 
-    def __post_init__(self):
-        self.template = loader.get_template(self.name)
-        self.path = Path(self.template.template.origin.name)
+    @classmethod
+    def from_name(cls, name: str) -> Self:
+        template = loader.get_template(name)
+        return cls(name, template=template, path=Path(template.template.origin.name))
```
inglesp: Nice improvement.

local_db/models.py
```diff
@@ -12,7 +12,9 @@ def local_request_id():
     return str(ulid())
 
 
-class EnumField(models.TextField):
+# Without the "ignore" this raises: Missing type parameters for generic type "TextField"
+# https://github.com/typeddjango/django-stubs/issues/285
+class EnumField(models.TextField):  # type: ignore
```
inglesp, line 17: Should we be documenting the `# type: ignore` comments?  This comment's quite useful but doesn't tell us anything that we wouldn't know by removing it and seeing how mypy complains.

We should be consistent.

airlock/renderers.py
```diff
@@ -40,7 +40,9 @@ class Renderer:
     def get_response(self):
         if self.template:
             context = self.context()
-            response = SimpleTemplateResponse(self.template.template, context)
+            response: HttpResponseBase = SimpleTemplateResponse(
+                self.template.template, context
+            )
         else:
             response = FileResponse(self.abspath.open("rb"), filename=self.filename)
```
inglesp, line 47: Why don't we need to do `response: HttpResponseBase = FileResponse(...)` here?  Is it because mypy's picked up the type from the first branch?

airlock/nav.py
```diff
@@ -12,7 +13,7 @@ def default_predicate(request):
 class NavItem:
     name: str
     url_name: str
-    predicate: Callable = default_predicate
+    predicate: Callable[[HttpRequest], bool] = default_predicate
```
inglesp, line 16: An interesting type signature -- but looks like `predicate` is never not the default.  @bloodearnest can we drop it?

airlock/business_logic.py
```diff
@@ -20,6 +17,10 @@
 from airlock.users import User
 
 
+# We use PurePosixPath as a convenient URL path representation (we reassign rather than
+# use `import as` to satisfy mypy that we intend to export this name)
+UrlPath = PurePosixPath
```
inglesp, line 22: Could we use [`NewType`](https://docs.python.org/3/library/typing.html#typing.NewType) here?

airlock/forms.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, *args, **kwargs):
             self.filegroup_names = set()
         group_names = sorted(self.filegroup_names - {"default"})
         group_choices = [(name, name) for name in ["default", *group_names]]
+        assert isinstance(self.fields["filegroup"], forms.ChoiceField)
```
inglesp, line 29: Is it worth commenting when we're doing type-narrowing?

airlock/file_browser_api.py
```diff
@@ -133,7 +133,7 @@ def html_classes(self):
         distinguish file/dirs, and maybe even file types, in the UI, in case we
         need to.
         """
-        classes = [self.type.value.lower()]
+        classes = [self.type.value.lower()] if self.type else []
```
inglesp, line 136: AFAICT `self.type` is only null during construction of a `PathItem` in `build_path_tree`.  Adding this check to propitiate mypy is a bit of a smell -- instead can we make `type` required, and tweak `build_path_tree`?

airlock/file_browser_api.py
```diff
@@ -320,6 +320,9 @@ def filter_files(selected, files):
             yield f
 
 
+NestedStrList = list[str] | list["NestedStrList"]
```
inglesp, line 323: This is a bit stringy -- can we make it more futurey?

inglesp (approved):
Thanks for trying this out.  A few small suggestions, but it doesn't look like typing (in this context) makes too many extra demands on the writers and readers of the code.

bloodearnest (approved):
Looks good. 

Some genuine corner cases found, and good to have accurate type info.

I'm a bit cautions about about the old_api/schema.py changes. It's possible it might change the serialization enough to break, and thus possibly not match [what job-server expects](https://github.com/opensafely-core/job-server/blob/main/jobserver/api/releases.py#L129)?

Probably not, but maybe worth a test with the local job-server?

Or, just ignore old_api/schema.py, as its imported wholesale from release-hatch?

airlock/nav.py
bloodearnest: I imported this nav code from job-server, so not 100% sure.

airlock/file_browser_api.py
bloodearnest: It might be possible. There's some awkwardness in PathItem construction, in that you have to create a parent PathItem before you create its children, so that you can pass a reference to the parent PathItem. But it may be that we can always know the type when we create the object, I'll have a look

bloodearnest: Yeah, it can work. Means branching twice, but ends up with just children being set after instance creation.
```
diff --git airlock/file_browser_api.py airlock/file_browser_api.py
index 8aa61eb..633fadf 100644
--- airlock/file_browser_api.py
+++ airlock/file_browser_api.py
@@ -34,8 +34,8 @@ class PathItem:
 
     container: AirlockContainer
     relpath: UrlPath
+    type: PathType
 
-    type: PathType = None
     children: list["PathItem"] = field(default_factory=list)
     parent: "PathItem" = None
 
@@ -346,13 +346,6 @@ def get_path_tree(
         for child, descendants in grouped.items():
             path = parent.relpath / child
             selected = path == selected_path
-            node = PathItem(
-                container=container,
-                relpath=path,
-                parent=parent,
-                selected=selected,
-                supporting_file=container.is_supporting_file(path),
-            )
 
             # If it has decendants, it is a directory. However, an empty
             # directory in workspace still needs to be classed as a PathType.DIR.
@@ -365,10 +358,22 @@ def get_path_tree(
             # is_dir(), but we are trying to avoid hitting the filesystem in
             # the tree recursion for speed.
             #
-            # We have a backstop check to not blow up in this case
+            # We have a backstop check to not blow up in this case in
             # PathItem.contents()
-            if descendants or path.suffix == "":
-                node.type = PathType.DIR
+            node_type = (
+                PathType.DIR if (descendants or path.suffix == "") else PathType.FILE
+            )
+
+            node = PathItem(
+                container=container,
+                relpath=path,
+                type=node_type,
+                parent=parent,
+                selected=selected,
+                supporting_file=container.is_supporting_file(path),
+            )
+
+            if node_type == PathType.DIR:
                 # recurse down the tree
                 node.children = build_path_tree(descendants, parent=node)
 
@@ -377,8 +382,6 @@ def get_path_tree(
                     node.expanded = True
                 else:
                     node.expanded = selected or (path in (selected_path.parents or []))
-            else:
-                node.type = PathType.FILE
 
             tree.append(node)
 
```

local_db/models.py
evansd: Yes, we should in general. Although I've got an improved workaround for this specific case.

airlock/renderers.py
evansd: Yes, in general mypy wants variables to have a single type. Without the explicit type assignment it assumes that `response` has type `SimpleTemplateResponse` which is incompatible with `FileResponse`. So we have to explicitly give it a wider type here. We could also be explicit within this branch but that would be redundant.

airlock/business_logic.py
evansd: Oh yes, good suggestion.

evansd: It turns out we can't, for reasons I don't fully understand. However I've added an alternative implementation which does effectively the same thing.

airlock/file_browser_api.py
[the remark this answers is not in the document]
evansd: We already have the future import in this file. But for this kind of construct that isn't sufficient (for some reason).

airlock/file_browser_api.py
```diff
@@ -331,7 +334,7 @@ def get_path_tree(
 
     def build_path_tree(path_parts, parent):
         # group multiple paths into groups by first part of path
-        grouped = dict()
+        grouped: dict[str, NestedStrList] = dict()
```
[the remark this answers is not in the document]
evansd: Ah, I don't think we do actually need the nesting, you're right. But the type you've given isn't quite correct either. It needs to be `dict[str, list[list[str]]`.

I think another thing that was confusing me here is that the `descendants` variable gets reused later in the function with a different type, which throws mypy off.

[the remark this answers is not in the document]
bloodearnest: Ah, yes, hmm. I guess not re-using names is good.

bloodearnest:
> > I'm a bit cautions about about the old_api/schema.py changes. It's possible it might change the serialization enough to break, and thus possibly not match [what job-server expects](https://github.com/opensafely-core/job-server/blob/main/jobserver/api/releases.py#L129)?
> 
> It might be better in that case to leave the schema as it is and change the code so it passes a datetime as specified rather than a string?

Yeah, I suspect we'll rework this when we get to building out the job-server API

evansd:
> > It might be better in that case to leave the schema as it is and change the code so it passes
> > a datetime as specified rather than a string?
>
> Yeah, I suspect we'll rework this when we get to building out the job-server API

I've ended up leaving the code exactly as it is and just ignoring this specific type error. I don't want to risk breaking code that obviously does work at the moment by fiddling the types here.

evansd:
I've rebased against latest `main` and addressed most of the comments above. A few commits worth looking at since last time:
 * https://github.com/opensafely-core/airlock/pull/188/commits/2b64ac0fb08fd1a42300e66ca23da231257e97d1
 * https://github.com/opensafely-core/airlock/pull/188/commits/d06a7d7f25bc5323bb6320f6ff5454929d40b283
 * https://github.com/opensafely-core/airlock/pull/188/commits/4b8dd743d80b4f8a69296ae0f518a70da0e28927

PR #189, opened 2024-03-21
Custom otel traces

rebkwok:
Adds an `@instrument` function decorator, following [this post](https://betterprogramming.pub/using-decorators-to-instrument-python-code-with-opentelemetry-traces-d7f1c7d6f632), plus some extra options to retrieve arg/kwarg values from the function signature, so we can add in the workspace name / request id.

Adds tracing for all request views, workspace views, `get_workspace_tree` and `get_request_tree`. Leaving the rendering for another PR.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -101,7 +101,15 @@ just assets/update
 ## Opentelemetry
 
 To log opentelemetry traces to the console in local environments,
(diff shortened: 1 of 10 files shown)


bloodearnest (approved):
This is great!

I think we should add this decorator to job-server/runner too!

PR #190, opened 2024-03-21
Clean up the otel config

bloodearnest:
The value of PYTHONPATH made no sense. And its only required to be present in the env, it can be set to nothing.  So I incorporated that in the code

Likewise, set a default DJANGO_SETTINGS_MODULE in the right place.

Also flipped OTEL_EXPORTER_CONSOLE to off by default, as its very
verbose.

### DIFF
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -21,13 +21,9 @@ AIRLOCK_API_TOKEN=
 AIRLOCK_API_ENDPOINT="https://localhost:9000/api/v2"
 
 # Opentelemetry
-export OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
-# key will be for specific prod/test/dev honeycomb environment
-export OTEL_EXPORTER_OTLP_HEADERS="x-honeycomb-team=VALIDKEY1234"
-
 # Set to True to log opentelemetry traces to the console in local env
-export OTEL_EXPORTER_CONSOLE=True
+# Warning can be verbose:
+OTEL_EXPORTER_CONSOLE=False
 
-# OTEL requires these, it needs to access settings before other stuff has initialised
-DJANGO_SETTINGS_MODULE="airlock.settings"
-PYTHONPATH="/usr/local/bin/python"
+# To send to honecomb in dev, create a token for the development and set it here.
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -120,10 +120,13 @@ fix: devenv
     $BIN/ruff --fix .
     $BIN/djhtml --tabwidth 2 airlock/
 
-
+# run airlock with django dev server
 run *ARGS: devenv
     $BIN/python manage.py runserver "$@"
 
+# run airlock with gunicorn, like in production
+gunicorn *args: devenv
```
madwort, line 128: wdyt about having them sit next to each other in the alphabetised `just default` list?

```suggestion
run-gunicorn *args: devenv
```

dotenv-sample
```diff
@@ -21,13 +21,9 @@ AIRLOCK_API_TOKEN=
 AIRLOCK_API_ENDPOINT="https://localhost:9000/api/v2"
 
 # Opentelemetry
-export OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
```
madwort, line 24: wdyt about keeping a line for `OTEL_EXPORTER_OTLP_ENDPOINT` here but with a note that it's provided by the default config in backend-server? feel like that's would document that the app requires it but our deployment provides it automatically? make it slightly more legible for someone unfamiliar with the app?

madwort (approved):
looks good, some style nits

dotenv-sample
bloodearnest: IMO, dotenv-sample is for setting up dev envs.

I don't see it as a skeleton/documentation for production config, just the minimal/useful set of things you need to run locally and may want to tweak.  It has no relation to the config in production.

Now, we may want something like that for prod, but I think dotenv-sample should be optimised for local dev use, not full belts and braces production config

madwort: fair enough

madwort: I guess I do see `dotenv-sample` as skeleton/documentation for any config prod or dev, although in practice once we've set up the deployment once we're not going to look at it again so it's not a big deal. Seems more convenient to me in general to maintain the list of all config options in there rather than somewhere else (a markdown doc file?), but we also don't really need to document every little thing right now.

PR #192, opened 2024-03-22
Refactor the instrument decorator to use inspect to find arguments

rebkwok:
By using the inspect module to identify the bound arguments passed to the function, we don't need to separate arg and kwarg attributes, and we can handle cases different types of function call, where keywords aren't enforced. We can also now handle keywords captured in a **kwargs parameter.

### DIFF
--- a/services/tracing.py
+++ b/services/tracing.py
@@ -1,5 +1,6 @@
 import os
 from functools import wraps
+from inspect import Parameter, signature
 from typing import Dict
 
 from opentelemetry import trace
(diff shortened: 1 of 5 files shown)


services/tracing.py
```diff
@@ -68,23 +69,30 @@ def instrument(
...
+    existing_tracer: pass an optional existing tracer to use. Defaults to
+      a tracer named with the value of the environment variable
+      `OTEL_SERVICE_NAME` if available, or the name of the module containing
+      the decoraated function.
     """
 
     def span_decorator(func):
-        tracer = existing_tracer or trace.get_tracer("airlock")
+        tracer = existing_tracer or trace.get_tracer(
+            os.environ.get("OTEL_SERVICE_NAME", "airlock")
```
bloodearnest, line 94: Nice touch

services/tracing.py
```diff
@@ -93,24 +101,19 @@ def _set_attributes(span, attributes_dict):
         @wraps(func)
         def wrap_with_span(*args, **kwargs):
             name = span_name or func.__qualname__
+            attributes_dict = attributes or {}
+            if func_attributes is not None:
+                func_signature = signature(func)
```
bloodearnest, line 106: Can we move this line up into the decorator body? Then its called once at import time, rather than on every func invocation?

services/tracing.py
```diff
@@ -120,3 +123,30 @@ def wrap_with_span(*args, **kwargs):
...
+    3) the parameter default value, if there is one
+    4) Finally, raises an exception if we can't find a value for the expected parameter
+    """
+    if parameter_name in func_kwargs:
+        return func_kwargs[parameter_name]
+    if parameter_name in func_bound_args:
+        return func_bound_args[parameter_name]
+    # an argument for this parameter was not passed explicitly in the function call;
+    # look for a default value
+    param = func_signature.parameters.get(parameter_name)
```
bloodearnest, line 147: Can we extract default parameters in the decorator body at import time into a dict, and then just look them up at call time?

Combined with pulling out the signature in the decorator at import time, we can build the default param dict then too.

If so, this function might be small enough to not bother with, we can just inline it in the loop, as we can just use ChainMap to look up the parameter in a series of dicts:

(Note, this also include removing the need for _set_attributes, as per my other comment)

```
...
from collections import ChainMap
...


def instrument(...):
    ...

    def span_decorator(func):
        ...
        default_args = {
             k: v.default for k, v in signature(func).parameters.items()
             if v.default != inspect.Parameter.empty,
        }
        bound_args = {}
     
        @wraps(func)
        def wrap_with_span(*args, **kwargs):
            ...
            if func_attributes is not None:
                bound_args = func_signature.bind(*args, **kwargs).arguments
  
            param_lookup = ChainMap(kwargs, bound_args, default_args)
            for k, v in func_attributes.items():
                try:
                    attributes_dict[k] = str(param_lookup[v])
                except ...:

            with tracer.start_as_current_span(
                name, record_exception=record_exception, attributes=attribute_dict,
            ):
                return func(*args, **kwargs)

        return wrap_with_span
 
```

services/tracing.py
```diff
@@ -93,24 +101,19 @@ def _set_attributes(span, attributes_dict):
...
+                func_signature = signature(func)
+                bound_args = func_signature.bind(*args, **kwargs).arguments
+                for k, v in func_attributes.items():
+                    func_arg = _get_argument_from_function_signature(
+                        func_signature, bound_args, kwargs, v
+                    )
+                    attributes_dict[k] = str(func_arg)
+
             with tracer.start_as_current_span(
                 name, record_exception=record_exception
```
bloodearnest, line 115: You can pass attributes at span creation, so we could maybe do

```
with tracer.start_as_current_span(name, record_exception=record_exception, attributes=attributes_dict):```

This would remove the need for _set_attributes alltogether?

bloodearnest (approved):
Looks good.

Have left some comments, this is a fun bit of python wrangling!

PR #193, opened 2024-03-22
create file approve & reject urls

madwort:
* I think this fixes https://github.com/opensafely-core/airlock/issues/137

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -386,6 +386,16 @@ class ReleaseRequest:
             for request_file in filegroup.output_files
         }
 
+    def get_file_review_for_reviewer(self, urlpath: UrlPath, reviewer: str):
+        return next(
+            (
+                r
+                for r in self.get_request_file(urlpath).reviews
+                if r.reviewer == reviewer
+            ),
+            None,
+        )
+
     def request_filetype(self, urlpath: UrlPath):
         try:
             return self.get_request_file(urlpath).filetype
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -10,7 +10,13 @@ from django.views.decorators.http import require_http_methods
 from django.views.decorators.vary import vary_on_headers
 from opentelemetry import trace
 
-from airlock.business_logic import RequestStatus, UrlPath, bll
+from airlock.business_logic import (
+    FileReviewStatus,
+    RequestFileType,
+    RequestStatus,
+    UrlPath,
+    bll,
+)
 from airlock.file_browser_api import get_request_tree
 from services.tracing import instrument
 
@@ -86,6 +92,38 @@ def request_view(request, request_id: str, path: str = ""):
         kwargs={"request_id": request_id},
     )
 
+    if (
+        is_directory_url
+        or release_request.request_filetype(path) == RequestFileType.SUPPORTING
+    ):
+        file_approve_url = None
+        file_reject_url = None
+    else:
+        # get_path_item_from_tree_or_404() guarantees this file exists
+        group, request_file = release_request.get_request_group_and_file(path)
+
+        file_approve_url = reverse(
+            "file_approve",
+            kwargs={"request_id": request_id, "path": path},
+        )
+        file_reject_url = reverse(
+            "file_reject",
+            kwargs={"request_id": request_id, "path": path},
+        )
+
+        existing_reviews = [
+            r
+            for r in release_request.filegroups[group]
+            .files[request_file.relpath]
+            .reviews
+            if r.reviewer == request.user.username
+        ]
+        if existing_reviews != []:
+            if existing_reviews[0].status == FileReviewStatus.APPROVED:
+                file_approve_url = None
+            else:
+                file_reject_url = None
+
     context = {
         "workspace": bll.get_workspace(release_request.workspace, request.user),
         "release_request": release_request,
@@ -96,6 +134,8 @@ def request_view(request, request_id: str, path: str = ""):
         # TODO file these in from user/models
         "is_author": is_author,
         "is_output_checker": request.user.output_checker,
+        "file_approve_url": file_approve_url,
+        "file_reject_url": file_reject_url,
         "request_submit_url": request_submit_url,
         "request_reject_url": request_reject_url,
         "release_files_url": release_files_url,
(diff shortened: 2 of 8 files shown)


airlock/views/request.py
```diff
@@ -81,6 +81,38 @@ def request_view(request, request_id: str, path: str = ""):
         "request_release_files",
         kwargs={"request_id": request_id},
     )
+
+    if path == "":
+        file_approve_url = None
+        file_reject_url = None
+    else:
```
madwort, line 88: not a big fan of this big indent block, might refactor before asking for review

rebkwok: Possible alternative: pass the file relpath as a POST parameter like we do for the `workspace_add_file_to_request` view, so the url is just `/requests/approve/<str:request_id>/`. The template logic already deals with things that are or aren't files (displaying content etc)

airlock/views/request.py
```diff
@@ -81,6 +81,36 @@ def request_view(request, request_id: str, path: str = ""):
...
+            for r in release_request.filegroups[group]
+            .files[request_file.relpath]
+            .reviews
+            if r.reviewer == request.user.username
+        ]
+        if existing_reviews != []:
+            if existing_reviews[0].status == FileReviewStatus.APPROVED:
+                file_approve_url = None
+            else:
+                file_reject_url = None
```
rebkwok, line 112: This check could move to the ReleaseRequest, and do something similar to `is_supporting_file` to get the RequestFile from the UrlPath, and look at its reviews for this user. It shouldn't need to know about the group then.
(We also don't want to show approve or reject buttons for supporting files, so it'll need to check the filetype too)

airlock/views/request.py
```diff
@@ -156,6 +188,47 @@ def request_reject(request, request_id):
...
+def file_approve(request, request_id, path: str):
+    release_request = get_release_request_or_raise(request.user, request_id)
+
+    try:
+        relpath = release_request.get_request_file(path).relpath
+    except bll.FileNotFound:
+        raise Http404()
+
+    try:
+        bll.approve_file(release_request, relpath, request.user)
```
rebkwok, line 200: This (and the reject equivalent) will need a new audit method - `bll.audit_request_file_approve` or something. The types (AuditEventType) are there already

airlock/views/request.py
```diff
@@ -156,6 +188,47 @@ def request_reject(request, request_id):
...
+
+    try:
+        bll.reject_file(release_request, relpath, request.user)
+    except bll.ApprovalPermissionDenied as exc:
+        raise PermissionDenied(str(exc))
+
+    # TODO: should this be messages.error() ?
+    messages.success(request, "File has been rejected")
+    # TODO: this should go to the previous page (file view?)
+    return redirect(release_request.get_url())
```
rebkwok, line 227: `release_request.get_url(path)` should be all it needs?

tests/integration/views/test_request.py
```diff
@@ -111,6 +113,62 @@ def test_request_view_with_authored_request_file(airlock_client):
...
+    )
+    factories.write_request_file(release_request, "group", "file.txt", "foobar")
+    response = airlock_client.get(
+        f"/requests/view/{release_request.id}/group/file.txt", follow=True
+    )
+    assert "Remove this file" not in response.rendered_content
+    assert "Approve File" in response.rendered_content
+    assert "Reject File" in response.rendered_content
+    assert '    id="file-approve-button" type="submit"' in response.rendered_content
+    assert '    id="file-reject-button" type="submit"' in response.rendered_content
```
rebkwok, line 130: Probably not worth checking for the ids here - the big functional test in `test_e2e.py` will need to use them though

rebkwok: In fact the functional test is the best place to check that the right buttons show up and are clickable or not where you expect them to be

airlock/views/request.py
madwort: indeed!

tests/integration/views/test_request.py
madwort: 👍 removed from here & added to `e2e.py`

airlock/views/request.py
madwort: did you see that the bll method [already does that internally](https://github.com/opensafely-core/airlock/blob/main/airlock/business_logic.py#L854-L868)? did you mean that we should pull it out there & call it from views/request instead?

rebkwok: Ah - no, I think I was looking at another view where it called an audit method and missed that this was wasn't needed

madwort: 👍

airlock/business_logic.py
```diff
@@ -401,6 +401,17 @@ def output_files_set(self):
             for request_file in filegroup.output_files
         }
 
+    def get_file_review_for_reviewer(self, urlpath: UrlPath, reviewer: str):
+        print(f"{urlpath} - {self.get_request_file(urlpath)}")
```
rebkwok, line 405: leftover debugging `print`?

madwort: indeed!

airlock/views/request.py
madwort: have added handling for supporting files & refactored a bit to use ReleaseRequest rather than think about groups here

airlock/business_logic.py
```diff
@@ -401,6 +401,16 @@ def output_files_set(self):
             for request_file in filegroup.output_files
         }
 
+    def get_file_review_for_reviewer(self, urlpath: UrlPath, reviewer: str):
```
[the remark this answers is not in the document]
madwort: hmm, yeah, the original ticket specified a `get_file_approvals()` like that, which I initially did, and then removed after PR review from Simon https://github.com/opensafely-core/airlock/pull/174#pullrequestreview-1943119900 in favour of keeping the interface between the BLL & DAL this one big call with the entire state of everything. 

Obviously a method that calls down to the DAL would have dodged the issue in tests with stale release_request objects, but refreshing the release_request seems ok in that situation?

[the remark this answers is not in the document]
rebkwok: I think this could be the use case Simon mentioned for just getting a file review without the rest of the release request? Although, I guess we did fetch it all already, and getting the release request already extracted all the reviews from the database, so this avoids another db call

rebkwok (approved):

PR #196, opened 2024-03-25
Add initial audit log

### LINKED ISSUE #169: Record all actions in audit log
We want to record the following kinds of event:

- Viewing a file
- Downloading a file
- Creating a request
- Changing the status of a request
- Adding a file to a request
- Changing the status of a file
- Approving/rejecting a file
- Releasing a request

This is in order to (a) help explain the history of a release request, and (b) provide support for a forensic investigation should one be necessary.

To do this, we'll record all events in a single database table with the following schema:

* event_type
* timestamp
* user
* workspace
* file (nullable)
* request (nullable)
* extra (some per-event JSON)

The DAL will be responsible for creating audit log records.  For some types of event (eg creating a request) this can happen from within an existing function.  For other types of event (eg viewing a file) the DAL will have to provide a `log_event` function (or functions) that can be called from the BLL.

We will also need functions in the DAL for retrieving all logs for a user/workspace/file/request.

bloodearnest:
Adds audit logging to airlock

Fixes #169 

This PR is a little on the large side, but its structured carefully, and all lot of the line count is because of parameter formatting, so its not as big as you might think.

It adds an audit log framework, and adds auditing for all the actions we currently support.


tests/unit/test_business_logic.py
```diff
@@ -436,15 +446,6 @@ def test_request_all_files_set(bll):
     assert len(filegroup.supporting_files) == 1
 
 
-def test_request_release_invalid_state():
```
bloodearnest, line 425: Removed because mypy catches this error now

local_db/data_access.py
```diff
@@ -188,3 +190,49 @@ def reject_file(self, request_id: str, relpath: UrlPath, username: str):
...
+        return [
+            AuditEvent(
+                type=audit.type,  # type: ignore
+                user=audit.user,
+                workspace=audit.workspace,
+                request=audit.request,
+                path=audit.path,
+                extra=audit.extra,
+            )
+            for audit in list(qs)
```
inglesp, line 268: Don't think you need to listify a queryset -- it should be iterable.

tests/local_db/test_data_access.py
```diff
@@ -0,0 +1,67 @@
...
+    ]
+
+    assert dal.get_audit_log(workspace="workspace") == [
+        other_user,
+        other_request_view,
+        request_view,
+        workspace_view,
+    ]
+
+    # excludes workspace views, as request=None
```
inglesp, line 63: I'm not surprised that getting logs for a request does not return logs for the its workspace and I found this comment more confusing than helpful.  I saw `request=None` in the comment and had to square that with `request="request"` in the line below.

local_db/data_access.py
```diff
@@ -188,3 +190,49 @@ def reject_file(self, request_id: str, relpath: UrlPath, username: str):
...
+        qs = AuditLog.objects.all().order_by("-created_at")
+
+        # TODO: we probably will need pagination?
+
+        if user:
+            qs = qs.filter(user=user)
+        if workspace:
+            qs = qs.filter(workspace=workspace)
+        if request:
+            qs = qs.filter(request=request)
```
inglesp, line 254: Does it ever make sense to ask for logs for both a workspace and a request?

tests/local_db/test_data_access.py
```diff
@@ -0,0 +1,67 @@
...
+    assert dal.get_audit_log(request="request") == [
+        other_user,
+        request_view,
+    ]
```
inglesp, line 67: Should we test filtering on multiple keys?

tests/conftest.py
```diff
@@ -26,6 +28,16 @@ def clear_all_traces():
     test_exporter.clear()
 
 
+@pytest.fixture(autouse=True)
+def clear_audit_log(request):
+    # only if we may have used the db
+    for mark in request.node.own_markers:
+        if mark.name == "django_db":
+            # truncate should be much faster than AuditLog.objects.all().delete()
```
inglesp, line 36: I know what you mean, but talking about truncating and then not calling `TRUNCATE ...` is a little bit jarring.

tests/conftest.py
```diff
@@ -26,6 +28,16 @@ def clear_all_traces():
     test_exporter.clear()
 
 
+@pytest.fixture(autouse=True)
+def clear_audit_log(request):
```
inglesp, line 32: A comment would be helpful -- I don't understand why we need to do this for audit events and not everything else.

airlock/business_logic.py
```diff
@@ -774,6 +774,50 @@ def reject_file(
...
+    ):
+        audit = AuditEvent(
+            type=AuditEventType.WORKSPACE_FILE_VIEW,
+            user=user.username,
+            workspace=workspace.name,
+        )
+        bll._dal.audit_event(audit)
+
+    def audit_request_file_access(
+        self, request: ReleaseRequest, path: UrlPath, user: User
```
inglesp, line 890: Why is `path` ignored?

airlock/business_logic.py
```diff
@@ -611,6 +611,15 @@ def get_outstanding_requests_for_review(self, user: User):
         ],
     }
 
+    STATUS_AUDIT_EVENT = {
+        RequestStatus.PENDING: AuditEventType.REQUEST_CREATE,
```
inglesp, line 667: Should we change the status to `.CREATED` for consistency?

tests/unit/test_business_logic.py
inglesp: woo

airlock/management/commands/audit.py
```diff
@@ -0,0 +1,42 @@
...
+        audit_log = bll.get_audit_log(
+            user=options["user"],
+            workspace=options["workspace"],
+            request=options["request"],
+        )
+
+        for log in audit_log:
+            if log.created_at:
+                ts = log.created_at.isoformat()[:-13]  # seconds precision
+            else:
```
inglesp, line 35: I don't see why we'd ever hit this branch

pyproject.toml
```diff
@@ -34,7 +34,11 @@ branch = true
 # Required to get full coverage when using Playwright
 concurrency = ["greenlet", "thread"]
 plugins = ["django_coverage_plugin"]
-omit = ["*/assets/*"]
+omit = [
+  "*/assets/*",
+  # remove this when we have a UI to view audit logs
+  "airlock/management/commands/audit.py",
```
inglesp, line 40: I don't think testing this would be hard!

airlock/business_logic.py
```diff
@@ -54,6 +56,62 @@ class FileReviewStatus(Enum):
...
+    REQUEST_FILE_REJECT = "REQUEST_FILE_REJECT"
+
+
+@dataclass
+class AuditEvent:
+    type: AuditEventType
+    user: str
+    workspace: str | None = None
+    request: str | None = None
+    path: str | None = None
```
inglesp, line 88: why `str` and not `UrlPath`?

inglesp (approved):
Nice clean code and a very easy to review PR, thanks.

Comments are mostly nits, but there's one fix I think we'll need to make -- one of the audit params isn't getting logged as far as I can tell.  This is probably quite easy to do, and suggests the tests need to be a bit more verbose.

local_db/data_access.py
bloodearnest: I guess not, as a request implied a workspace

tests/local_db/test_data_access.py
bloodearnest: Hah, I added it because when writing the test I made the incorrect assumption

tests/conftest.py
bloodearnest: Ah, that's how you spell TRUNCATE in sqlite.

airlock/business_logic.py
bloodearnest: Maybe -  CREATED doesn't tell you much as user though?

airlock/management/commands/audit.py
bloodearnest: It's to satisfy mypy, as created_at can be none.

I could add a comment? Or do `type: ignore`?

airlock/business_logic.py
bloodearnest: Because all of the others are simple python types, and its stored as a string in the db.

We could change it, but it just moves the conversion around into the DAL.

pyproject.toml
bloodearnest: I was concerned that diff had grown too large already, and I only really added to this to for a post-deploy check in prod.

tests/local_db/test_data_access.py
bloodearnest: I'll add an additional assert

airlock/business_logic.py
bloodearnest: oops!

evansd (approved):
Mostly echoing Peter's comments here: this was a really nice PR to review. I think there's just one place where the BLL/DAL boundary isn't consistent with what we do elsewhere (which is not to say it's either right or wrong).

Another thing that your approach here would make relatively easy it to have a test which introspects all the methods on `DataAccessLayerProtocol` and checks that they all take an argument of type `AuditEvent` (with a list of allowed exceptions). This would mean that we couldn't accidentally forget all about auditing when adding a new DAL method.

bloodearnest:
> You could also do assert log.created_at is not None which would satisfy mypy.

For some reason, I couldn't reply to this directly.

Anyway, TIL. However, since then things have developed so we need it for the management command tests.

However, this has made me question the value of having created_at=None, and have the db create it.  I think it might be simpler to have `created_at: datetime = field(default_factory=timezone.now)`, as it just simplifies a lot of stuff


However

tests/local_db/test_data_access.py
bloodearnest: removed

tests/conftest.py
bloodearnest: Have removed this commit, as in hind sight it was doing nothing.

airlock/management/commands/audit.py
bloodearnest: This branch has been refactored away now

pyproject.toml
bloodearnest: tests added

airlock/business_logic.py
bloodearnest: Have moved to UrlPath

pyproject.toml
inglesp: Fine.

local_db/data_access.py
```diff
@@ -188,3 +190,50 @@ def reject_file(self, request_id: str, relpath: UrlPath, username: str):
...
+
+        if workspace:
+            qs = qs.filter(workspace=workspace)
+        elif request:
+            qs = qs.filter(request=request)
+
+        # Note: we ignore the type here as we haven't figured out how to make
+        # EnumField type-correct yet
+        return [
+            AuditEvent(
```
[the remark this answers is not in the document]
evansd: Oh I see, yeah, that is different. OK, I'm happy to leave it as you've got it given that it's not just arbitrary variation.

evansd:
> I will try this, but there are lot of exceptions.

You could automatically exclude any methods with names starting `get_`. I don't know if that cuts it down to a sensible size?

PR #197, opened 2024-03-26
Require two approvals for all output files before releasing

### LINKED ISSUE #139: Require that all files have been approved twice before a request can be released
An output checker should not be able to release the files from a request until all non-withdrawn files have been approved by two output checkers.  The "release files" button should appear greyed out with an explanation.

#136 describes how this fits in to the full release request lifecycle.

Depends on #137 and #138.

rebkwok:
Fixes #139

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -254,6 +254,21 @@ class RequestFile:
             reviews=[FileReview.from_dict(value) for value in attrs.get("reviews", ())],
         )
 
+    def approved_for_release(self):
+        """
+        A file is approved for release if it has been approved by two reviewers
+        """
+        return (
+            len(
+                [
+                    review
+                    for review in self.reviews
+                    if review.status == FileReviewStatus.APPROVED
+                ]
+            )
+            >= 2
+        )
+
 
 @dataclass(frozen=True)
 class FileGroup:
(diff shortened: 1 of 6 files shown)


rebkwok:
Still draft because the UI for file approval isn't done yet, so if it's merged, there will be no (UI) way to release

airlock/business_logic.py
```diff
@@ -189,6 +189,21 @@ def from_dict(cls, attrs) -> Self:
...
+    def approved_for_release(self):
+        """
+        A file is approved for release if it has been approved by two reviewers
+        """
+        return (
+            len(
+                [
+                    review
+                    for review in self.reviews
+                    if review.status == FileReviewStatus.APPROVED
```
bloodearnest, line 201: I think it should probably also have no REJECTED reviews?

I guess this would only happen when three output-checker have reviewed the file?

So maybe its a corner case we don't need to worry about

bloodearnest (approved):

airlock/business_logic.py
rebkwok: I considered that, but my understanding is that it has to have 2 approvals, not that is has to have consensus among all output-checkers. Worth asking Louis, I guess

PR #198, opened 2024-03-26
Remove the index page.

bloodearnest:
Instead, redirect the workspaces index


tests/integration/views/test_workspace.py
```diff
@@ -10,11 +10,6 @@
 pytestmark = pytest.mark.django_db
 
 
-def test_index(client):
-    response = client.get("/")
-    assert "Hello World" in response.rendered_content
```
inglesp, line 15: Worth having a test that everything's wired up right?

inglesp (approved):

PR #199, opened 2024-03-27
Retrieve class instance attributes in otel decorator

rebkwok:


services/tracing.py
```diff
@@ -116,13 +118,18 @@ def wrap_with_span(*args, **kwargs):
...
-                    # 4) Finally, raises an exception if we can't find a value for the expected parameter
+                    # 4) the attribute on the class instance, if there is one
+                    # 5) Finally, raises an exception if we can't find a value for the expected parameter
                     if parameter_name in kwargs:
                         func_arg = kwargs[parameter_name]
                     elif parameter_name in bound_args:
                         func_arg = bound_args[parameter_name]
                     elif parameter_name in default_params:
                         func_arg = default_params[parameter_name]
+                    elif "self" in bound_args and hasattr(
```
bloodearnest, line 129: So, a bound method will always have the instance as the first arg, so we could we also check its the first arg?

```
...
elif bound_args.get("self") == args[0] and hasattr(args[0], parameter_name)
    func_arg = getattr(args[0], parameter_name)
else:
...
```

bloodearnest (approved):

services/tracing.py
rebkwok: This gets an index error if there are no args and no args/kwargs matching the func_attributes passed to the decorator (the first two versions of [this test](https://github.com/opensafely-core/airlock/blob/main/tests/test_tracing.py#L137) fail), so it would also have to check `len(args) > 0` or something similar. I'm not sure it gains anything by looking at the first arg.

bloodearnest: Ah, yes, sorry, red herring.

PR #200, opened 2024-03-27
Refresh authz information periodically

bloodearnest:
- **Periodically refresh the authorization information for users**
- **Extend session expiry from 2 to 8 weeks**

### DIFF
--- a/airlock/login_api.py
+++ b/airlock/login_api.py
@@ -1,4 +1,5 @@
 import json
+import time
 from pathlib import Path
 
 import requests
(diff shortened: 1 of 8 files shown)


airlock/login_api.py
```diff
@@ -43,4 +58,6 @@ def get_user_data_dev(dev_users_file: Path, user: str, token: str):
     if user not in dev_users or dev_users[user]["token"] != token:
         raise LoginError("Invalid user or token")
     else:
-        return dev_users[user]["details"]
+        details = dev_users[user]["details"]
+        details["last_refresh"] = time.time() + (365 * 24 * 60 * 60)
```
evansd, line 62: I can see why you're doing this, but worth a comment I think.

evansd (approved):
Nice.

PR #202, opened 2024-03-28
Add support for withdrawing files

bloodearnest:
Add support for withdrawing a file

If the request is still pending, the RequestFile is just deleted.

If the request is submitted, then it is marked as a new request filetype, WITHDRAWN, and shown in the UI.  The preserves and context and discussion about this file.


evansd (approved):
I really like the sequence of adding business logic, then urls/views and then UI. It made the whole thing really easy to follow.

PR #203, opened 2024-04-02
Only allow valid file types

### LINKED ISSUE #201: Ensure only valid file types are allowed
Only the valid set of file types should be:

a) shown in workspace view
b) allowed to be added to release
c) allowed to be uploaded

The currently documented list of filetypes is here: https://docs.opensafely.org/releasing-files/#allowed-file-types

The canonical source of truth in code is https://github.com/opensafely-core/pipeline/blob/main/pipeline/constants.py#L4

Note: there is potential logical distinction between Level 4 files and reviewable/releaseable file types. However, currently, they are identical lists. That may change.

This probably means adding pipeline library to airlock

rebkwok:
Fixes #201 

- Adds the pipeline library for knowledge of valid file types
- Invalid file types are:
    - not displayed in the workspace views
    - not allowed to be added to a request
    - not allowed to be released


airlock/file_browser_api.py
```diff
@@ -227,7 +229,15 @@ def get_workspace_tree(workspace, selected_path=ROOT_PATH, selected_only=False):
...
         # listing all files in one go is much faster than walking the tree
-        pathlist = [p.relative_to(root) for p in root.glob("**/*")]
+        # Note we limit the files to valid LEVEL4_FILE_TYPES here, so as to
+        # reduce some unnecessary iteration in get_path_tree. However, get_path_tree
+        # ALSO needs to check each file-type node it finds as it builds the tree for
+        # each path in the pathlist
+        pathlist = [
+            p.relative_to(root)
+            for p in root.glob("**/*")
+            if p.is_dir() or p.suffix in LEVEL4_FILE_TYPES
```
bloodearnest: Nice.

I'm wondering - should we also exclude files starting with `.` here too?

tests/unit/test_file_browser_api.py
```diff
@@ -386,7 +389,7 @@ def test_workspace_tree_contents(workspace):
     with pytest.raises(Exception):
         tree.get_path("some_dir").contents()
 
-    tree.get_path("dir.ext").contents() == "dir.ext is not a file"
+    tree.get_path("dir.txt").contents() == "dir.txt is not a file"
```
bloodearnest, line 392: oh nice, this means our empty-dir-with-extension problem is even less likely?

bloodearnest (approved):
Thanks, this gives us some confidence that we cannot repeat the mistakes of the past.

I do wonder if the fact the files are not shown at all might cause some confusion.  Maybe keeping them in the tree view, but not making them links, or just rendering "invalid file type" when clicked on might be a better UX.  Would that be a lot of work to do?

In the new stricter L4 file world, its unlikely that this would ever be the case.  But if we were to remove a valid filetype in future (e.g. svg), then there would be old files lying around perhaps.

tests/unit/test_file_browser_api.py
rebkwok: yes, seems so

rebkwok:
> I do wonder if the fact the files are not shown at all might cause some confusion. Maybe keeping them in the tree view, but not making them links, or just rendering "invalid file type" when clicked on might be a better UX. Would that be a lot of work to do?

I'll have a look at this

PR #218, opened 2024-04-04
Use simple-datatables for rendering CSV files

### LINKED ISSUE #216: Use datatables library for table rendering
We use https://github.com/fiduswriter/simple-datatables in job-server.

It can handle sorting, filtering and, most importantly, pagination (with server side pages).  

Rendering 100k+ rows in one go is a) not fun and b) not particularly useful.

rebkwok:
Fixes #216 

The template, and uses the upstream `_datatables.js` from job-server to handle the sorting/filtering/search/pagination and styling of the table.  That script adds the simple-datatables table, and replaces any server-side pagination buttons from django with new elements. All the pagination is actually done client-side. We could pass a paginator from the renderer, but the only possible benefit that I could see would be to show the correct "Page 1 of N" text instead of "Page # of #" before the js has time to replace that content.

I couldn't get the csv.html template to recognise the slippers components as valid block tags, so it does the icons and buttons longhand.

### DIFF
--- a/airlock/templates/file_browser/csv.html
+++ b/airlock/templates/file_browser/csv.html
@@ -1,3 +1,4 @@
+{% load django_vite %}
 <style>
   .datatable thead {
     position: sticky;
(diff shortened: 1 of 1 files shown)


bloodearnest (commented):
This is great, I think its a nice improvement.

There is a load time on larger tables however, with a weird FOUC - initially the spinner, then the partially styled table, then the full size.  And it always goes back to the server for all the css/js, it never uses cache, and the server never 304s. So viewing multiple tables doesn't benefit from cached js/css.

I wonder if its worth:

1. having a proper `<head>` and `<body>` sections, and manually load the css that the js currently loads in the `<head>`? I saw base, components and _datatable css. That should get the styles *and fonts* loaded before the table renders, and might prevent the FOUC?

2. It looks like whitenoise isn't caching or handling the vite assets like I would expect. We should dig into why, and get it to serve them with far future expires and 304 them if need be. That would help when looking at multiple tables, and probably in general too.

3. The dumb js I wrote to display the table when its loaded - can we hook that into an event in datatables when the table is ready? So we'd see the spinner until the table is fully ready?

4. Regards server side pagination, the benefit is not having to render the entire table into the dom. We have some very large table in Level 4 atm, e.g. ~200k rows.

bloodearnest:
> 2. It looks like whitenoise isn't caching or handling the vite assets like I would expect. We should dig into why, and get it to serve them with far future expires and 304 them if need be. That would help when looking at multiple tables, and probably in general too.

I figured out what this is. In #215, I added the sandbox attribute to the iframe. It turns out, this means it also has no access to cache!

I figured this out by viewing the table on its own with the "View" button, and observing the caching behaviour working correctly, and whitenoise doing the right thing.

Adding the [allow-same-origin sandbox permission](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#allow-same-origin) restored the desired caching behaviour.

https://github.com/opensafely-core/airlock/pull/221

bloodearnest:
> 3. The dumb js I wrote to display the table when its loaded - can we hook that into an event in datatables when the table is ready? So we'd see the spinner until the table is fully ready?

I think we can do this: https://fiduswriter.github.io/simple-datatables/documentation/Events#datatableinit

rebkwok:
So I initially started doing server-side pagination, but then I realised it wasn't actually doing anything useful, because job-server's simple-datatables script just looks for the django pagination and replaces all the buttons and pagination text. And it expects to have all the data, so all search/filter/pagination is done client-side (for the views that use it - just the staff dashboards afaict). Other paginated lists in job-server use the table component, but aren't using simple-datatables, and don't have sorting/searching/filtering (at least, not in the simple-datatables way).

I guess the question is, what's more useful? Server side pagination so that the page loads faster, or sort/search/filtering?

It looks like there has been some work to support server-side processing, but it'll involve writing some custom js, not using the upstream job-server code, and doing the search etc server-side too, I think.
https://github.com/fiduswriter/Simple-DataTables/issues/97

For now, I'll definitely look at the improvements to the initial UI on loading.

bloodearnest:
> So I initially started doing server-side pagination, but then I realised it wasn't actually doing anything useful, because job-server's simple-datatables script just looks for the django pagination and replaces all the buttons and pagination text. And it expects to have all the data, so all search/filter/pagination is done client-side (for the views that use it - just the staff dashboards afaict). Other paginated lists in job-server use the table component, but aren't using simple-datatables, and don't have sorting/searching/filtering (at least, not in the simple-datatables way).

Ah, ok, fair enough. That sounds messy.

> I guess the question is, what's more useful? Server side pagination so that the page loads faster, or sort/search/filtering?

I think the problem with come with large tables.  Can you find/generate a large csv and see the performance?

In the original outputs viewer, we would truncate the the first n rows and show a warning that the table was truncated.  We could do a similar thing here with large csvs, perhaps, which would be much simpler than a server side pagination/sorting.

I think that would align with the push from ops (Ben G particularly) to limit the row count of L4 table to much lower (e.g. 10k max).  

> It looks like there has been some work to support server-side processing, but it'll involve writing some custom js, not using the upstream job-server code,

I lean towards not using job-server code anyway, as it pulls in a lot of stuff we don't need.  But am totally fine with using it wholesale for now!

> and doing the search etc server-side too, I think. [fiduswriter/simple-datatables#97](https://github.com/fiduswriter/simple-datatables/issues/97)
> 
> For now, I'll definitely look at the improvements to the initial UI on loading.

OK, so the caching is now fixed, and server side is not really tractable.  So I am happy to +1 this. I think that showing the table on `datatable.init` will remove the FOUC, hopefully.

rebkwok:
Updated to:
1) include head/body tags and load the vite assets in the head; can't load just the css alone without changes to the upstream vite config (it didn't seem to make any difference to how the page loaded)
2) rebased to include the change to the sandbox
3) Added an extra bit of js to show the loading message until the table loads properly. I didn't do that by hooking into the datatable.init(), I did it by checking for the page number to update, which happens on datatable.init() (which means we don't need to change the upstream js, or write our own)
4) leaving server side pagination alone for now

rebkwok:
Oh, also checked that the wait-for-table-to-load code works with the old L4 chrome version

rebkwok:
So. A 200K row CSV takes about twice as long to load now, I think (about 13s for me, compared to 8ish with the old csv rendering). Once it's loaded, pagination, searching and filtering is nice. So perhaps truncating the rows is the way to go.  50K rows takes about 3s to load for me.

madwort (approved):
Looks good, filters etc seem really useful & I think performance is acceptable. On my browser/screen the contents of the iframe come out taller than the iframe itself, leading to a double-scrollbar that obscures the page count in a slightly annoying way - would be nice if we could tweak this but not a blocker for merging.

rebkwok:
Have opened #223 for the double scrollbar issue (it was already there, but the pagination has made it more obvious)

PR #220, opened 2024-04-04
c3

### LINKED ISSUE #207: Add C3 (context, controls, comments) to request and file groups
Through Airlock we want to allow researchers to provide narrative information about their request, and we want to support conversations between researchers and output checkers.

This is intended to replace the workflow that captures this information in a .docx or a Google Form.

We should add:

* a "context" textbox for the request as a whole
* textboxes for "context" and "controls" for each filegroup
* a comment thread for discussion of each textbox

The non-comment textboxes can only be created/edited by the author of the request.  It should be clear if they have been edited, and we should consider showing previous versions.

Comments can be created by any output checker or any researcher with access to the workspace.

bloodearnest:
- **Add initial C3 models and BLL/DAL methods**
- **Add views and urls for C3**
- **Add UI for C3**

Fixes #207


airlock/templates/file_browser/contents.html
```diff
@@ -1,8 +1,46 @@
...
-  {% if path_item.is_directory %}
+  {% if path_item.type.name == "FILEGROUP" %}
 
-    {% #card title=path_item.name %}
+    {% #card title=path_item.name container=True %}
+      <form action="{{ group_edit_url }}" method="POST" aria-label="group-edit-form">
+        {% csrf_token %}
+        <div style="display: flex; flex-direction: row; width: 100%; flex-wrap: wrap; gap: 1rem;">
+          <div style="display: flex: flex-direction: column; flex-basis: 100%; flex: 1;">
+            {% form_textarea field=group_edit_form.context show_placeholder=True resize=True placeholder="Describe the data to be released in this file" class="w-full max-w-lg mx-auto" rows=6 %}
```
rebkwok, line 10: "in this file" -> "in the files in this group"/"in these files"?

airlock/templates/file_browser/contents.html
```diff
@@ -1,8 +1,46 @@
...
+          </div>
+          <div style="display: flex: flex-direction: column; flex-basis: 100%; flex: 1; gap: 1rem;">
+            {% form_textarea field=group_edit_form.controls resize=True show_placeholder=True placeholder="Describe the disclosure controls that have been applied to this file" class="w-full max-w-lg mx-auto" rows=6 %}
```
rebkwok, line 13: "to this file" -> "to the files in this group"/"to these files"?

madwort (changes requested):
looks good to me, pending some of Becky's observations

airlock/templates/file_browser/contents.html
madwort: "to this group of files"?

airlock/templates/file_browser/contents.html
madwort: "to this group of files"?

airlock/views/request.py
```diff
@@ -292,3 +293,63 @@ def request_release_files(request, request_id):
...
     return redirect(release_request.get_url())
+
+
+@instrument(func_attributes={"release_request": "request_id", "group": "group"})
+@require_http_methods(["POST"])
+def group_edit(request, request_id, group):
+    release_request = get_release_request_or_raise(request.user, request_id)
+
+    form = GroupEditForm(request.POST)
+    form.is_valid()  # force validation - the form cannot fail to validate
```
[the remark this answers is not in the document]
bloodearnest: Ah, I always end up holding forms wrong, thanks

rebkwok (approved):

bloodearnest:
> Should context/controls be editable after a release has been submitted?

I think so, yes. The output checker might ask for more detail. Arguably, the author could add that info as a comment.

But they should only be editable if the request is PENDING or SUBMITTED - will fix that. We can always drop SUBMITTED later if we want to.

madwort (approved):

Issue #219, opened 2024-04-04
Files can be added to a request in the withdrawn state


madwort:
This is fine, it just makes no sense. I assume it's because we've just enumerated an enum & offered it to the user.

<img width="524" alt="Screenshot 2024-04-04 at 15 22 50" src="https://github.com/opensafely-core/airlock/assets/2082895/9bfd8b7a-4418-4c8d-b194-754281a459ec">

PR #224, opened 2024-04-05
Don't offer to add a file as withdrawn

madwort:
* fixes https://github.com/opensafely-core/airlock/issues/219

### DIFF
--- a/airlock/forms.py
+++ b/airlock/forms.py
@@ -14,7 +14,9 @@ class AddFileForm(forms.Form):
     filetype = forms.ChoiceField(
         required=True,
         choices=[
-            (i.name, i.name.title()) for i in RequestFileType if i.name != "WITHDRAWN"
+            (i.name, i.name.title())
+            for i in RequestFileType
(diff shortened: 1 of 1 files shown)


airlock/forms.py
```diff
@@ -13,7 +13,9 @@ class AddFileForm(forms.Form):
     new_filegroup = forms.CharField(required=False)
     filetype = forms.ChoiceField(
         required=True,
-        choices=[(i.name, i.name.title()) for i in RequestFileType],
+        choices=[
+            (i.name, i.name.title()) for i in RequestFileType if i.name != "WITHDRAWN"
```
rebkwok, line 17: Should probably use the enum
```
(i.name, i.name.title()) for i in RequestFileType if i != RequestFileType.WITHDRAWN
```

rebkwok (approved):

airlock/forms.py
madwort: oh yeah!

Issue #211, opened 2024-04-02
Add navigation links to other requests for a workspace


inglesp:
It will be helpful for both researchers and output checkers to have easy access to other requests for a workspace.

PR #231, opened 2024-04-11
Add navigation links to other requests for a workspace

madwort:
* fix https://github.com/opensafely-core/airlock/issues/211

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -190,7 +190,7 @@ class Workspace:
 
     def get_requests_url(self):
         return reverse(
-            "request_workspace",
+            "requests_for_workspace",
             kwargs={"workspace_name": self.name},
         )
 
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -475,6 +475,9 @@ class DataAccessLayerProtocol(Protocol):
     def get_active_requests_for_workspace_by_user(self, workspace: str, username: str):
         raise NotImplementedError()
 
+    def get_requests_for_workspace(self, workspace: str):
+        raise NotImplementedError()
+
     def get_requests_authored_by_user(self, username: str):
         raise NotImplementedError()
 
@@ -673,6 +676,19 @@ class BusinessLogicLayer:
             raise BusinessLogicLayer.RequestPermissionDenied(workspace_name)
 
         return self._create_release_request(workspace_name, user.username)
+
+    def get_requests_for_workspace(
+        self, workspace_name: str, user: User
+    ) -> list[ReleaseRequest]:
+        """Get all release requests in workspaces a user has access to."""
+
+        if not user.output_checker and workspace_name not in user.workspaces:
+            raise BusinessLogicLayer.RequestPermissionDenied(workspace_name)
+
+        return [
+            ReleaseRequest.from_dict(attrs)
+            for attrs in self._dal.get_requests_for_workspace(workspace=workspace_name)
+        ]
 
     def get_requests_authored_by_user(self, user: User) -> list[ReleaseRequest]:
         """Get all current requests authored by user."""
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -689,7 +689,9 @@ class BusinessLogicLayer:
         """Get all release requests in workspaces a user has access to."""
 
         if not user.output_checker and workspace_name not in user.workspaces:
-            raise BusinessLogicLayer.RequestPermissionDenied(workspace_name)
+            raise self.RequestPermissionDenied(
+                f"you do not have permission to view requests for {workspace_name}"
+            )
 
         return [
             ReleaseRequest.from_dict(attrs)
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -188,6 +188,12 @@ class Workspace:
             kwargs={"workspace_name": self.name, "path": relpath},
         )
 
+    def get_requests_url(self):
+        return reverse(
+            "request_workspace",
+            kwargs={"workspace_name": self.name},
+        )
+
     def get_contents_url(self, relpath, download=False):
         url = reverse(
             "workspace_contents",
--- a/airlock/templates/file_browser/index.html
+++ b/airlock/templates/file_browser/index.html
@@ -125,7 +125,7 @@
       {% elif current_request %}
         {% #button variant="success" type="link" href=current_request.get_url id="current-request-button" %}Current release request{% /button %}
       {% endif %}
-      {% #button type="link" href=workspace.get_requests_url variant="success" id="requests-workspace-button" %}All release requests for Workspace{% /button %}
+      {% #button type="link" href=workspace.get_requests_url variant="success" id="requests-workspace-button" %}All release requests for workspace{% /button %}
     </div>
   {% endfragment %}
 
--- a/airlock/templates/file_browser/index.html
+++ b/airlock/templates/file_browser/index.html
@@ -125,6 +125,7 @@
       {% elif current_request %}
         {% #button variant="success" type="link" href=current_request.get_url id="current-request-button" %}Current release request{% /button %}
       {% endif %}
+      {% #button type="link" href=workspace.get_requests_url variant="success" id="requests-workspace-button" %}All release requests for Workspace{% /button %}
     </div>
   {% endfragment %}
 
--- a/airlock/templates/requests_workspace.html
+++ b/airlock/templates/requests_workspace.html
@@ -1,0 +1,13 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+  {% #card title="All requests in workspace "|add:workspace %}
+    {% #list_group id="requests-workspace" %}
+      {% for request in requests_workspace %}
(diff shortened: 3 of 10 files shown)


airlock/business_logic.py
```diff
@@ -674,6 +677,19 @@ def get_or_create_current_request(
 
         return self._create_release_request(workspace_name, user.username)
 
+    def get_requests_for_workspace(
+        self, workspace_name: str, user: User
+    ) -> list[ReleaseRequest]:
+        """Get all release requests in workspaces a user has access to."""
+
+        if not user.output_checker and workspace_name not in user.workspaces:
+            raise BusinessLogicLayer.RequestPermissionDenied(workspace_name)
```
inglesp, line 692: Elsewhere we use `self.RequestPermissionDenied`, and usually we provide a fuller error message than just the name of the workspace.

local_db/data_access.py
```diff
@@ -128,6 +128,14 @@ def get_requests_authored_by_user(self, username: str):
             )
         ]
 
+    def get_requests_for_workspace(self, workspace: str):
+        return [
+            self._request(request)
+            for request in RequestMetadata.objects.filter(workspace=workspace).order_by(
+                "status"
```
inglesp, line 135: Would ordering by date be more useful?

tests/unit/test_business_logic.py
```diff
@@ -183,6 +183,38 @@ def test_provider_request_release_files(mock_old_api):
...
+def test_provider_get_requests_for_workspace(bll):
+    user = factories.create_user("test", ["workspace"])
+    other_user = factories.create_user("other", ["workspace"])
+    factories.create_release_request("workspace", user, id="r1")
+    factories.create_release_request("workspace", other_user, id="r2")
+
+    assert [r.id for r in bll.get_requests_for_workspace("workspace", user)] == [
+        "r1",
+        "r2",
+    ]
```
inglesp, line 195: This test would pass even if `get_requests_for_workspace` returned all requests.  I'd suggest creating a request that doesn't belong to this workspace, and ensuring that it doesn't get returned

airlock/templates/requests_workspace.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+  {% #card title="All requests in workspace "|add:workspace %}
```
inglesp, line 5: Stray space before the second `"`?

airlock/templates/requests_workspace.html
inglesp, line 1: Could we call this `workspace_requests` or `requests_for_workspace`?  Or is there a naming convention I'm missing?

airlock/views/request.py
```diff
@@ -226,6 +226,20 @@ def request_withdraw(request, request_id):
     return redirect(redirect_url)
 
 
+@instrument
+def request_workspace(request, workspace_name: str):
+    requests_workspace = bll.get_requests_for_workspace(workspace_name, request.user)
```
inglesp, line 231: I don't know what `requests_workspace` is from the name.  How about `requests` or `requests_for_workspace`?

tests/integration/views/test_request.py
```diff
@@ -443,6 +443,28 @@ def test_request_submit_not_author(airlock_client):
     assert persisted_request.status == RequestStatus.PENDING
 
 
+def test_request_workspace(airlock_client):
+    airlock_client.login(workspaces=["test1"])
+    other_user = factories.create_user("other", ["test1"], False)
```
inglesp, line 448: Why not just `user`?

tests/integration/views/test_request.py
```diff
@@ -443,6 +443,28 @@ def test_request_submit_not_author(airlock_client):
...
+
+    release_request1 = factories.create_release_request(
+        "test1", user=other_user, status=RequestStatus.PENDING
+    )
+    factories.write_request_file(release_request1, "group", "path/test.txt")
+
+    release_request2 = factories.create_release_request(
+        "test1", user=other_user, status=RequestStatus.PENDING
+    )
```
inglesp, line 457: I'm surprised that a workspace can have two pending requests for the same user.  Have I misunderstood something?

airlock/templates/file_browser/index.html
```diff
@@ -108,6 +108,7 @@
       {% elif current_request %}
         {% #button variant="success" type="link" href=current_request.get_url id="current-request-button" %}Current release request{% /button %}
       {% endif %}
+      {% #button type="link" href=workspace.get_requests_url variant="success" id="requests-workspace-button" %}All release requests for Workspace{% /button %}
```
inglesp, line 111: Are we consistent with capitalising "Workspace"?  (In general I don't think we should capitalise Nouns that happen to be the Names of Objects in our Domain...)

inglesp (approved):
Looks good -- a handful of comments, but I don't need to see this again.

airlock/templates/requests_workspace.html
madwort: I agree that the natural thing in English is to say `workspace_requests`, but it seemed that the naming convention was that the templates match the urls. I think this page belongs at `/requests/workspace/{workspace_name}` in our hierarchy, and therefore the controller method, template, etc were (approximately) `requests_workspace`.

airlock/views/request.py
madwort: hah, yeah, it's the controller for the url `/requests/workspace/`, but `requests_for_workspace` is better

tests/integration/views/test_request.py
madwort: it's to make the point that the release requests we're to be about to create in this test are from a different user than the one that we're logged in to `airlock_client` with. That could be clearer, though, I could make it that explicit (by creating a second user in this test & explicitly logging in to `airlock_client` with that user)?

madwort: also, this same pattern is used in the existing `test_request_submit_not_author(airlock_client)` - should I change that too?

tests/integration/views/test_request.py
madwort: that is a mistake - I assume that this is only possible because we've going via the factory route & this would not be possible in production. Is it worth adding a safety mechanism to `factories.create_release_request`?

inglesp: I'd strongly prefer it if it was impossible to put the system in an invalid state in tests.

A safety mechanism in the factory would be one solution to this specific problem, but better would be if test objects were created via the same route that production objects are created.  That way, any invariants that are enforced by the production code are also enforced in tests.  I've not looked into how we'd do that here but would be happy to pair.

tests/unit/test_business_logic.py
madwort: yep, done

local_db/data_access.py
madwort: probably. I was not worrying too much about this list on the basis that it would be short for now & probably need more work later to make it useful if/when it gets long (i.e. pagination)

airlock/views/request.py
madwort: oh you mean the variable not the method name? :/

airlock/templates/requests_workspace.html
madwort: using `requests_for_workspace` everywhere

airlock/views/request.py
madwort: changed both anyway

airlock/templates/requests_workspace.html
madwort: no, that's correct - if we remove the space we get output like `All requests in workspaceexample-workspace2`

airlock/business_logic.py
madwort: yes, fair enough, we also do exactly this here https://github.com/opensafely-core/airlock/blob/main/airlock/business_logic.py#L673 but will change & change the other instance in a follow-up PR

local_db/data_access.py
madwort: we can order by created_at - I would prefer to order by updated_at but I don't want to have to calculate that! https://github.com/opensafely-core/airlock/blob/main/local_db/models.py#L80-L90

tests/integration/views/test_request.py
madwort: just looking at this - I think the production objects are created via `bll.get_or_create_current_request()` which means that you simply can't create a second request on a workspace (as a given user) because you would get the first one. The test objects are created by `_create_release_request` which obviously bypasses this, but also has some extra args that let us do useful things for writing tests - i.e. creating new requests in certain states (e.g. `RequestStatus.SUBMITTED`). 

One possible refactor would be:

* `_create_current_request` becomes `_get_or_create_current_request` which gets a request if a matching one exists - we would match using the same criteria as the current `get_or_create_current_request`, i.e. just the workspace-user pair, not e.g. `status`, I would guess this might cause an existing test or two to fail
* `get_or_create_current_request` then becomes a dumb method that calls `_get_or_create_current_request` with the correct defaults

What do you think?

madwort: have amended this test to create the two release requests attached to two different users. we can look at the wider question of inviolates in testing elsewhere.

PR #235, opened 2024-04-12
Prevent release of an empty request

### LINKED ISSUE #232: Output checkers can release an empty release request with no reviews
An edge case that I noticed whilst testing the app. 

Doing the final release to job server requires two reviews (from different output checkers) for every file in the release request - however if a release request contains no files it can be "released" by a single output checker without involvement of a second output checker. At the moment this is a weird edge case but not really a problem, although it might be nice to tighten it up. 

This can be reproduced by:

* log in as researcher
* create a new release request by adding a file from a workspace
* submit the request
* withdraw the file from the request
* the request now contains an empty filegroup
* log in as output_checker
* click "Release files"

Two possible ways to resolve this:

* remove filegroups when they become empty (maybe this is not possible because we want to retain info about the withdrawn file)
* add something like this to `business_logic.py` (& fix the tests accordingly)

```python
            if (
                to_status == RequestStatus.APPROVED
                and not len(release_request.all_files_set()) > 0
            ):
                raise self.RequestPermissionDenied(
                    f"Cannot set status to {to_status.name}; request contains no files."
                )
```

madwort:
* fix #232

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -853,6 +853,14 @@ class BusinessLogicLayer:
                     f"Cannot set status to {to_status.name}; request has unapproved files."
                 )
 
+            if (
+                to_status == RequestStatus.APPROVED
+                and not len(release_request.all_files_set()) > 0
+            ):
+                raise self.RequestPermissionDenied(
+                    f"Cannot set status to {to_status.name}; request contains no files."
+                )
+
     def set_status(
         self, release_request: ReleaseRequest, to_status: RequestStatus, user: User
     ):
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -855,7 +855,7 @@ class BusinessLogicLayer:
 
             if (
                 to_status == RequestStatus.APPROVED
-                and not len(release_request.all_files_set()) > 0
+                and not release_request.output_files_set()
             ):
                 raise self.RequestPermissionDenied(
                     f"Cannot set status to {to_status.name}; request contains no files."
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -858,7 +858,7 @@ class BusinessLogicLayer:
                 and not release_request.output_files_set()
             ):
                 raise self.RequestPermissionDenied(
-                    f"Cannot set status to {to_status.name}; request contains no files."
+                    f"Cannot set status to {to_status.name}; request contains no output files."
                 )
 
     def set_status(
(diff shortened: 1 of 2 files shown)


airlock/business_logic.py
```diff
@@ -853,6 +853,14 @@ def check_status(
                     f"Cannot set status to {to_status.name}; request has unapproved files."
                 )
 
+            if (
+                to_status == RequestStatus.APPROVED
+                and not len(release_request.all_files_set()) > 0
```
rebkwok, line 858: I think this should probably check `output_files_set` rather than `all_files_set` to check for at least one output file. We don't want to allow releasing of a request that only has supporting files.

We also shouldn't need to check the file count, just the truthiness of `release_request.output_files_set()`

airlock/business_logic.py
```diff
@@ -853,6 +853,14 @@ def check_status(
...
+            ):
+                raise self.RequestPermissionDenied(
+                    f"Cannot set status to {to_status.name}; request contains no files."
```
rebkwok, line 861:
```suggestion
                    f"Cannot set status to {to_status.name}; request contains no output files."
```

tests/unit/test_business_logic.py
```diff
@@ -384,6 +384,23 @@ def test_set_status(current, future, valid_author, valid_checker, bll):
             bll.set_status(release_request1, future, user=author)
 
     if valid_checker:
+        if current == RequestStatus.SUBMITTED:
```
rebkwok, line 387: The number of `if` branches in this test make me feel like it should be split up into 2 or 3 separate ones now

rebkwok: (not necessarily for this PR)

tests/unit/test_business_logic.py
```diff
@@ -438,6 +455,16 @@ def test_set_status_cannot_action_own_request(bll):
...
 
 
+def test_set_status_approved_no_files_denied(bll):
+    user = factories.create_user("checker", [], True)
+    release_request1 = factories.create_release_request(
+        "workspace", status=RequestStatus.SUBMITTED
+    )
+
+    with pytest.raises(bll.RequestPermissionDenied):
+        bll.set_status(release_request1, RequestStatus.APPROVED, user=user)
```
rebkwok, line 465: Maybe also test that a request with only a supporting file can't be released?

airlock/business_logic.py
madwort: good points, done

tests/unit/test_business_logic.py
madwort: done

tests/unit/test_business_logic.py
madwort: yes, I think this could definitely use a refactor, I'll do in a follow-up

rebkwok (approved):

PR #236, opened 2024-04-12
Madwort/get or create tests

madwort:
* includes contents of #234


madwort:
I'm not sure this is worth pursuing

PR #239, opened 2024-04-15
Refactor file/directory detection, and improve typing

bloodearnest:
- **Refactor file/directory detection**
- **Add missing types to tree functions.**

### DIFF
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -355,28 +355,18 @@ def get_request_tree(
             expanded=selected or expanded,
         )
 
-        if selected_only:
-            if expanded:
-                if group_path == selected_path:
-                    # we just need the group's immediate child paths
-                    pathlist = [UrlPath(p.parts[0]) for p in group.files]
-                else:
-                    # filter for just the selected path and any immediate children
-                    selected_subpath = selected_path.relative_to(group_path)
-                    pathlist = list(filter_files(selected_subpath, group.files))
-            else:
-                # we don't want any children for unselected groups
-                pathlist = []
-        else:
+        if expanded or not selected_only:
+            # we want to list the children of this group,
             pathlist = list(group.files)
 
-        group_node.children = get_path_tree(
-            release_request,
-            pathlist=pathlist,
-            parent=group_node,
-            selected_path=selected_path,
-            expanded=expanded,
-        )
+            group_node.children = get_path_tree(
+                release_request,
+                pathlist=pathlist,
+                parent=group_node,
+                selected_path=selected_path,
+                expanded=expanded,
+            )
+
         root_node.children.append(group_node)
 
     return root_node
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -228,7 +228,7 @@ def scantree(root: Path) -> tuple[list[UrlPath], set[UrlPath]]:
     """
 
     paths = []
-    directories = set()
+    leaf_directories = set()
 
     def scan(current: str) -> int:
         children = 0
@@ -242,7 +242,7 @@ def scantree(root: Path) -> tuple[list[UrlPath], set[UrlPath]]:
                 if dir_children == 0:
                     # add empty dir to pathlist, or else it will not be shown
                     paths.append(path)
-                    directories.add(path)
+                    leaf_directories.add(path)
             else:
                 paths.append(path)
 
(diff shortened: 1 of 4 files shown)


airlock/file_browser_api.py
```diff
@@ -216,6 +217,44 @@ def build_string(node, indent):
...
+        for entry in os.scandir(current):
+            children += 1
+            path = UrlPath(entry.path).relative_to(root)
+
+            if entry.is_dir():
+                dir_children = scan(entry.path)
+                if dir_children == 0:
+                    # add empty dir to pathlist, or else it will not be shown
+                    paths.append(path)
+                    directories.add(path)
```
rebkwok, line 245: So `directories` here is ONLY empty directories?

airlock/file_browser_api.py
```diff
@@ -233,15 +272,21 @@ def get_workspace_tree(workspace, selected_path=ROOT_PATH, selected_only=False):
...
+        # the content area. We don't mind using stat() on the children here, as
+        # we are only loading a single directory, not an entire tree
         abspath = workspace.abspath(selected_path)
         if abspath.is_dir():
-            pathlist.extend(child.relative_to(root) for child in abspath.iterdir())
+            for child in abspath.iterdir():
+                path = child.relative_to(root)
+                pathlist.append(path)
+                if child.is_dir():
+                    directories.add(path)
```
rebkwok, line 288: here `directories` is all dir paths, whether empty or not

airlock/file_browser_api.py
```diff
@@ -233,15 +272,21 @@ def get_workspace_tree(workspace, selected_path=ROOT_PATH, selected_only=False):
...
 
     else:
-        # listing all files in one go is much faster than walking the tree
-        pathlist = [p.relative_to(root) for p in root.glob("**/*")]
+        pathlist, directories = scantree(root)
```
rebkwok, line 291: but the `directories` from `scantree` are only the empty ones

airlock/file_browser_api.py
```diff
@@ -371,21 +421,9 @@ def build_path_tree(
...
-            # However, in theory, there could be an empty directory with
-            # a suffix in its name, so this will treat these as a file rather
-            # than a directory.  If this is a problem, we could instead call
-            # is_dir(), but we are trying to avoid hitting the filesystem in
-            # the tree recursion for speed.
-            #
-            # We have a backstop check to not blow up in this case
-            # PathItem.contents()
-            if descendants or path.suffix == "":
+            if descendants or (directories and path in directories):
```
rebkwok, line 420: I guess it doesn't matter if `directories` is all directory paths or only empty ones, because the descendants check here will get the non-empty ones.  But if `scantree` added all dir paths instead of only empty ones, could we skip the descendants check here?

rebkwok (approved):
LGTM, much nicer than before.  I had some questions about what `directories` refers to in various places. I don't think it's a problem, but it might be clearer in `scantree` to call it `empty_directories`, if that's what it should be. Or just make it a list of all directories, not jus the empty ones

airlock/file_browser_api.py
bloodearnest: yes, as that's the only leaf node directory that is not a file.

airlock/file_browser_api.py
bloodearnest: good point, I will fix to be just empty, for consistency.

This part of the tree is *only* used to partially render a directory contents - not render the actually tree, so in practice, I'm not sure it makes a difference.

But I think it makes sense to be consistent :+1:

airlock/file_browser_api.py
bloodearnest: We could be it would make pathlist much larger, as it would include all the directories.

For example, in my local 45k node tree, as currently is, `len(pathlist) == 45000`. If I include directories, its 71k+.  And we don't need explicit parent directories - that information is parsed out of the paths.  So I opted just to track files, and the occasionally extra empty dir.

This aligns with the API for requests data and git repo data two, where we only have a flat list of files.  We would have to manually reconstruct pathlist entries for those if we had every directory in the pathlist.

bloodearnest:
> LGTM, much nicer than before. I had some questions about what `directories` refers to in various places. I don't think it's a problem, but it might be clearer in `scantree` to call it `empty_directories`, if that's what it should be. Or just make it a list of all directories, not jus the empty ones

Will rename to empty_directories, thanks.  I don't think it can be all directories, as that doesn't work for requests and git repo trees, as hopefully explained in a comment.

PR #240, opened 2024-04-16
Renderer refactores

bloodearnest:
In prepartion for serving git files, refactoring the renderer framework,
to more clearly separate obtaining a renderer, and then using it to
serve the file.

Specifically, rather than passing a ReleaseFile object into
`get_renderer`, and having it do the logic, we pass just the two
attributes we need (both of which are needed for git files).

Likewise, rather than have serve_file call get_renderer, we make the
caller pass the renderer in. This is much simpler, and makes the tests
for serve_file almost trivial.


evansd (approved):
Nice. It's always pleasing when the need to add new features pushes the code into a better factoring, rather than making things more complex.

bloodearnest:
> Nice. It's always pleasing when the need to add new features pushes the code into a better factoring, rather than making things more complex.

Yeah, its also sign that you've likely found the right abstraction seam.

PR #241, opened 2024-04-17
Make airlock docs viewable in L4 and in main docs

### LINKED ISSUE #206: Make Airlock docs viewable in L4
To support viewing outputs and creating/reviewing/releasing requests inside L4, we should ensure that all Airlock documentation is available within Airlock itself.

It may be worth cribbing from ehrQL's approach, where the ehrQL docs belong to the ehrQL repo but are presented as part of the main OpenSAFELY documentation.

rebkwok:
Fixes #206 

Uses a similar process to ehrQL to maintain all Airlock docs within the Airlock repo, and build them as part of the main docs site. Requires a [minor update](https://github.com/opensafely/documentation/pull/1484) to the documentation repo (PR still draft as it's pointing at this branch).

We add the built docs directory to STATICFILES_DIR, and use a minimal wrapper around django's static `serve` view to serve them at /docs, so they can be available locally to airlock wherever airlock is running.

Also involved pulling the Airlock-related docs out of the team manual (and updating them a bit).  It includes C3 because that's almost merged.


rebkwok:
We'll need to give this repo access to some additional secrets (cloudflare account id and token for the pages deployment, and a write token for the main docs repo so that it can trigger the build there)

docs/reviewing.md
```diff
@@ -0,0 +1,36 @@
...
+!!!warning "Retrictions for authors"
+    Airlock does not permit users to approve or reject files that are part of release requests they created themselves.
+
+
+## Viewing release requests on Airlock
+
+1. Output checkers navigate to the Requests list to view all open requests
+    awaiting review. Note that if an output checker is the author of a request, they
+    will not see it in the outstanding requests list. 
+1. Output checkers identify the request by workspace name and requester username.
```
LFISHER7, line 16: Will there only ever be one active request per workspace? Or is it one active request per user in a given workspace? Maybe worth mentioning here

docs/reviewing.md
```diff
@@ -0,0 +1,36 @@
...
+1. The files included in the request can be viewed in the file
+    browser within Airlock. If required (e.g. in order to perform calculations
+    in a spreadsheet), files can also be downloaded from Airlock. Please note
```
LFISHER7, line 19:
```suggestion
    in a spreadsheet), output checkers can also download files from Airlock. Please note
```

LFISHER7: Just to make it a bit clearer this is only for output checking, not for viewing outputs

LFISHER7 (commented):
I've reviewed the docs content and it looks good. Just a couple of small questions and a minor suggestion

docs/index.md
```diff
@@ -1 +1,43 @@
...
+    the TPP backend, go to `https://tpp.backends.opensafely.org`.
+1. Log in using your GitHub username or email and the single use token. 
+
+(Note: Airlock is not supported on browsers other than Chrome.)
+
+
+## Viewing files
+
+Users can view medium privacy outputs from any workspace they have permission to
+access (workspaces for which they have the ProjectCollaborator role) via Airlock if:
```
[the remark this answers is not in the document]
rebkwok: Yes, will update.
I took this from this bit of the [current team manual docs](https://bennettinstitute-team-manual.pages.dev/products/output-checking/#__tabbed_8_1), but now I realise that that's ambiguous and probably is meant to refer to viewing _released_ outputs on job-server outside of L4 (and not unreleased outputs inside L4).

docs/reviewing.md
rebkwok: There will only every be one active request _per user_ per workspace. It's unlikely, but possible, for different users to have a release for the same workspace.  I'll add a clarification here.

docs/reviewing.md
rebkwok: I've rewritten this last point in the same style as the previous ones, and added a bit more clarification about what the download option is intended for

airlock/settings.py
```diff
@@ -209,16 +210,22 @@ def run_connection_init_queries(*, connection, **kwargs):
 
 USE_TZ = True
 
+# Get the location of the built docs from the mkdocs config
+# The build docs dir is added to STATICFILE_DIRS so that we
+# can serve it within airlock
+with (BASE_DIR / "mkdocs.yml").open() as mkdocs_config:
+    DOCS_SITE_DIR = yaml.load(mkdocs_config, Loader=yaml.Loader)["site_dir"]
```
evansd, line 217: Not suggesting this needs changing but just mentioning an alternative approach here which I've only recently started using: you could hardcode the value here to just:
```py
DOCS_DIR = BASE_DIR / "docs"
```

Then you have tiny test (maybe added to `tests/integration/test_settings.py`) which loads and parses `mkdocs.yaml` and checks that the configuration matches what's in settings.

This prevents us changing the config in one place and forgetting to change it in another, but without introducing any additional runtime complexity (admittedly not much in the case, but still it would be nicer not to have to load and parse a yaml file in settings.)

Obviously the disadvantage is that it doesn't _automatically_ stay in sync, but for values which we don't expect to change very frequently I don't think that matters too much.

I mention this more for the sake of spreading the pattern around than because I think it's particularly important to apply it in this cas.

mkdocs.yml
```diff
@@ -37,9 +37,13 @@ theme:
     - content.code.copy
     - navigation.footer
     - navigation.indexes
+  # Note: specific to in-airlock docs (main docs have different template overrides);
+  # adds the link back to airlock
+  custom_dir: docs/overrides
 
 extra_css:
   - stylesheets/extra.css
+  - stylesheets/airlock.css  # Note: sextra styling pecific to in-airlock docs
```
evansd, line 50: Please do not change this typo :joy:

evansd (approved):
This is great. Having docs in the same repo really improves our changes of keeping things up-to-date. And having them available inside Airlock itself will be a massive usability win I suspect.

airlock/settings.py
rebkwok: Ah - I had it hard coded before (it would be `BASE_DIR` / "mkdocs_build") but changed it to this...I'd forgotten about the settings test

mkdocs.yml
rebkwok: I have fixed it. Sorry :laughing:

docs/reviewing.md
```diff
@@ -0,0 +1,41 @@
...
+* **Approve** — output meets disclosure requirements and is safe to be released
+* **Reject** — output is not currently acceptable for release. 
+
+!!!warning "Retrictions for authors"
+    Airlock does not permit users to approve or reject files that are part of release requests they created themselves.
+
+
+## Viewing release requests on Airlock
+
+1. Output checkers navigate to the Requests list to view all open requests
```
rose-higgins, line 13: > **After logging into Airlock, o**utput checkers navigate to the Requests

Suggest adding the bold for completeness of instructions

PR #255, opened 2024-04-18
Improve SQLite configuration

### LINKED ISSUE #248: Backport SQLite improvements from unreleased Django 5.1
In particular we want the "transaction mode immediate" support:
 * https://github.com/django/django/pull/17760

And the `init_command` support:
 * https://github.com/django/django/pull/14824

The latter is currently handled by our custom `CONNECTION_INIT_QUERIES` option but it would be good to replace this with the upcoming standard mechanism.

There's some useful information in this blogpost:
https://kerkour.com/sqlite-for-servers

(Note: the author has some weird anti-scraping measures in place so you may need to access this page from the Feb 2024 section of this [list of posts](https://kerkour.com/tags/programming).)

Slack thread:
https://bennettoxford.slack.com/archives/C01T2HACV3K/p1711836958649859

evansd:
This involves backporting some SQLite improvements from unreleased Django 5.1 and then using them to configure SQLite more gooder.

Closes #248

### DIFF
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -165,7 +165,25 @@ DATABASES = {
         "ENGINE": "django_sqlite3_backport",
         "NAME": WORK_DIR / "db.sqlite3",
         "OPTIONS": {
-            "init_command": "PRAGMA journal_mode=wal;",
+            # For details on these pragmas see: https://www.sqlite.org/pragma.html
+            "init_command": """
+                /* These settings give much better write performance than the default
+                   without sacrificing consistency guarantees */
+                PRAGMA journal_mode = WAL;
+                PRAGMA synchronous = NORMAL;
+
+                /* Enforce foreign keys which SQLite doesn't do by default only for
+                   backwards compatibility reasons */
+                PRAGMA foreign_keys = ON;
+
+                /* How long (in ms) to let one write transaction wait for another */
+                PRAGMA busy_timeout = 5000;
+
+                /* The default cache size is 2MB but we can afford more! Note negative
+                   values set cache size in KB, positive numbers set it by number of
+                   database pages */
+                PRAGMA cache_size = -256000;
+            """,
         },
     }
 }
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -184,6 +184,12 @@ DATABASES = {
                    database pages */
                 PRAGMA cache_size = -256000;
             """,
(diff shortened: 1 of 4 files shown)


django_sqlite3_backport/base.py
```diff
@@ -0,0 +1,66 @@
...
+        conn = super().get_new_connection(conn_params)
+        for init_command in self.init_commands:
+            if init_command := init_command.strip():
+                conn.execute(init_command)
+        return conn
+
+
+if django.VERSION[:2] < (5, 1):
+    DatabaseWrapper = ExtendedDatabaseWrapper
+else:  # pragma: no cover
```
inglesp, line 90: Why not assert on the version?  (Or are warnings escalated to errors?)

airlock/settings.py
```diff
@@ -165,7 +165,25 @@ def get_env_var(name):
...
+                /* Enforce foreign keys which SQLite doesn't do by default only for
+                   backwards compatibility reasons */
+                PRAGMA foreign_keys = ON;
+
+                /* How long (in ms) to let one write transaction wait for another */
+                PRAGMA busy_timeout = 5000;
+
+                /* The default cache size is 2MB but we can afford more! Note negative
+                   values set cache size in KB, positive numbers set it by number of
+                   database pages */
```
inglesp, line 184: wat

inglesp (approved):
Nice to have the breadcrumbs about where these changes came from.

airlock/settings.py
evansd: IKR? https://www.sqlite.org/pragma.html#pragma_cache_size

It must be that SQLite originally only supported setting the cache size in terms of number of pages (which is how it is actually implemented internally) and then later, as a convenience, wanted to allow setting it in terms of memory used. And this was the backwards compatibility trick they come up with. It's not _particularly_ lovely, is it?

bloodearnest (approved):
LGTM.

Is it worth adding an explicit [https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-DATABASE-ATOMIC_REQUESTS](`ATOMIC_REQUESTS=False`) in settings.py and a comment?

AIUI, atomic requests would work badly with BEGIN IMMEDIATE?

evansd:
@bloodearnest Yeah, good suggestion. That's my understanding too. I'll add the config and the comment.

PR #257, opened 2024-04-18
Do not allow rejection or withdrawal of an approved release

madwort:
* In the current process, an output checker does one action which sets the status
  to APPROVED, and then immediately does the release & sets the status to RELEASED
* As such, a release request is only APPROVED if an external error has occurred, or the
  release is currently in progress. It does not make sense to allow rejection or
  withdrawal in either of these situations.


tests/unit/test_business_logic.py
```diff
@@ -343,11 +343,11 @@ def test_provider_get_current_request_for_user_output_checker(bll):
         (RequestStatus.SUBMITTED, RequestStatus.APPROVED, False, True),
         (RequestStatus.SUBMITTED, RequestStatus.REJECTED, False, True),
         (RequestStatus.SUBMITTED, RequestStatus.WITHDRAWN, True, False),
-        (RequestStatus.SUBMITTED, RequestStatus.PENDING, True, False),
+        (RequestStatus.SUBMITTED, RequestStatus.PENDING, False, False),
```
madwort, line 346: keep this one
```suggestion
        (RequestStatus.SUBMITTED, RequestStatus.PENDING, True, False),
```

PR #261, opened 2024-04-19
Add file sizes to the list view for directories

madwort:
Screenshot 2024-04-19 at 16 28 59

### DIFF
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -62,6 +62,8 @@ class PathItem:
     # but this allow it to be overridden.
     display_text: str | None = None
 
+    size: int | None = None
+
     def is_directory(self):
         """Does this contain other things?"""
         return self.type != PathType.FILE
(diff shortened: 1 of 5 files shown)


bloodearnest (commented):
Hmm, I get that we only want to stat a directories children, but decorating the sizes in the view seems awkward to me.

I think we could defer this to the container, which is the pattern we've generally used to do this kind of specialisation. Make it a AirlockContiner function, expose a parameterless version via PathItem and then do it on demand when rendering the template.

So, add a method `get_metadata(relpath: UrlPath)` to `AirlockContainer`.

Then add a wrapper method to PathItem:

```
def get_metadata(self):
    return self.container.get_metadata(self.relpath)
```

For workspace and request, the implementation would look something like:
```
def get_metadata(relpath):
    stat = self.abspath(relpath).stat()

    return {
        "timestamp": stat.st_mtime,  # or convert to datetime
        "size": stat.st_size,  # convert to Mb?
    }
```

Then, the stat is performed on demand when we request it in the template, when rendering the directory contents.

```
{{ path_item.get_metadata.size }}
```

This way, the specialization stays inside the container, and there's a uniform template friend way to get it on demand?

Obviously, there's some additional error handling - we'd only want to do this on PathType.FILE, for example.

bloodearnest:
If we do this, it might have some impact on this work: https://github.com/opensafely-core/job-runner/issues/701#issuecomment-2071807529

madwort:
think we're going to do this differently using manifest.json

PR #288, opened 2024-04-25
Minimal viable spinner for request release

### LINKED ISSUE #270: Show progress message/spinner when files are releasing
Feedback from @milanwiedemann ([slack thread](https://bennettoxford.slack.com/archives/C01UJLWNZJT/p1712917633599879?thread_ts=1712753568.007979&cid=C01UJLWNZJT)):
> for a brief moment after clicking the "Release files" button I didn't know if anythig is happening (this is probalbly because there were so many files in this request). after a couple seconds I started getting notifications from [#opensafely-releases](https://bennettoxford.slack.com/archives/C03KD4LURV4) in slack that the files are being released ... only after ALL files were released I got a green msg on the airlock page that "Files have been released to [jobs.opensafely.org](http://jobs.opensafely.org/)" ... I was thinking that  maybe another msg right after clicking the "Release files" button with a msg like "Files are currently being released to [jobs.opensafely.org](http://jobs.opensafely.org/)" followed by "Files have been successfully released to [jobs.opensafely.org](http://jobs.opensafely.org/)" would be helpful?

madwort:
* disable the "release" button, although don't gray it out
* spinner takes up space in the layout even when not visible, but I'm not keen to get into overlaying it over the button text with CSS
* return 502 for job server debug errors so that htmx does the right thing & replaces the whole  page rather than putting the whole debug page where the button was
* use a different mechanism for redirects when htmx as `redirect()` doesn't (it replaces the whole page with the body content of the redirect response, i.e. blank) - we may need to put this into a helper method & use elsewhere
* fixes #270


airlock/views/request.py
```diff
@@ -364,14 +365,18 @@ def request_release_files(request, request_id):
                     "response": err.response,
                     "type": err.response.headers["Content-Type"],
                 },
+                status=502,
```
madwort, line 368: open to suggestions on which status code to use here, although I think I'm looking for  4xx or 5xx

bloodearnest: how about `status=err.status_code`? That should pass through the error from job-server API, which will be 4xx or 5xx (in order to have raised HTTPError?)

bloodearnest (approved):
This is nice, thanks.

HTMX is fun!

airlock/views/request.py
madwort: yeah, fair enough, in my local dev env it returns `404` (as the workspace doesn't exist on prod job-server) and & in tests it returns `403`, that should work.

PR #290, opened 2024-04-25
Remove Code as a selectable filetype when adding a file

rebkwok:


airlock/forms.py
```diff
@@ -16,7 +16,7 @@ class AddFileForm(forms.Form):
         choices=[
             (i.name, i.name.title())
             for i in RequestFileType
-            if i != RequestFileType.WITHDRAWN
+            if i not in [RequestFileType.WITHDRAWN, RequestFileType.CODE]
```
madwort, line 19: should we just allow-list?

```suggestion
            if i in [RequestFileType.OUTPUT, RequestFileType.SUPPORTING]
```

tests/unit/test_forms.py
```diff
@@ -12,6 +12,14 @@ def test_add_file_form_no_release_request():
     assert form.fields["filegroup"].choices == [("default", "default")]
 
 
+def test_add_file_form_file_type_choices():
+    form = AddFileForm(release_request=None)
+    assert form.fields["filetype"].choices == [
```
madwort, line 17: thanks for adding a test!

airlock/forms.py
rebkwok: Maybe? I thought about it, wasn't sure if there could ever be another type in future. There's a test now at least, if we did this again

madwort: 🤷 I wouldn't have expected CODE to be added tbh, but not really fussed.

madwort (approved):

PR #296, opened 2024-04-26
Add navigable summary pages

### LINKED ISSUE #245: Show a count of supporting files in the request view
https://bennettoxford.slack.com/archives/C065F8P58AW/p1712755051357289?thread_ts=1712753890.166379&cid=C065F8P58AW

> In the review request view, the number of output files is shown. Could the number of supporting files also be shown. Having this was useful because I knew how many files I wanted and this allowed me to check I hadn't missed any.

### LINKED ISSUE #264: Root summary page
Each kind of tree we view in Airlock has some overview/summary information.  For ReleaseRequests in  particular, there is some high level information needed to be displayed and edited:
 - current status
 - how many in request files/what type
 - whole request editable context the author can supply
 - whole request comments?
 - request activity (audit log)

For workspaces and repos, there may be similar information, but probably read-only.

We can show this information when first loading the relevant tree page - this is basically the root page. But we need some place to be able to navigate back to this overview.  We may want to have final whole request action buttons (approve/reject/withdrawn) only available from this page, rather than on every page, perhaps.

The obvious solution is have a top level item in the nav tree, that when clicked on takes you to this content.  We already have this root node in the tree data structure, we just currently don't render it. We'd also need to specialise the rendering of this root node to be different for workspace/requests/repos too.

bloodearnest:
This adds initial summary page content when rendering the root node of the tree.

This provides specific initial overview content for Workspaces and
ReleaseRequests. Both have an initial basic implementation of recent activity
from the audit log.  ReleaseRequest has some additional things from user
feedback, and is where we can add request level Context/comments down the line.

Fixes #264
Fixes #245


madwort:
The HTMX version of e.g. http://localhost:8000/workspaces/view/example-workspace/ is raising an exception for me

airlock/business_logic.py
```diff
@@ -535,6 +557,9 @@ def output_files_set(self):
             for request_file in filegroup.output_files
         }
 
+    def supporting_files_count(self):
+        return len(self.all_files_set()) - len(self.output_files_set())
```
madwort, line 561: will this count `RequestFileType.WITHDRAWN` files as supporting files?

airlock/business_logic.py
```diff
@@ -84,6 +84,25 @@ class AuditEventType(Enum):
...
+    AuditEventType.WORKSPACE_FILE_VIEW: "Viewed file",
+    AuditEventType.REQUEST_FILE_VIEW: "Viewed file",
+    AuditEventType.REQUEST_FILE_DOWNLOAD: "Downloaded file",
+    AuditEventType.REQUEST_CREATE: "Created request",
+    AuditEventType.REQUEST_SUBMIT: "Submitted request",
+    AuditEventType.REQUEST_WITHDRAW: "Withdrew request",
+    AuditEventType.REQUEST_APPROVE: "Approved request",
+    AuditEventType.REQUEST_REJECT: "Rejected request",
+    AuditEventType.REQUEST_RELEASE: "Released request",
+    AuditEventType.REQUEST_EDIT: "Edited the Context/Controls",
```
madwort, line 97:
```suggestion
    AuditEventType.REQUEST_EDIT: "Edited the group Context/Controls",
```

airlock/templates/activity.html
```diff
@@ -0,0 +1,25 @@
...
+    {% /table_head %}
+    {% #table_body %}
+      {% for log in activity %}
+        {% #table_row %}
+          {% #table_cell %}{{ log.created_at|date:'Y-m-d H:i' }}{% /table_cell %}
+          {% #table_cell %}{{ log.user }}{% /table_cell %}
+          {% #table_cell %}{{ log.description }}{% /table_cell %}
+          {% #table_cell %}
+            {% if log.path %}{{ log.path }}{% endif %}
+            {% if 'group' in log.extra %}(group: {{ log.extra.group }}){% endif %}
```
madwort, line 19: wdyt about giving group its own column? I realise my local dev audit log isn't going to be representative, but a lot of the events have a group, so my guess is that it would be tidier

Screenshot 2024-04-26 at 17 44 48

airlock/file_browser_api.py
```diff
@@ -272,11 +286,14 @@ def get_workspace_tree(
     selected_path is a directory, its contents can be partially rendered.
     """
 
+    breakpoint()
```
madwort, line 289: typo?
```suggestion
```

bloodearnest:
> The HTMX version of e.g. http://localhost:8000/workspaces/view/example-workspace/ is raising an exception for me
> 
> ```
> ValueError at /workspaces/view/example-workspace/
> 
> not enough values to unpack (expected at least 1, got 0)
> ```
> 
> ```
> Exception Location: 	/Users/tom/Documents/bennett-institute/airlock/airlock/file_browser_api.py, line 458, in build_path_tree
> ```

Good spot. We'd never been able to load the root over HTMX previously, and there was a bug in the partial tree loading for HTMX usage. Fixed in #7cb13cc

airlock/templates/activity.html
bloodearnest: Yeah, I tried that, it's messy.

For example many workspace audit logs (high volume) have no group

I think we will need to probably specialise these activity logs for each type thing you're looking at.

airlock/business_logic.py
bloodearnest: That is the longest string. While your version is clearer, it adds 6 characters to the table column width, and we're tight on horizontal space. We have group in the Context column anyway, I think that's maybe enough?

airlock/business_logic.py
bloodearnest: oh, yes, it will, will fix thanks

airlock/business_logic.py
madwort: fair point - wdyt about going the other way & dropping `verb group` from the others?

airlock/templates/activity.html
madwort: ok, it's a nit we always pick at later if we want

bloodearnest: Yeah, I just got the most basic version in, mainly so I had *something* to show in the workspace summary!

airlock/business_logic.py
bloodearnest: done

bloodearnest:
I have tweaked the UI of Recent Activity a bit:

 - do only show mutations (read only events are not interesting for users, just operators)
 - show all extra values in an audit log
 - make it a datatable, so we have nice sorting and filtering.

airlock/templates/file_browser/csv.html
```diff
@@ -27,19 +28,11 @@
...
-                  </svg>
-                  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" class="h4 w-4 datatable-icon--descending" role="presentation">
-                    <path fill-rule="evenodd" d="M10 3a.75.75 0 01.75.75v10.638l3.96-4.158a.75.75 0 111.08 1.04l-5.25 5.5a.75.75 0 01-1.08 0l-5.25-5.5a.75.75 0 111.08-1.04l3.96 4.158V3.75A.75.75 0 0110 3z" clip-rule="evenodd" />
-                  </svg>
-                </span>
-
+              <th>
+                <div class="flex flex-row gap-2">
+                  {{ header }}
+                  {% datatable_sort_icon %}
```
rebkwok, line 34: Is there a reason for putting this in a template tag and not just as an included html snippet?

rebkwok: Hmm, it is used multiple times isn't it? Maybe I've answered my own question

airlock/templates/activity.html
```diff
@@ -0,0 +1,45 @@
...
+
+
+{% #card title="Recent Activity" container=True %}
+  <table class="datatable" id="customTable">
+    <thead>
+      <tr>
+        <th><div class="flex flex-row gap-2">Time{% datatable_sort_icon %}</div></th>
+        <th><div class="flex flex-row gap-2">User{% datatable_sort_icon %}</div></th>
+        <th><div class="flex flex-row gap-2">Action{% datatable_sort_icon %}</div></th>
+        <th><div class="flex flex-row gap-2">Context{% datatable_sort_icon %}</div></th>
```
rebkwok, line 21: Is there some other header we can use other than "Context", given that we have a Context already for filegroups, and this isn't that. It confused me initially when i looked at the activity table locally.

rebkwok: I spent a while trying and failing to think of a different word; chatgpt suggests "Target"

rebkwok (approved):

airlock/templates/file_browser/csv.html
bloodearnest: Yeah, I didn't fancy copy pasting it.

I actually think the js should inject it, ideally, so that if the js fails, we're not left with these left over non-active icons.

airlock/templates/activity.html
bloodearnest: "Details"?

madwort: +1 for "details"

PR #297, opened 2024-04-26
Expand all files in the request tree by default.

bloodearnest:
We rarely expect >20 files here. Expanding all by default saves the user
clicks, and gives them  a) full view of the status of every file and b)
one click navigation to every file.

They can always collapse if they need to.


madwort (approved):
yes please

PR #298, opened 2024-04-26
Notifications

### LINKED ISSUE #256: Notifying users of events in Airlock
See implementation notes [here](https://docs.google.com/document/d/1MOOY13qpDJ6eBHS2VHtl7UezHkUl0RxuLNj2KfltwTs/edit).

rebkwok:
Fixes #256 

Sends notifications  on:
- set status (on moving to any status except pending)
- edit group (editing context/controls)
- add file
- withdraw file
- add comment
For updates (i.e. the last 4 above), notifications are only sent for requests in non-pending state.

Currently update events are not batched, with the exception of context and controls editing (because those are part of the same form). The DAL now returns a list of NotificationUpdateTypes on group_edit to indicate which fields have changed and need to be notified.

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1281,16 +1281,6 @@ class BusinessLogicLayer:
             release_request.id, group, comment, user.username, audit
         )
 
-        if release_request.status != RequestStatus.PENDING:
-            updates = [
-                self._get_notification_update_dict(
-                    NotificationUpdateType.COMMENT_ADDED, group, user
-                )
-            ]
-            self.send_notification(
-                release_request, NotificationEventType.REQUEST_UPDATED, user, updates
-            )
-
     def get_audit_log(
         self,
         user: str | None = None,
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -33,7 +33,7 @@ from airlock.utils import is_valid_file_type
 
 ROOT_PATH = UrlPath()  # empty path
 
-log = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
 
 class RequestStatus(Enum):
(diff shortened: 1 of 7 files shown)


airlock/notifications.py
```diff
@@ -0,0 +1,33 @@
...
+
+session = requests.Session()
+
+
+def send_notification_event(event_json: str, username: str):
+    """API call to job server to create a release."""
+    if not settings.AIRLOCK_API_TOKEN:
+        # Skip attempting to send notifications when we're running a
+        # local dev server in isolation
+        return {"status": "ok"}
```
evansd, line 13: Is it worth logging these (assuming that the logs appear in the console) so that it's easy to see what would have been sent when running locally?

airlock/business_logic.py
```diff
@@ -1215,6 +1279,16 @@ def group_comment(
 
         bll._dal.group_comment(release_request.id, group, comment, user.username, audit)
 
+        if release_request.status != RequestStatus.PENDING:
```
bloodearnest, line 1282: I think we might want to nix this for now.

It looks like output-checker comments may need to be private until both checkers have completed their review, and then confirmed to send to comments to user.  So I think we should probably punt on the notification of comments for now, until that flow is clearer?

airlock/business_logic.py
```diff
@@ -1227,7 +1247,19 @@ def group_edit(
             extra={"context": context, "controls": controls},
         )
 
-        bll._dal.group_edit(release_request.id, group, context, controls, audit)
+        change_notifications = bll._dal.group_edit(
```
evansd, line 1250: Shouldn't this be `self._dal`? And same for a few other instances of this pattern that were in this file already before this diff.

evansd (approved):

airlock/business_logic.py
rebkwok: yes, it should - I just edited what was already there. I'll fix the others too

airlock/business_logic.py
rebkwok: OK, sure. Easy enough to add it back in later

PR #312, opened 2024-05-08
Update styles for header, login, requests, and workspaces

tomodwyer:
## Header

- Fixed wonky account dropdown
- Moved logout button under account dropdown
- Fixed associated tests

### After

---

## Login

- Moved text under article header to improve visibility
  - This can be tweaked once users have used the system
- Improved form spacing and success button styles

### After

---

## Workspaces

- Update card layout and use article header
- Add an empty state

### After

---

## Requests

- Update card layout and use article header
- Add an empty state

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -30,7 +30,7 @@ You'll need an appropriate version of Python on your PATH. Check the
 
 ### Docker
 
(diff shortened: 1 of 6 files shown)


madwort:
a nice improvement

PR #313, opened 2024-05-08
Add support for generating test manifests.

bloodearnest:
This is used by default in test factories, but also is exposes as `just
manifests` as a dev tool to create manifests for your local test
workspaces.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -2,6 +2,7 @@ import json
 import subprocess
 import tempfile
 import time
+from hashlib import file_digest
 from pathlib import Path
 
 from django.conf import settings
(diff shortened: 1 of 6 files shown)


tests/factories.py
```diff
@@ -50,6 +51,62 @@ def ensure_workspace(workspace_or_name):
...
+        "size": stat.st_size,
+        "timestamp": stat.st_mtime,
+        "content_hash": content_hash,
+        "excluded": excluded,
+        "message": message,
+    }
+
+
+def update_manifest(workspace, files=None):
+    """Write a manfiest based on the files currently in the directory.
```
rebkwok, line 77:
```suggestion
    """Write a manifest based on the files currently in the directory.
```

tests/factories.py
```diff
@@ -50,6 +51,62 @@ def ensure_workspace(workspace_or_name):
...
+
+    if manifest_path.exists():
+        manifest = json.loads(manifest_path.read_text())
+        manifest["workspace"] = workspace.name
+        manifest.setdefault("outputs", {})
+    else:
+        manifest = {"workspace": workspace.name, "repo": None, "outputs": {}}
+
+    if files is None:  # pragma: nocover
+        files = [f.relative_to(root) for f in root.glob("**/*") if f.is_file()]
```
rebkwok, line 93: Should this skip the metadata and logs dirs? I guess it doesn't really matter for tests and local workspaces

tests/factories.py
```diff
@@ -50,6 +51,62 @@ def ensure_workspace(workspace_or_name):
...
+        "job_id": job_id,
+        "job_request": job_request,
+        "action": action,
+        "commit": commit,
+        "size": stat.st_size,
+        "timestamp": stat.st_mtime,
+        "content_hash": content_hash,
+        "excluded": excluded,
+        "message": message,
+    }
```
rebkwok, line 73: It also needs the row_count and col_count for csvs. In the version of this function that I wrote in parallel, I just had this extra bit of code (can add as part of my PR instead)
```
rows = cols = None
if abspath.suffix == ".csv":
    with abspath.open() as fp:
        reader = csv.DictReader(fp)
        first_row = next(reader, None)
        if first_row:
            cols = len(first_row)
            rows = sum(1 for _ in reader) + 1
        else:
            cols = rows = 0
```

rebkwok (approved):

tests/factories.py
bloodearnest: oh, yes we should

tests/factories.py
bloodearnest: right

Issue #249, opened 2024-04-17
Pre-launch database tidy up


evansd:
Prior to getting users starting to do their day job with Airlock it would be good to do a few tidy up tasks:
 * [ ] Squash all migrations
 * [ ] Remove the existing database file and replace with an empty one

inglesp:
This ship has sailed

PR #315, opened 2024-05-08
Store manifest.json data in the db

### LINKED ISSUE #303: Store manifest.json information in the database
We want to use the information from the manifest.json file as the source of truth for file data (size, commit etc). However, the manifest.json only contains information about the latest output files. If a job is re-run after a release request is created, we lose the old data from the manifest.json.

Issues #267 and #243 will need this.

RequestFile should grow attributes:
- timestamp
- size
- job_id
- commit
- row_count (pending https://github.com/opensafely-core/job-runner/pull/727)
- col_count (pending https://github.com/opensafely-core/job-runner/pull/727)

And `add_file_to_request()` should read these from the current manifest.json and store them on the RequestFileMetadata object.

manifest.json also has the content_hash, so we can use that when creating the file_id too.

rebkwok:
Fixes #303 

Store manifest.json data in the database when adding a file to a request.

This will need to add `repo` to the RequestFile and RequestFileMetadata once it's available


airlock/business_logic.py
```diff
@@ -269,11 +269,18 @@ def get_manifest_data(self):
         manifest_path = self.abspath("metadata/manifest.json")
         try:
             return json.loads(manifest_path.read_text())
-        except json.JSONDecodeError as exc:
+        except json.JSONDecodeError:
```
evansd, line 272: It looks like we were intending to use `exc` in the exception message but neglected to make it an f-string.

airlock/business_logic.py
```diff
@@ -269,11 +269,18 @@ def get_manifest_data(self):
...
             raise BusinessLogicLayer.FileNotFound(
                 "Could not parse manifest.json file: {manifest_path}:\n{exc}"
```
evansd, line 274: Looks like we're missing an `f` here.

local_db/migrations/0015_update_manifest_data.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 5.0.4 on 2024-05-07 12:46
+
+from django.db import migrations
+
+from airlock.business_logic import BusinessLogicLayer, Workspace
```
evansd, line 5: This might be misplaced anxiety on my part but I always get a bit nervous about db migrations which pull in a lot of application logic because they stick around for a long time during which the application logic tends to evolve.

Another option is to have a one-off, "data backfill" commands that you can generally delete after they've been used.

evansd (approved):
Looks good, mainly just a question about where best to put the data migration logic.

local_db/migrations/0015_update_manifest_data.py
rebkwok: I dithered about which way to do it, and thought the migration would be simpler, and would get squashed soon anyway in #249.
Although, I now notice that some things changed since I wrote the data migration, so it won't work properly now anyway, so I might as well move it to a separate script

evansd: > thought the migration would be simpler, and would get squashed soon anyway in #249.
Oh yeah, that is a fair point. Fine to leave it where it is then if that's easier.

rebkwok: Ah, and I just noticed a bug in the test factories that generate `manifest.json`s which has meant that I've made it lookup absolute paths when they should be relative.

rebkwok: I've made it a temporary backpopulate command so that I can also add a test for it (which would have found the   factories bug if I'd written it in the first place). Also it'll be easier to rerun when we need to add in the repo, when that's available on the manifest.json

Issue #286, opened 2024-04-25
Enable deleting or editing of user comments


bloodearnest:
Users might make mistakes (e.g. comment on the wrong group) and want to remove comments.

The two output-checkers might have written conflicting comments, and need tor remove or edit them before communicating to the user.

For all these reasons, we should enable deleting and editing of comments by the comment author.

- [x] deleting comments #317 
- [ ] editing comments

Nb. this is for day-to-day editing of comments, not for redaction. All actions & comment content should be audited in full.

inglesp:
Less necessary now we have private comments

PR #317, opened 2024-05-08
Add ability to delete comments

madwort:
* authors may delete their own comments
* full comment stored in audit log on deletion - not to be used for redaction
* part of https://github.com/opensafely-core/airlock/issues/286


airlock/urls.py
```diff
@@ -124,10 +124,15 @@
         name="group_edit",
     ),
     path(
-        "requests/comment/<str:request_id>/<str:group>",
+        "requests/comment/create/<str:request_id>/<str:group>",
```
madwort, line 127: nb. I've slightly cut a corner here but revising this url but leaving all the methods as `group_comment` instead of updating to `group_comment_create` - should I do this?

local_db/data_access.py
```diff
@@ -425,3 +426,30 @@ def group_comment(
...
+            # but let's verify we're looking at the right thing
+            if not (
+                comment.author == username
+                and comment.filegroup.name == group
+                and comment.filegroup.request.id == request_id
+            ):
+                raise BusinessLogicLayer.APIException(
+                    "Comment for deletion has inconsistent attributes "
+                    f"(in file group '{comment.filegroup.name}')"
+                )
```
madwort, line 452: nb. maybe we don't need to pass these params? (& could then remove a test or two?)

airlock/urls.py
madwort: changed

evansd (approved):
Looks good, and definitely worth merging now rather than waiting for comments to re-enabled.

Just one small question about a bit that confused me.

airlock/business_logic.py
```diff
@@ -574,13 +574,17 @@ def from_dict(cls, attrs) -> Self:
...
     comment: str
     author: str
     created_at: datetime
 
     @classmethod
     def from_dict(cls, attrs):
-        return Comment(**attrs)
+        return cls(
+            **{k: v for k, v in attrs.items() if k not in ["id"]},
+            id=f"{attrs.get('id')}",
```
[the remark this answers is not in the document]
evansd: Ah yeah, that makes sense now. I think it's still worth a comment along the lines of "IDs happen to be ints in the DAL at the moment but from the point of view of the BLL they should be opaque strings".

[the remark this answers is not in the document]
madwort: cool, yep, will add

PR #318, opened 2024-05-10
Airlock components to reduce whitespace

tomodwyer:
...and add metatitle for pages.


tomodwyer:
> 1. What's the thinking behind a header component? AIUI, we'll only ever use it in the base template? So why not just define it there?  Or is it about breaking down the html/css/js into discrete chunks?

It's initially for the chunking of the components to make styling easier, but also I've got plans to add more to the header, like moving the "Back to the workspace" button into that (when required).

> 2. We used to have the button to the right of the title, to save vertical whitespace. It looks like that got removed at somepoint - now that this PR reduces the title sizeing/whitespace, could we reintroduce it?

Sorting the buttons is the next PR that will be coming after this, and it will restore them to the right-hand side, next to the title, probably in an "Actions" type section where the buttons will be grouped and smaller.

bloodearnest:
> > 1. What's the thinking behind a header component? AIUI, we'll only ever use it in the base template? So why not just define it there?  Or is it about breaking down the html/css/js into discrete chunks?
> 
> It's initially for the chunking of the components to make styling easier, 
:+1: 
> but also I've got plans to add more to the header, like moving the "Back to the workspace" button into that (when required).

I am not sure about moving the back to workspace button into the header.  I see the "view request" and "view workspace" buttons as duals of each other, to easily switch between views, and ideally should be in the same location?

And I think I generally prefer static headers that don't change depending on context, I think that's usually a better UX?

> > 2. We used to have the button to the right of the title, to save vertical whitespace. It looks like that got removed at somepoint - now that this PR reduces the title sizeing/whitespace, could we reintroduce it?
> 
> Sorting the buttons is the next PR that will be coming after this, and it will restore them to the right-hand side, next to the title, probably in an "Actions" type section where the buttons will be grouped and smaller.

bloodearnest (approved):

PR #325, opened 2024-05-13
Temporarily remove c3 form fields

### LINKED ISSUE #324: Temporarily remove C3 form fields
While the review/conversation process is still under discussion, we will temporarily remove the C3 form fields (context, controls and comments) from Airlock and return to the google form/word doc process.

This includes updating the docs to be consistent with the form/word doc process. We also want to update the google form to have users define and name file groups (one or more output and supporting files with their context and controls) to follow the structure that airlock uses.

rebkwok:
Fixes #324 

Hides the form fields for context, comments and controls based on a new SHOW_C3 env variable and setting.

Also updates docs to be consistent with the (temporary) new process. The [minimally-adapted Word doc form](https://docs.google.com/document/d/1uWRiFps6tDA2SpxSxf0C2G9mOVWMQ6TQ/edit?usp=drive_link&ouid=114747311902500453985&rtpof=true&sd=true) should also be reviewed before this is merged.

### DIFF
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -402,3 +402,5 @@ LOGGING = {
         },
     },
 }
+
+SHOW_C3 = os.environ.get("AIRLOCK_SHOW_C3", "false") == "True"
--- a/airlock/templates/file_browser/group.html
+++ b/airlock/templates/file_browser/group.html
@@ -1,40 +1,41 @@
 {% #card title=path_item.name container=True %}
-  <form action="{{ group_edit_url }}" method="POST" aria-label="group-edit-form">
-    {% csrf_token %}
-    <div style="display: flex; flex-direction: row; width: 100%; flex-wrap: wrap; gap: 1rem;">
-      <div style="display: flex: flex-direction: column; flex-basis: 100%; flex: 1;">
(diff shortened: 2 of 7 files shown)


airlock/settings.py
```diff
@@ -268,7 +268,7 @@ def get_env_var(name):
 
 DJANGO_VITE = {
     "default": {
-        # vite assumes collectstatic, so tell it where the manifest is directly
+        # vite assumes collectstatic, so te"WARNING"ll it where the manifest is directly
```
evansd, line 271: Stray copy/paste?

airlock/settings.py
```diff
@@ -402,3 +402,5 @@ def filter(self, record):  # pragma: no cover
         },
     },
 }
+
+SHOW_C3 = os.environ.get("SHOW_C3", "false") == "True"
```
evansd, line 406: Because of the shared, global nature of env vars I'd be tempted to call the env var `AIRLOCK_SHOW_C3` just so it's obvious at a glance what system it relates to. I think the setting name can stay as it is.

As this is (hopefully) a short-lived setting it probably doesn't matter much either way, but just trying to spread the general convention around.

airlock/settings.py
rebkwok: bah

airlock/settings.py
rebkwok: Oh yes, I did that with the last one I added. will fix

PR #336, opened 2024-05-22
Rework directory view to use datatable

bloodearnest:
This uses a datatable to list the contents in a directory. It was spun
out of the multiselect work.

This means we can sort and order on columns, and so I also add size and
modified times for this too, to provide a use case.

There will be other columns coming soon, one for current file state
(unreleased/in-request/released for workspace files,
approved/rejected/withdrawn for request files), as well as a checkbox
column for the multiselect.

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -308,18 +308,38 @@ class Workspace:
             return self.manifest["outputs"][str(relpath)]
         except KeyError:
             raise BusinessLogicLayer.ManifestFileError(
-                f"Could not parse data for {relpath} from manifest.json file"
+                f"No entry for {relpath} from manifest.json file"
             )
 
     def get_size(self, relpath: UrlPath) -> int:
-        return int(self.get_manifest_for_file(relpath).get("size", 0))
+        try:
+            return int(self.get_manifest_for_file(relpath).get("size", 0))
+        except BusinessLogicLayer.ManifestFileError:
+            pass
+
+        # not in manifest, e.g. log file. Check disk
+        try:
+            abspath = self.abspath(relpath)
+        except BusinessLogicLayer.FileNotFound:
+            return 0
+
+        return int(abspath.stat().st_size)
 
     def get_modified_time(self, relpath: UrlPath) -> datetime | None:
-        ts = self.get_manifest_for_file(relpath).get("timestamp")
-        if ts:
-            return datetime.utcfromtimestamp(ts)
-        else:  # pragma: no cover
+        try:
+            return datetime.utcfromtimestamp(
+                self.get_manifest_for_file(relpath).get("timestamp")
+            )
+        except BusinessLogicLayer.ManifestFileError:
+            pass
+
+        # not in manifest, check disk
+        try:
+            abspath = self.abspath(relpath)
+        except BusinessLogicLayer.FileNotFound:
             return None
+
+        return datetime.utcfromtimestamp(abspath.stat().st_mtime)
 
     def abspath(self, relpath):
         """Get absolute path for file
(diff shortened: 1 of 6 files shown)


airlock/file_browser_api.py
```diff
@@ -510,3 +531,8 @@ def children_sort_key(node: PathItem):
     # this works because True == 1 and False == 0
     name = node.name()
     return (name != "metadata", node.type == PathType.FILE, name)
+
+
+def to_kb(value_in_bytes):
+    """Convert given value to Mb"""
```
madwort, line 537:
```suggestion
    """Convert given value to kb"""
```

madwort: or is it 
```suggestion
    """Convert given value to kb"""
```
🤔

airlock/templates/file_browser/directory.html
```diff
@@ -1,16 +1,61 @@
...
+
+  #customTable td.name a.directory {
+    background-image: url(/static/folder.png);
+    background-repeat: no-repeat;
+    background-size: 1.4rem;
+    background-position: left -0.2rem top 0;
+    padding-left: 1.1rem;
+  }
+
+</style>
```
madwort, line 21: I think @tomodwyer recently put all the custom CSS into a separate file - can we do that here?

airlock/business_logic.py
```diff
@@ -305,6 +311,16 @@ def get_manifest_for_file(self, relpath: UrlPath):
                 f"Could not parse data for {relpath} from manifest.json file"
             )
 
+    def get_size(self, relpath: UrlPath) -> int:
+        return int(self.get_manifest_for_file(relpath).get("size", 0))
+
+    def get_modified_time(self, relpath: UrlPath) -> datetime | None:
+        ts = self.get_manifest_for_file(relpath).get("timestamp")
```
madwort, line 318: also catch `ManifestFileError`?

airlock/templates/file_browser/directory.html
bloodearnest: He did?

Hmm, that might cause some problems.

madwort: https://github.com/opensafely-core/airlock/pull/314/commits/1a56f1a70d063d635d5bfb50b598511de1aa621b

airlock/business_logic.py
```diff
@@ -305,6 +311,16 @@ def get_manifest_for_file(self, relpath: UrlPath):
                 f"Could not parse data for {relpath} from manifest.json file"
             )
 
+    def get_size(self, relpath: UrlPath) -> int:
+        return int(self.get_manifest_for_file(relpath).get("size", 0))
```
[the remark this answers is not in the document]
bloodearnest: Good call, I had forgotten about log files.

I will fall back to filesystem stat in that case.

airlock/templates/file_browser/directory.html
bloodearnest: Hmm, I am not sure we want all datatables to have sticky header.

I am a fan of keeping style local to their use.  The current datatables implementation makes this a bit more awkward, as you have to use #customTable

bloodearnest: Also the the a.directory styles are very specific to *this* table, not others.

madwort: I think ultimately we'll want this to move out of the template & into a CSS file, but it doesn't have to be in this pr. If all the tables are `#customTable`, could we put an id on the parent card & refer to it like `#directoryTable #customTable thead` etc?

madwort (approved):

PR #345, opened 2024-05-24
Add code browser links

rebkwok:
Adds a link to view code from workspace or request files.
Uses a return_url query param so user can go back to whichever file they came from.

I've also updated `just manifests` to create a local git repo for test workspaces as well as the manifest.json, which mean that code links from workspaces will work. 

However, any existing test request files will have been created with old dummy commits from the manifest.json, so you also need to run `just manage backpopulate_file_manifest_data` to fix them. (If you see "Could not update manifest.json data..." messages, it's probably because you have requests with files that no longer exist in the workspace folder)

### DIFF
--- a/airlock/templates/file_browser/file.html
+++ b/airlock/templates/file_browser/file.html
@@ -65,6 +65,9 @@
       {% endif %}
     {% endif %}
     {% #button variant="primary" type="link" href=path_item.contents_url external=True id="view-button" %}View ↗{% /button %}
+    {% if code_url %}
+      {% #button type="link" variant="primary" id="file-code-button" href=code_url %}View code{% /button %}
+    {% endif %}
   </div>
 
 {% endfragment %}
--- a/airlock/templates/file_browser/file.html
+++ b/airlock/templates/file_browser/file.html
@@ -66,7 +66,7 @@
     {% endif %}
     {% #button variant="primary" type="link" href=path_item.contents_url external=True id="view-button" %}View ↗{% /button %}
     {% if code_url %}
-      {% #button type="link" variant="primary" id="file-code-button" href=code_url %}View code{% /button %}
+      {% #button type="link" variant="primary" id="file-code-button" href=code_url external=True %}View code{% /button %}
     {% endif %}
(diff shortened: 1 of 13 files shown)


airlock/views/workspace.py
```diff
@@ -93,6 +93,17 @@ def workspace_view(request, workspace_name: str, path: str = ""):
...
         )
 
+    if path_item.is_directory() or path not in workspace.manifest["outputs"]:
+        code_url = None
+    else:
+        code_url = reverse(
+            "code_view",
+            kwargs={
+                "workspace_name": workspace.name,
+                "commit": workspace.get_manifest_for_file(path)["commit"],
```
bloodearnest: Is it worth handling the case where the "commit" key isn't found?

airlock/templates/file_browser/file.html
```diff
@@ -65,6 +65,9 @@
       {% endif %}
     {% endif %}
     {% #button variant="primary" type="link" href=path_item.contents_url external=True id="view-button" %}View ↗{% /button %}
+    {% if code_url %}
+      {% #button type="link" variant="primary" id="file-code-button" href=code_url %}View code{% /button %}
```
bloodearnest, line 69: I wonder if we should we default to opening this is new tab?

bloodearnest (approved):

airlock/views/workspace.py
rebkwok: Can do, but will there ever be one? `code_url` will be `None` if the path isn't in `manifest["outputs"]`, shouldn't every output file have a commit key?

bloodearnest: It should, but we don't currently validate the manifest.json beyond parsing it as valid json, its possible a format change or an error might

I'd rather just break the code link than the whole page?

rebkwok: I've changed it to a .get() so the page doesn't break on a KeyError, but I've also updated the code view so if the repo/commit doesn't exist it gives you an error message instead of just 404ing when you click the code button

PR #346, opened 2024-05-24
Add workspace file state detection

### LINKED ISSUE #343: Add support for querying a files state in the process
For specific filepath and content_hash a file in a workspace can be

RELEASED - has already been released
UNDER_REVIEW - is currently in an outstanding request
UPDATED - filename is in an outstanding request, but has been updated

These states are need to implement various UI hints and workflows about adding/withdrawing/updating files.

This information needs to be included in the tree rendering, so we can style tree nodes accordingly. As the same tree rendering is used for both workspaces and requests, there needs to be a get_state() api similar to the get_url or get_contents_url API, that will in future work for request files states (votes, approved, etc)

bloodearnest:
This adds a core capability - knowing whether a file is part of the current reqeust and if it has been updated since added.

It uses this new capability to a) decorate the tree with classes for later styling and b) decide if a file can be added to a request or not.

fixes #343

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -606,7 +606,7 @@ class ReleaseRequest:
         return url
 
     def get_renderer(self, relpath: UrlPath) -> renderers.Renderer:
-        request_file = self.get_request_file(relpath)
+        request_file = self.get_request_file_from_urlpath(relpath)
         renderer_class = renderers.get_renderer(relpath)
         return renderer_class.from_file(
             self.abspath(relpath),
@@ -615,14 +615,15 @@ class ReleaseRequest:
         )
 
     def get_file_metadata(self, relpath: UrlPath) -> FileMetadata:
-        rfile = self.get_request_file(relpath)
+        rfile = self.get_request_file_from_urlpath(relpath)
         return FileMetadata(
             rfile.size,
             rfile.timestamp,
             _content_hash=rfile.file_id,
         )
 
-    def get_request_file(self, relpath: UrlPath | str):
+    def get_request_file_from_urlpath(self, relpath: UrlPath | str):
+        """Get the request file from the url, which includes the group."""
         relpath = UrlPath(relpath)
         group = relpath.parts[0]
         file_relpath = UrlPath(*relpath.parts[1:])
@@ -640,7 +641,7 @@ class ReleaseRequest:
 
         The first part of the relpath is the group, so we parse and validate that first.
         """
-        request_file = self.get_request_file(relpath)
+        request_file = self.get_request_file_from_urlpath(relpath)
         return self.root() / request_file.file_id
 
     def all_files_set(self):
@@ -672,7 +673,7 @@ class ReleaseRequest:
         return next(
             (
                 r
-                for r in self.get_request_file(urlpath).reviews
+                for r in self.get_request_file_from_urlpath(urlpath).reviews
                 if r.reviewer == reviewer
             ),
             None,
@@ -680,7 +681,7 @@ class ReleaseRequest:
 
     def request_filetype(self, urlpath: UrlPath):
         try:
-            return self.get_request_file(urlpath).filetype
+            return self.get_request_file_from_urlpath(urlpath).filetype
         except BusinessLogicLayer.FileNotFound:
             return None
 
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -26,7 +26,7 @@ from airlock.lib.git import (
     read_file_from_repo,
 )
 from airlock.notifications import send_notification_event
-from airlock.types import FileMetadata, UrlPath
+from airlock.types import FileMetadata, UrlPath, WorkspaceFileState
 from airlock.users import User
 from airlock.utils import is_valid_file_type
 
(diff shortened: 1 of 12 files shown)


airlock/types.py
```diff
@@ -12,3 +18,70 @@
     class UrlPath(PurePosixPath): ...
 else:
     UrlPath = PurePosixPath
+
+
+@dataclass
+class FileMetadata:
```
inglesp, line 35: Is types.py the right place for this?  What's the criteria for something going in this file?

airlock/types.py
```diff
@@ -12,3 +18,70 @@
...
+    """Represents the base properties of file metadata.
+
+    Often these are in the manifest.json, so can be read from there. But
+    sometimes they are not, so need reading from the filesystem.
+    """
+
+    size: int | None
+    timestamp: int | None
+    _content_hash: str | None = None
+    path: Path | None = None
```
inglesp, line 45: It'd be helpful to understand when these values can be null or not.  If they're either all set or all null, that suggests we've not got the design quite right here.

tests/unit/test_business_logic.py
```diff
@@ -112,25 +112,48 @@ def test_workspace_manifest_for_file_not_found(bll):
...
-def test_get_size():
+def test_get_file_metadata():
     workspace = factories.create_workspace("workspace")
-    assert workspace.get_size(UrlPath("metadata/foo.log")) == 0
+    empty = workspace.get_file_metadata(UrlPath("metadata/foo.log"))
+    assert empty.size is None
+    assert empty.size_mb == ""
+    assert empty.timestamp is None
+    assert empty.modified_at is None
+    assert empty.content_hash is None
```
inglesp, line 123: What's the scenario that this is testing?

airlock/types.py
inglesp: Ah -- so a directory can have a `FileMetadata`, whose values are `None`.  This doesn't feel quite right.

airlock/business_logic.py
```diff
@@ -276,6 +279,29 @@ def get_url(self, relpath: UrlPath = ROOT_PATH) -> str:
             kwargs["path"] = str(relpath)
         return reverse("workspace_view", kwargs=kwargs)
 
+    def get_workspace_state(self, relpath: UrlPath) -> WorkspaceFileState | None:
+        try:
+            self.abspath(relpath)
+        except BusinessLogicLayer.FileNotFound:
```
inglesp, line 286: Why would a file not be found?  This could do with a comment.

airlock/types.py
inglesp: Could we have a `MetadataInterface` or whatever you'd call it in Python, with `FileMetadata` and `DirMetadata` implementations?

inglesp (approved):
Looks really useful.

I've got a handful of minor questions/suggestions.  But I'd prefer to get this merged before you go away, so feel free to ignore.

airlock/types.py
bloodearnest: The criteria is that the type needed by both business_logic.py and file_browser_api.py. We can get in a mess with circular imports otherwise, especially since we're statically typing things.

It's basically the root of the type dependency tree.

airlock/types.py
bloodearnest: We could, but I'm not sure what problem that type would solve? 

This was actually the simplest implementation I could got to. I spent some time in the weeds trying to get the typing to play nice here.

For context, there are two consumers of the FileMetadata interface - we use it in regular logic code, where we can could easily just return a None instead of a FileMetadata in the directory case.

But the other consumer is in the templates, where its much trickier to handle a None value that we need to access attributes of if its not None. Perhaps we can use some template filters to do this? As so make `def get_file_metadata(...) -> FileMetadata | None`, and not need the the FileMetadata.empty constructor and None-able fields?

tests/unit/test_business_logic.py
bloodearnest: There is no manifest and no file on disk to supply the metadata

airlock/business_logic.py
bloodearnest: "Shit happens" :rofl: 

Seriously, was just defense in depth so we don't blow up if we get passed a bad path.

airlock/types.py
bloodearnest: I had a quick look at this, and I can't see a way to do it that doesn't give template variable errors. I think this is a consequence of us not letting template variables error.

We could have `size=-1` instead of None, but that's just as bad.

The default template filter doesn't work on errors, on truthy or none.

In jinja, you can do `x.y|default('foo')` and if x is None, it will work.

bloodearnest: I have found a fix for the awkward type.

It means more boiler plate properties on PathItem, but that is standard fair for django django templates I think.

tests/unit/test_business_logic.py
bloodearnest: yep

airlock/business_logic.py
bloodearnest: Blowing up here would mean the entire workspace cannot be viewed, as this is called when rendering the tree,

But it is a very unlikely case, so maybe that's ok.

Issue #284, opened 2024-04-25
Support un-approving/rejecting


bloodearnest:
Users may make a mistake, and want to unapprove or unreject something.

We should add a button to remove any current FileReviews for this file.

It can possibly a pill like we used in SACRO, but with 3 sections?

![Image](https://github.com/opensafely-core/airlock/assets/1042/c033a282-7658-4fde-b513-b26aacceba97)

bloodearnest:
That's a good question, I'm not sure.

Depending on how we do the #254, I am wondering if we do need this. I just originally captured it as part of Catherine's feed back.

However, if we move to a world where to progress, an output-checker has to vote accept/reject on every file, then maybe this isn't needed?

I'd put it on the back burner until we know?

madwort:
Yeah, it's a small qol improvement, but I think it would make life easier for output-checkers regardless of #254. As you say, if we had a "review completed" button then the overall process wouldn't move forward until they'd hit that button - but in terms of completing their own work if they initially approved a file & then had doubts, they would have to store a dirty flag for that file in their brains until they get time to complete the review, whereas if they had a "remove review" button they could quickly do that as a placeholder until they get time to do the work to approve/reject.

bloodearnest:
Yeah, that's good point.

In that case, do we delay audit until submitting the final votes/comments?

And just delete the FileReview?

madwort:
I think we audit everything as it happens - that way we can audit our systems as well as what the users did - so the audit log would look like:
* file x review approved
* file x review deleted
* file x review ...etc...

bloodearnest:
Right, that is simpler.

But the audit trail serves double purpose. It's a forensic trail (for which recording everything makes sense, even if its immediately reversed). But its also a UX element for users to know what activity has happened. In that case, the additional spurious logs could be quite noisy.

PR #347, opened 2024-05-24
Support undo for a file review

madwort:
* UI buttons could be better, but we can fix that in post 
* fixes https://github.com/opensafely-core/airlock/issues/284
Screenshot 2024-05-24 at 17 08 16

### DIFF
--- a/local_db/data_access.py
+++ b/local_db/data_access.py
@@ -322,6 +322,19 @@ class LocalDBDataAccessLayer(DataAccessLayerProtocol):
 
             self._create_audit_log(audit)
 
+    def reset_review_file(
+        self, request_id: str, relpath: UrlPath, username: str, audit: AuditEvent
+    ):
+        with transaction.atomic():
+            request_file = RequestFileMetadata.objects.get(
+                request_id=request_id, relpath=relpath
+            )
+
+            review = FileReview.objects.get(file=request_file, reviewer=username)
+            review.delete()
+
+            self._create_audit_log(audit)
+
     def _create_audit_log(self, audit: AuditEvent) -> AuditLog:
         event = AuditLog.objects.create(
             type=audit.type,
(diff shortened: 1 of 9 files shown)


local_db/data_access.py
```diff
@@ -322,6 +322,19 @@ def reject_file(
...
 
+    def reset_review_file(
+        self, request_id: str, relpath: UrlPath, username: str, audit: AuditEvent
+    ):
+        with transaction.atomic():
+            request_file = RequestFileMetadata.objects.get(
+                request_id=request_id, relpath=relpath
+            )
+
+            review = FileReview.objects.get(file=request_file, reviewer=username)
```
rebkwok, line 333: Should this catch the exception if the review doesn't exist? (It's reasonably easy to trigger if you have 2 airlock tabs open )

rebkwok (commented):
One question, and a failing test because of the recent change to `get_request_file`; otherwise it generally lgtm, although the buttons are getting a bit crazy. Would it make more sense not to have a separate button for the reset review, but have re-clicking approve/reject for an already approved/rejected file do the resetting (could also the tooltip make the tooltip say "File already approved; click to reset"?

local_db/data_access.py
madwort: yep, have added handler & tests for this.

madwort:
Yeah, I agree that we've got too many buttons right now, but I don't think we should aggregate the buttons right now. I think these buttons are currently the only way that output checkers can see whether they'vereviewed a file or not, and so overloading the buttons in that way would make that a lot more difficult. If we had review status in the file menu on the left, I think it might be ok to overload them?

I think there's a few other ideas for improving the UI - in the original ticket Simon suggested adding another type of widget, another option might be to copy Github & have a "review button" that produces a drop-down with three options, or something else.

What do you think about merging this as-is and then creating another ticket to look at the UI, probably involving Tom O tbh?

madwort:
created an issue to improve the UI https://github.com/opensafely-core/airlock/issues/357

Issue #344, opened 2024-05-23
Record when files are released


bloodearnest:
Currently, when we release files, we only audit and change the state of the request, not the individual files.

We should 
- [x] a) emit an audit event for each file #358 
- [x] b) mark in the db when/who released the file. #365

In future work, this will allow us to show (in the tree and in contents) that a file you are viewing has been already been released. And we can prevent you from adding that file *as an output file* to a new request.

madwort:
> And we can prevent you from adding that file to a new request.

Seems good in principle but wondering about edge cases, e.g. if there was a file that has been released as an output file & then a researcher wanted to add it as a supporting file to a future release request?

bloodearnest:
Good point. I don't think at affects this ticket, but will any future one about adding files based on this. Worth writing that up now?

madwort:
emitting an audit log event may be trickier than I first thought, as [release_files is using old_api](https://github.com/opensafely-core/airlock/blob/main/airlock/business_logic.py#L1353-L1358)

PR #358, opened 2024-05-29
Add an audit log entry for every file released

madwort:
* part 1 of #344


airlock/business_logic.py
```diff
@@ -1369,28 +1376,35 @@ def withdraw_file_from_request(
...
 
         for relpath, abspath in file_paths:
             old_api.upload_file(jobserver_release_id, relpath, abspath, user.username)
+            audit = AuditEvent.from_request(
+                request=release_request,
+                type=AuditEventType.REQUEST_FILE_RELEASE,
+                user=user,
+                path=relpath,
+            )
+            self._dal.release_file(release_request.id, relpath, user.username, audit)
```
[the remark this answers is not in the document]
madwort: sorry, audit should be before `upload_file`

airlock/business_logic.py
```diff
@@ -1350,28 +1357,35 @@ def withdraw_file_from_request(
...
 
-        filelist = old_api.create_filelist(file_paths, request)
+        filelist = old_api.create_filelist(file_paths, release_request)
         jobserver_release_id = old_api.create_release(
-            request.workspace, filelist.json(), user.username
+            release_request.workspace, filelist.json(), user.username
         )
 
         for relpath, abspath in file_paths:
             old_api.upload_file(jobserver_release_id, relpath, abspath, user.username)
```
evansd: I think ideally we'd reverse the order of operations here because, as things stand, a failure at just the wrong point in the process would lead to an uploaded file with no associated audit event. If we create the audit event first we risk having an event with no uploaded file, but I think that's a safer failure mode.

evansd (approved):

airlock/business_logic.py
madwort: yep done

PR #360, opened 2024-05-29
Handle empty CSV files

evansd:
We had one of these in production and it triggered a 500 which then triggered a confusing iframe permissions error.

### DIFF
--- a/airlock/renderers.py
+++ b/airlock/renderers.py
@@ -121,7 +121,7 @@ class CSVRenderer(Renderer):
 
     def context(self):
         reader = csv.reader(self.stream)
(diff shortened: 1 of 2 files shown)


madwort (changes requested):
I thought empty files were supposed to be an invalid file type & not be able to be viewed - which also means that they're not able to be added to a release request. It's good to avoid the crash, but I think actually we need a slightly different fix.

inglesp (approved):

madwort (approved):

madwort:
my mistake - empty files (e.g. txt files) are indeed viewable!

PR #365, opened 2024-05-31
Store file release metadata in db

madwort:
* second part of #344


madwort:
CI failing due to https://github.com/spulec/freezegun/issues/435 - may need to remove freezegun?

airlock/business_logic.py
```diff
@@ -513,9 +513,11 @@ class RequestFile:
     job_id: str
     commit: str
     repo: str
+    release_output_checker: str
```
evansd: I guess this is non-nullable because we just use an empty string instead of NULL? Feels like it might be conceptually neater and more consistent if it was nullable though.

evansd (approved):

airlock/business_logic.py
madwort: yes, I went with empty string, could be NULL though? probably conceptually makes more sense - I'll change.

Issue #373, opened 2024-06-04
Change "rejected" to "changes requested"


inglesp:
Output checkers have told us that they're unsure what "rejected" means.  Rejection sounds final, and apparently rejecting an output outright has been very rare.

Instead, we should change what we display to "request changes"/"changes requested".

An alternative would be to have three states: "approved", "changes requested", "rejected".  But given how rare outright rejection is, this feels like a waste of valuable screen space.

- [x] change the wording in the UI #377
- ~migrate the ENUM to a matching value (later, not necessarily for beta)~ moved to a new ticket

madwort:
we think this is about files, not release requests

madwort:
closing this as done & creating a new ticket for the Enum refactor to follow

PR #377, opened 2024-06-05
Reword 'reject' to 'request changes' for files

madwort:
* https://github.com/opensafely-core/airlock/issues/373


docs/reviewing.md
```diff
@@ -5,7 +5,7 @@ Whilst viewing a file in an Airlock release request, they have the option of:
 * **Reject** — output is not currently acceptable for release.
```
evansd, line 5: I think this text here needs updating as well

evansd (approved):

docs/reviewing.md
madwort: yep done

PR #385, opened 2024-06-07
Update layout and styling of file buttons

tomodwyer:
So that there are fewer buttons visible, and add a clear hierarchy of buttons.

This also allows output checkers to revert the status of a file by clicking on the active status, rather than having a specific button for revert the status.


airlock/views/workspace.py
```diff
@@ -135,6 +135,7 @@ def workspace_view(request, workspace_name: str, path: str = ""):
             # for code button
             "code_url": code_url,
             "return_url": "",
+            "is_output_checker": request.user.output_checker or False,
```
tomodwyer, line 138: Wasn't sure if this was the right thing to do?

airlock/templates/file_browser/file.html
```diff
@@ -43,45 +43,91 @@
           </form>
         {% endif %}
       {% elif is_output_checker %}
+        <div>
+          {% if not file_reset_review_url %}
+            <span class="file-status file-status--warning">No status</span>
+          {% elif file_approve_url %}
+            <span class="file-status file-status--danger">Rejected</span>
```
rebkwok, line 50: The button says "Request changes" button, can this say "changes requested" (or something similar) rather than "rejected". (There's a ticket to do the more complicated work to change the state from rejected to changes-requested in the backend)

airlock/templates/file_browser/file.html
```diff
@@ -43,45 +43,91 @@
...
+            <img height="16" width="16" class="icon inline-block group-open:hidden" src="/static/icons/keyboard_arrow_down.svg" alt="">
+            <img height="16" width="16" class="icon hidden group-open:inline-block" src="/static/icons/keyboard_arrow_up.svg" alt="">
+          </span>
+        </summary>
+        <ul class="more-dropdown__nav">
+          {% if is_output_checker %}
+            <li class="more-dropdown__item">
+              <a href="{{ path_item.download_url }}" class="more-dropdown__link" id="download-button">
+                <img class="more-dropdown__icon" src="/static/icons/cloud_download.svg" alt="">
+                Download file
```
rebkwok, line 106: This makes the download option visible from a workspace view for all output-checkers. It should (currently at least) only be allowed for a file in a release request, and only for output checkers if they are not the author of the request.
(For release requests, the download url 403s for an author who's also output checker; for workspace files the method that generates the `path_item.download_url` just ignores the download bit and returns the content url, so it doe sthe same as the "View file" button)

airlock/views/workspace.py
rebkwok: Shouldn't be needed for workspace files, I think? If it's needed for the template validation, the `or False` isn't needed (it's a boolean)

rebkwok (commented):
Buttons look good, but some comments on where the download button should/shouldn't be shown. If we had some playwright tests that checked for that, it would have been more obvious to you, sorry

airlock/views/workspace.py
tomodwyer: Lovely, I'll remove that.

airlock/templates/file_browser/file.html
tomodwyer: Good spot, this bit was done before I was told that was changing 👍

airlock/templates/file_browser/file.html
tomodwyer: I think I've covered that off now with:

```py
{% if is_output_checker and context == "request" and not is_author %}
```

Is that all that is required?

rebkwok: I think `{% if context == "request" and not is_author %}` is enough - a user shouldn't be able to see the request at all unless  they're an output checker or the author

rebkwok: Yes, `get_release_request_or_raise` is called for every request-related view, and it'll raise a 403 if the user isn't either an output-checker or the author

tomodwyer: 👌 perfect, I'll change that

rebkwok (approved):

PR #390, opened 2024-06-13
Allow updating a file in a release request

madwort:


tests/unit/test_business_logic.py
```diff
@@ -1340,6 +1442,201 @@ def test_add_file_to_request_with_filetype(bll, filetype, success):
...
+            workspace_file = factories.request_file(
+                path=path,
+                user=author,
+                rejected=True,
+            )
+    else:
+        workspace_file = factories.request_file(
+            path=path,
+            user=author,
+        )
```
madwort, line 1556: very open to a better way to do this

madwort: e.g. 
```suggestion
    if reviewed:
        if approved:
          approved = True
          rejected = None
        else:
          approved = None
          rejected = True
    else:
      approved = None
      rejected = None
    workspace_file = factories.request_file(
        path=path,
        user=author,
        approved=approved,
        rejected=rejected,
    )
```

local_db/data_access.py
```diff
@@ -194,13 +194,18 @@ def delete_file_from_request(
         with transaction.atomic():
             # defense in depth
             request = self._find_metadata(request_id)
-            assert request.status == RequestStatus.PENDING
+
+            assert (
+                request.status == RequestStatus.PENDING
+                or request.status == RequestStatus.RETURNED
+            )
```
rebkwok, line 201: This could use the new concept of status owner instead of listing specific statuses (`BusinessLogicLayer.STATUS_OWNERS[request.status] == RequestStatusOwner.AUTHOR`)

tests/functional/test_e2e.py
```diff
@@ -453,6 +453,66 @@ def test_e2e_release_files(
...
+    # subdirectories start off collapsed; the file links are not present
+    assert page.get_by_role("link", name="file.txt").all() == []
+    assert page.get_by_role("link", name="file.foo").all() == []
+
+    # Click on the subdir and then the file link to view in the workspace
+    # There will be more than one link to the folder/file in the page,
+    # one in the explorer, and one in the main folder view
+    # Click on the first link we find
+    find_and_click(page.get_by_role("link", name="subdir").first)
+
```
rebkwok, line 494: Everything from logging in to here is already tested in other functional tests, I think. You could skip it and go straight to the directory page with `page.goto(workspace.get_url("subdir")`

tests/unit/test_business_logic.py
```diff
@@ -1441,6 +1442,201 @@ def test_add_file_to_request_with_filetype(bll, filetype, success):
...
+    workspace = factories.create_workspace("workspace")
+    factories.write_workspace_file(workspace, relpath)
+    release_request = factories.create_release_request(
+        "workspace",
+        user=author,
+    )
+
+    # This code reproduces bll.add_file_to_request() - except without the
+    # safeguards - in order that we can give update_file_in_request()
+    # the correct context to hit this test
```
rebkwok, line 1458: There's another test that creates a request file with an invalid filetype so it can test that you can't release files of type .foo. It does it by patching the LEVEL4_FILE_TYPES so it can make the request. 
i.e. I think you should be able to do this, and avoid the lengthy workaround:
```
with patch("airlock.utils.LEVEL4_FILE_TYPES", [".foo"]):
    release_request = factories.create_release_request_at_status(
        "workspace",
        author=author,
        files=[factories.request_file(path="path/file.foo"]
    )
```

tests/unit/test_business_logic.py
```diff
@@ -1441,6 +1442,201 @@ def test_add_file_to_request_with_filetype(bll, filetype, success):
...
+
+    with pytest.raises(bll.RequestPermissionDenied, match=r"not updated"):
+        bll.update_file_in_request(release_request, relpath, author)
+
+
+@pytest.mark.parametrize(
+    "status,reviewed,approved,success,notification_sent",
+    [
+        (RequestStatus.PENDING, False, None, True, False),
+        (RequestStatus.SUBMITTED, False, None, False, False),
```
rebkwok, line 1516: `approved=None` doesn't mean anything different to `approved=False`; is it intended to?

tests/unit/test_business_logic.py
```diff
@@ -1441,6 +1442,201 @@ def test_add_file_to_request_with_filetype(bll, filetype, success):
...
+    checkers = factories.get_default_output_checkers()
+    workspace = factories.create_workspace("workspace")
+    path = UrlPath("path/file.txt")
+
+    if status == RequestStatus.WITHDRAWN:
+        withdrawn_after = RequestStatus.RETURNED
+    else:
+        withdrawn_after = None
+
+    if reviewed:
```
rebkwok, line 1539: In `request_file`, approved=True means that it's approved by 2 reviewers, rejected=True means that it's rejected by 2 reviewers (and it's an if/elif, so if you call it with approved=True and rejected=True, it'll be approved.). So I think you can get rid of the reviewed parameter here, and get the same file with:
```
workspace_file = factories.request_file(
    path=path,
    user=author,
    approved=approved,
    rejected=rejected
)
```

airlock/business_logic.py
```diff
@@ -1539,6 +1542,108 @@ def add_file_to_request(
...
+            repo=manifest["repo"],
+            size=manifest["size"],
+            job_id=manifest["job_id"],
+            row_count=manifest["row_count"],
+            col_count=manifest["col_count"],
+            audit=audit,
+        )
+        release_request.set_filegroups_from_dict(filegroup_data)
+
+        if release_request.status != RequestStatus.PENDING:
```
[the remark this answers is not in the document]
madwort: Fine, it seemed like we were just notifying on every action, so I just did that - will remove.

local_db/data_access.py
madwort: done

tests/unit/test_business_logic.py
madwort: no, I just used that to differentiate cases where it wouldn't be used at all from cases where it would be set but be false. but actually with the tiny refactor from your other comment that becomes a moot point - I'll just set them to `False`.

tests/unit/test_business_logic.py
madwort: good idea, done

tests/unit/test_business_logic.py
madwort: done

tests/functional/test_e2e.py
madwort: good shout, thanks, done

tests/unit/test_business_logic.py
madwort: great shout, done

airlock/business_logic.py
```diff
@@ -1539,6 +1542,108 @@ def add_file_to_request(
...
+
+        src = workspace.abspath(relpath)
+        file_id = store_file(release_request, src)
+
+        manifest = workspace.get_manifest_for_file(relpath)
+        assert (
+            manifest["content_hash"] == file_id
+        ), "File hash does not match manifest.json"
+
+        reviews = release_request.get_request_file_from_output_path(relpath).reviews
```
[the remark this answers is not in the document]
madwort: > I think it is necessary to reset the reviews & withdraw the file

yes, I had some discussion with @bloodearnest about this whilst working on the ticket - I think it is necessary to reset reviews & withdraw because of the foreign keys. Therefore if we build that using the current functions (by which I mean e.g. `dal.reset_review_file()`) we need audit events for everything (or allow those functions to support not creating an audit event :/ ).

One reason that I wanted to keep the reset review audit log event was because I thought that an output checker who was skimming the audit logs in order to see why a file that they had previously reviewed now didn't have a review would be more likely to notice "review reset" than "file updated". But I'm really not wedded to that idea! 

> Perhaps we need to push this down to the DAL

yes, I wondered about that before. I partly didn't because in order to do that right now, I'd need to iterate through the reviews in the BLL to create the audit events for each review/withdrawal & pass through some kind of dict of audit events to the DAL, then iterate again in the DAL. The way it is in the PR, the DAL implementation bleeds into the BLL a bit, but that would still happen with the audit logs - and I think we have to generate them in the BLL. It could be done, though, especially if we decide to drop audit events for the e.g. reset review.

> multiple calls to DAL methods that will each be within their own transaction

that would be a nice benefit!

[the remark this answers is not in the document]
bloodearnest: This is exactly the kind of issue that prompted me to add this topic for discussion in the team meeting today.

We need to figure out the transactional of the DAL API. We've attempted to keep the logic in the BLL, but that means adding a list of audits or similar. Pushing some parts of the logic down into the DAL will improve the transactionality, but the DAL will grow responsiblity and complexit.

I think we also need some discussion about what exactly needs auditing, and when, which might impact on this question.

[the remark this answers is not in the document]
rebkwok: Given that the BLL/DAL/auditing discussion may take a while, I'm happy to +1 this as it is.

rebkwok (approved):

airlock/business_logic.py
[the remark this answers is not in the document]
madwort: ok, thanks, let's merge

PR #391, opened 2024-06-14
Fallback to fetching repo from first manifest output

rebkwok:
For older manifests, the top level repo may be None, but it can be retrieved from the output file data.

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -437,9 +437,7 @@ class CodeRepo:
     @classmethod
     def from_workspace(cls, workspace: Workspace, commit: str):
         try:
-            repo = workspace.manifest.get("repo")
-            if repo is None:
-                repo = list(workspace.manifest["outputs"].values())[0]["repo"]
+            repo = list(workspace.manifest["outputs"].values())[0]["repo"]
         except (BusinessLogicLayer.ManifestFileError, IndexError, KeyError) as exc:
             raise cls.RepoNotFound(
                 f"Could not parse manifest.json file: {workspace.manifest_path()}:\n{exc}"
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -306,6 +306,9 @@ class Workspace:
     def root(self):
         return settings.WORKSPACE_DIR / self.name
 
+    def manifest_path(self):
+        return self.root() / "metadata/manifest.json"
+
     def get_id(self) -> str:
         return self.name
 
@@ -434,10 +437,12 @@ class CodeRepo:
     @classmethod
     def from_workspace(cls, workspace: Workspace, commit: str):
         try:
-            repo = workspace.manifest["repo"]
-        except (BusinessLogicLayer.ManifestFileError, KeyError):
+            repo = workspace.manifest.get("repo")
+            if repo is None:
+                repo = list(workspace.manifest["outputs"].values())[0]["repo"]
+        except (BusinessLogicLayer.ManifestFileError, IndexError, KeyError) as exc:
             raise cls.RepoNotFound(
-                "Could not parse manifest.json file: {manifest_path}:\n{exc}"
+                f"Could not parse manifest.json file: {workspace.manifest_path()}:\n{exc}"
             )
 
         try:
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -214,7 +214,10 @@ def create_repo(workspace, files=None, temporary=True):
 
     commit = response.stdout.strip()
     update_manifest(workspace)
-    workspace.manifest["repo"] = str(repo_dir)
+    if not workspace.manifest["outputs"]:
+        write_workspace_file(workspace, "foo.txt")
+
+    workspace.manifest["repo"] = None  # match job-runner output
     for output in workspace.manifest["outputs"].values():
         output["repo"] = str(repo_dir)
         output["commit"] = str(commit)
--- a/tests/functional/test_request_pages.py
+++ b/tests/functional/test_request_pages.py
@@ -133,12 +133,18 @@ def test_request_return(live_server, context, page, bll):
         "file2.txt",
         "file 2 content",
     )
-    factories.bll.set_status(release_request, RequestStatus.SUBMITTED, author)
+    bll.set_status(release_request, RequestStatus.SUBMITTED, author)
 
     return_request_button = page.locator("#return-request-button")
+    page.goto(live_server.url + release_request.get_url())
+
+    def _logout():
+        # logout by clearing cookies
+        context.clear_cookies()
 
     def _review_files(username):
-        # login as first output-checker
+        # logout current user, login as username
+        _logout()
         login_as_user(
             live_server,
             context,
@@ -172,7 +178,8 @@ def test_request_return(live_server, context, page, bll):
     # Return the request
     return_request_button.click()
 
-    # login as author again
+    # logout, login as author again
+    _logout()
     login_as_user(
         live_server,
         context,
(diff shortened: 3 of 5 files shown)


airlock/business_logic.py
```diff
@@ -407,10 +410,12 @@ class CommitNotFound(Exception):
     @classmethod
     def from_workspace(cls, workspace: Workspace, commit: str):
         try:
-            repo = workspace.manifest["repo"]
-        except (BusinessLogicLayer.ManifestFileError, KeyError):
+            repo = workspace.manifest.get("repo")
+            if repo is None:
+                repo = list(workspace.manifest["outputs"].values())[0]["repo"]
```
bloodearnest, line 415: I would prefer a different solution here. This API was written before we had manifest.json data, and I think I got it wrong, and it needs a refactor now that we do.

We only ever need a CodeRepo from the context of a specific output file in a workspace.

So I think we should get rid of `from_workspace(workspace: Workspace, commit: str)`, and instead have `from_workspace_output(workspace: Workspace, relpath: UrlPath)`. This looks up the output path in the manifest, and uses the repo/commit metadata found there to construct the CodeRepo object specific for this output. 
This means we don't use the top level `repo` key at all, and we don't need the related job-runner PR.

wdyt?

rebkwok: This doesn't entirely work (or at least, it only works for workspace files). This method is used for viewing code for both workspace output files and request files. A manifest only has the latest commit for each workspace file, but a request file could be an older version, based on a previous commit. 

I think we actually want a `from_file(workspace: Workspace, relpath: UrlPath, commit: str = None)` which gets the repo from the manifest for the file and optionally takes a commit (which will have come from a RequestFile), or if it's None, also look it up on the manifest.  

(RequestFileMetadata saves the repo as well as the commit when it's added to a request, but I don't think repo for a workspace should ever change)

rebkwok: No, I think it does need it for the views, nvm

bloodearnest: > This doesn't entirely work (or at least, it only works for workspace files). This method is used for viewing code for both workspace output files and request files. A manifest only has the latest commit for each workspace file, but a request file could be an older version, based on a previous commit.
> 
> I think we actually want a `from_file(workspace: Workspace, relpath: UrlPath, commit: str = None)` which gets the repo from the manifest for the file and optionally takes a commit (which will have come from a RequestFile), or if it's None, also look it up on the manifest.

How about two constructors?

`from_workspace(workspace, relpath)`

and

`from_request(release_request, relpath)`

> (RequestFileMetadata saves the repo as well as the commit when it's added to a request, but I don't think repo for a workspace should ever change)

It could in theory, if the repo is renamed by an admin during the lifetime of the project (I think this has happened). Which is one of the reasons to take it from the output metadata.  But its a very edge case.

Also, multiple workspaces can share the same repo, but with different branches.  Here, we only ever care about a specific commit, so we don't need branch at all.

rebkwok: Right, I've spent a while trying to figure this out, and i think it may have to stay as `from_workspace`.  It's perfectly straightforward to get the repo from an output or request file by looking up a relpath in the manifest. The problem is that it's only used in the code views, which have a path, but it's not the output/request path that the code comes from, it's the path to the file within the code repo. 

I can't put both the path of the code repo file and the path of the source file in the URL; I tried parsing the return_url, from which you can get the source relpath; but then you also need to pass it on to the contents view, and deal with the htmx requests that don't have the GET params; or you to have have a way to retrieve the repo without it, which ends up back at getting the repo from the first output file.

I can make it always retrieve the repo from the first output file though; they should always be the same for all outputs, and we shouldn't ever be trying to get a repo without an output file - so the job-runner PR still isn't required.

rebkwok: > How about two constructors?

No, still doesn't work with the code views. Retrieving it from the first output file for both workspace and request files (as it does now) handles the issue of a repo changing name, and both pass the commit when they build the code url. I think the current method is still the best way of doing this.

bloodearnest: > Right, I've spent a while trying to figure this out, and i think it may have to stay as `from_workspace`. It's perfectly straightforward to get the repo from an output or request file by looking up a relpath in the manifest. The problem is that it's only used in the code views, which have a path, but it's not the output/request path that the code comes from, it's the path to the file within the code repo.

> I can't put both the path of the code repo file and the path of the source file in the URL; I tried parsing the return_url, from which you can get the source relpath; but then you also need to pass it on to the contents view, and deal with the htmx requests that don't have the GET params; or you to have have a way to retrieve the repo without it, which ends up back at getting the repo from the first output file.

Ah, right, the current url structure bakes in the assumption of a workspace -> repo mapping.  I think maybe that is the issue - we should probably have the repo name in the url, rather than workspace? So we can just resolve it once for the output in question, and just re-use it for browsing after that? Something for later on.

> I can make it always retrieve the repo from the first output file though; they should always be the same for all outputs, and we shouldn't ever be trying to get a repo without an output file - so the job-runner PR still isn't required.

Right, that seems like a reasonable compromise. Sorry for sending you on a wild goose chase here :(

rebkwok: So yes, I also thought about having repo name in the URL instead of workspace, but then it's not straightforward to get the repo URL without the workspace (which has the manifest, with the full url). Presumably it should always be https://github.com/opensafely/<reponame> (except in tests) - but i think using the workspace is still probably more reliable. For now at least

rebkwok: Updated now to always use the first output; plus a change to the request_return functional test because it had started failing locally for me with a database locked error on this branch. Hoping that's also the fix for the intermittent CI errors too.

tests/factories.py
```diff
@@ -214,7 +214,10 @@ def create_repo(workspace, files=None, temporary=True):
 
     commit = response.stdout.strip()
     update_manifest(workspace)
-    workspace.manifest["repo"] = str(repo_dir)
+    if not workspace.manifest["outputs"]:
```
bloodearnest, line 217: Hmm, awkward workarounds like this do make me think that a root repo in the manifest may actually simplify a lot.

There's a history of what the root `repo` key used to be used for (github based releases), and I was hoping to not reuse it, as it's potentially brittle state. But I think the complexity it adds may be more costly than unlikely edge cases.

tests/functional/test_request_pages.py
```diff
@@ -133,12 +133,18 @@ def test_request_return(live_server, context, page, bll):
         "file2.txt",
         "file 2 content",
     )
-    factories.bll.set_status(release_request, RequestStatus.SUBMITTED, author)
+    bll.set_status(release_request, RequestStatus.SUBMITTED, author)
 
     return_request_button = page.locator("#return-request-button")
+    page.goto(live_server.url + release_request.get_url())
+
+    def _logout():
```
bloodearnest, line 141: could we just clear the browser cookies here? Same effect, but avoids a couple of navigations and page loads?

bloodearnest (approved):

airlock/business_logic.py
bloodearnest: Logging out is the fix to the db issue? That's interesting.  Something about sessions maybe?

tests/functional/test_request_pages.py
rebkwok: Maybe, I'll give it a try

tests/factories.py
rebkwok: Can a job run and create a manifest with no outputs? This is a slightly awkward workaround, but it's only for tests and made-up local manifests, to make them have at least one output. If real manifests will always have at least one output to look up, i don't think we need to worry about it. Plus, we should only ever be fetching the repo (in prod, that is) for either a workspace file or a request file, which should have a manifest entry.

rebkwok: I'm going to leave this as is for now and close the job-runner PR. We can revisit later if necessary.

bloodearnest: Yes, sorry, was a comment on the need for it because of other decisions, not a thing to be addressed. Sorry, should have made that clearer.

PR #395, opened 2024-06-17
Track and expose request file state in tree

bloodearnest:
- **Rename FileReviewStatus to UserFileReviewStatus**
- **Refactor FileReview.reviews to be a dict with username key**
- **Make request.output_files_set a dictionary**
- **Add RequestFileReviewStatus to capture the full state**
- **Clean up old code still using container.request_filetype**
- **Expose request states in request tree.**

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -349,7 +349,7 @@ class Workspace:
     def get_user_request_state(
         self, relpath: UrlPath, user: User
     ) -> UserFileReviewStatus | None:
-        return None
+        return None  # pragma: nocover
 
     def get_requests_url(self):
         return reverse(
@@ -524,7 +524,7 @@ class CodeRepo:
     def get_user_request_state(
         self, relpath: UrlPath, user: User
     ) -> UserFileReviewStatus | None:
-        return None
+        return None  # pragma: no cover
 
 
 @dataclass(frozen=True)
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -517,7 +517,7 @@ class RequestFile:
     relpath: UrlPath
     group: str
     file_id: str
-    reviews: list[FileReview]
+    reviews: dict[str, FileReview]
     timestamp: int
     size: int
     job_id: str
@@ -533,7 +533,10 @@ class RequestFile:
     def from_dict(cls, attrs) -> Self:
         return cls(
             **{k: v for k, v in attrs.items() if k != "reviews"},
-            reviews=[FileReview.from_dict(value) for value in attrs.get("reviews", ())],
+            reviews={
+                value["reviewer"]: FileReview.from_dict(value)
+                for value in attrs.get("reviews", ())
+            },
         )
 
     def approved_for_release(self):
@@ -544,7 +547,7 @@ class RequestFile:
             len(
                 [
                     review
-                    for review in self.reviews
+                    for review in self.reviews.values()
                     if review.status == UserFileReviewStatus.APPROVED
                 ]
             )
@@ -554,7 +557,7 @@ class RequestFile:
     def rejected_reviews(self):
         return [
             review
-            for review in self.reviews
+            for review in self.reviews.values()
             if review.status == UserFileReviewStatus.REJECTED
         ]
 
@@ -566,7 +569,7 @@ class RequestFile:
             len(
                 [
                     review
-                    for review in self.reviews
+                    for review in self.reviews.values()
                     if review.status
                     in [UserFileReviewStatus.APPROVED, UserFileReviewStatus.REJECTED]
                 ]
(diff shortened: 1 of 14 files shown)


airlock/business_logic.py
```diff
@@ -230,6 +241,14 @@ def get_file_metadata(self, relpath: UrlPath) -> FileMetadata | None:
     def get_workspace_state(self, relpath: UrlPath) -> WorkspaceFileState | None:
         """Get workspace state of file."""
 
+    def get_request_state(self, relpath: UrlPath) -> RequestFileReviewStatus | None:
```
rebkwok, line 244: nit: `get_request_status` for consistency with `RequestFileReviewStatus`?

airlock/business_logic.py
```diff
@@ -230,6 +241,14 @@ def get_file_metadata(self, relpath: UrlPath) -> FileMetadata | None:
...
+        """Get request status of file."""
+
+    def get_user_request_state(
```
rebkwok, line 247: nit: `get_user_request_status`?

airlock/business_logic.py
```diff
@@ -539,20 +574,35 @@ def from_dict(cls, attrs) -> Self:
             },
         )
 
-    def approved_for_release(self):
-        """
-        A file is approved for release if it has been approved by two reviewers
+    def status(self) -> RequestFileReviewStatus:
+        """The status of RequestFile, based on mutliple reviews.
+
+        We specificially only require 2 APPROVED votes, rather than all votes
```
rebkwok, line 580: s/specificially/specifically

airlock/business_logic.py
```diff
@@ -539,20 +574,35 @@ def from_dict(cls, attrs) -> Self:
...
+            return RequestFileReviewStatus.APPROVED
+
+        # do the reviews disagree?
+        if len(set(all_reviews)) > 1:
+            return RequestFileReviewStatus.CONFLICTED
+
+        # only case left is all reviews are REJECTED
+        return RequestFileReviewStatus.REJECTED
+
+    def status_for_user(self, user: User) -> UserFileReviewStatus | None:
```
rebkwok, line 601: I think this method actually belongs in the previous commit (no need to change though)

airlock/business_logic.py
```diff
@@ -539,20 +574,35 @@ def from_dict(cls, attrs) -> Self:
             },
         )
 
-    def approved_for_release(self):
-        """
-        A file is approved for release if it has been approved by two reviewers
+    def status(self) -> RequestFileReviewStatus:
```
rebkwok, line 577: Would it be worth calling this something like `get_status`, so it's clear that its a calculated status based on other things? We have a lot of `status`es that are attributes set on something. I mainly noted it because there's a change on line 603 that includes `.status` and then one on line 770 that includes `.status()` and I had to come back here to remind myself which sort of status this was.
Maybe the fact that you have to call it with parentheses is enough

tests/unit/test_business_logic.py
```diff
@@ -1343,11 +1344,13 @@ def test_release_request_add_same_file(bll):
     assert len(release_request.filegroups["default"].files) == 1
 
 
-def _get_current_file_reviews(bll, release_request, path, author):
+def _get_request_file(bll, release_request, path):
     """Syntactic sugar to make the tests a little more readable"""
     # refresh
-    release_request = bll.get_release_request(release_request.id, author)
-    return release_request.get_request_file_from_output_path(path).reviews
+    release_request = bll.get_release_request(
```
rebkwok, line 1350: There is `factories.refresh_release_request()` which will do the user-creating if necessary for you

tests/unit/test_business_logic.py
```diff
@@ -1539,14 +1583,49 @@ def test_reset_review_file_no_reviews(bll):
...
+    assert rfile.status_for_user(checker) is None
+    assert rfile.status() == RequestFileReviewStatus.INCOMPLETE
+
+
+@pytest.mark.parametrize(
+    "reviews, final_review",
+    [
+        (["APPROVED", "APPROVED"], "APPROVED"),
+        (["REJECTED", "REJECTED"], "REJECTED"),
+        (["APPROVED", "REJECTED"], "CONFLICTED"),
```
rebkwok, line 1603: Maybe also check the UNDECIDED status that a previously rejected file review could be in after a returned request is resubmitted?

rebkwok (approved):
:+1: Some questions, nothing blocking

airlock/business_logic.py
bloodearnest: ah, yes, good point

airlock/business_logic.py
bloodearnest: Yeah, I ended up doing lots of targeted fixups, suspect I got the wrong ones

airlock/business_logic.py
bloodearnest: That's a fair point, will do.

tests/unit/test_business_logic.py
bloodearnest: ooh, good shout

bloodearnest: Yeah, because UNDECIDED didn't exist when I started this, I'd missed handling it.

I've added a separate test, and UNDECIDED votes count as INCOMPLETE

PR #397, opened 2024-06-18
Autogenerate diagram for RequestStatus state machine

bloodearnest:
Integrate with just fix/check like other tools.

### DIFF
--- a/justfile
+++ b/justfile
@@ -106,6 +106,7 @@ check: devenv
     check "docker run --rm -i ghcr.io/hadolint/hadolint:v2.12.0-alpine < docker/Dockerfile"
     check "find docker/ airlock/ job-server -name \*.sh -print0 | xargs -0 docker run --rm -v \"$PWD:/mnt\" koalaman/shellcheck:v0.9.0"
     check "$BIN/mypy airlock/ local_db/"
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -121,7 +122,7 @@ mypy *ARGS: devenv
 
 
 # fix formatting and import sort ordering
```
inglesp, line 124: This comment's now out of date

inglesp (commented):
Great!

How are we going to make sure it doesn't get stale?  Currently it looks like it relies on a developer running `just fix` at the right time, and committing the changes.

bloodearnest:
> Great!
> 
> How are we going to make sure it doesn't get stale? Currently it looks like it relies on a developer running `just fix` at the right time, and committing the changes.

I included it in just check for exactly that reason - it will fail if its out of date. And we run `just check` in CI

justfile
bloodearnest: fixed

inglesp:
> I included it in just check for exactly that reason - it will fail if its out of date. And we run `just check` in CI

Ah yes, sorry, missed that

inglesp (approved):

PR #409, opened 2024-06-21
Disable CSV pagination

rebkwok:
Remove pagination, which doesn't affect load times since it was all being done on the frontend in datatables anyway.

JS updated to wait until an element of the loaded datatable is present before hiding the loading spinner (avoids displaying the unformatted table before the datatable has loaded).


bloodearnest (approved):
Nice, this should be a nicer load experience too.

PR #417, opened 2024-06-24
Improve styles for file tree

tomodwyer:
- Change background colour on selected file/folder
- Reduce horizontal spacing around tree elements
- Add arrows to show dropdown states
- Add colours for supporting and withdraw items

### DIFF
--- a/airlock/static/assets/file_browser/index.css
+++ b/airlock/static/assets/file_browser/index.css
@@ -45,22 +45,12 @@
   display: inline-block;
 }
 
-.tree__item > .supporting {
-  font-style: italic;
-}
-
-.tree__item > .withdrawn {
-  font-style: italic;
-  text-decoration: line-through;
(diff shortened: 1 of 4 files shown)


rebkwok:
In a request view, I have one supporting file (report.html) and one output file (bar.txt). When I go to the request, I'm looking at the top level of the request, so the top of the tree is highlighted blue.  The supporting file is yellow, with the darker vertical bar on the left, but the output file isn't highlighted, which makes it seem like the supporting file is important in some way (or at least, it took me a few seconds to figure out what the highlighting meant, despite having just read the PR description).

image

tomodwyer:
@rebkwok hmmm good point! It makes sense when each other file has a status too.

Let me remove that until we have the other colours and icons in place in the next PR.

rebkwok (approved):

PR #419, opened 2024-06-24
Add support for tracking 2 independent reviews in a "round"

bloodearnest:
- **Expose factories.review_file helper**
- **Refactor DAL get_requests functions**
- **Formalise who owns what RequestStatus**
- **Add a completed_reviews attribute to RequestMetadata**
- **WIP commit of adding the REVIEWED state and UI**
- **WIP fix of all the existing tests**


airlock/business_logic.py
```diff
@@ -1591,7 +1619,18 @@ def submit_request(self, request: ReleaseRequest, user: User):
...
         if request.status == RequestStatus.RETURNED:
+            audit = AuditEvent(
+                AuditEventType.REQUEST_REVIEW_RESET,
+                user=user.username,
+                request=request.id,
+            )
+            # reset completed review tracking
+            self._dal.reset_reviews(request.id, audit)
+
+            # any files that have not been updated are set to UNDECIDED
```
madwort, line 1633:
```suggestion
            # any unapproved files that have not been updated are set to UNDECIDED
```

airlock/business_logic.py
```diff
@@ -1669,6 +1708,48 @@ def reset_review_file(
...
+        release_request = self.get_release_request(release_request.id, user)
+        n_reviews = len(release_request.completed_reviews)
+
+        # this method is called twice, by different users. It advances the
+        # state differently depending on whether its the 1st or 2nd review to
+        # be completed.
+        if n_reviews == 1:
+            self.set_status(release_request, RequestStatus.PARTIALLY_REVIEWED, user)
+        elif n_reviews == 2:
+            self.set_status(release_request, RequestStatus.REVIEWED, user)
```
madwort, line 1751: this block is screaming 'race condition' to me, but we're hardly google-scale, maybe it's fine?

madwort: (to properly fix it we'd need to push some business logic out of the BLL & into the DAL??)

tests/factories.py
```diff
@@ -243,6 +248,145 @@ def create_release_request(workspace, user=None, **kwargs):
     return release_request
 
 
+def create_request_at_state(
```
madwort, line 251: I realise this is going to mean a lot of search-replace, but we renamed the enum from state to status, so....

```suggestion
def create_request_at_status(
```

tests/factories.py
```diff
@@ -243,6 +248,145 @@ def create_release_request(workspace, user=None, **kwargs):
...
+        request = refresh_release_request(request)
+
+        if (
+            status == RequestStatus.WITHDRAWN
+            and withdrawn_after == RequestStatus.RETURNED
+        ):
+            bll.set_status(request, RequestStatus.WITHDRAWN, author)
+            return refresh_release_request(request)
+        else:
+            return request
```
madwort, line 359: I like this guard-clause type pattern of returning, but we can get rid of some of these else clauses to tidy up 
```suggestion
    if status in [RequestStatus.RETURNED, RequestStatus.WITHDRAWN]:
        bll.set_status(request, RequestStatus.RETURNED, checker)
        request = refresh_release_request(request)

        if (
            status == RequestStatus.WITHDRAWN
            and withdrawn_after == RequestStatus.RETURNED
        ):
            bll.set_status(request, RequestStatus.WITHDRAWN, author)
            return refresh_release_request(request)
  
        return request
```

tests/factories.py
```diff
@@ -243,6 +248,145 @@ def create_release_request(workspace, user=None, **kwargs):
...
+
+    elif status == RequestStatus.REJECTED:
```
madwort, line 361:
```suggestion
    if status == RequestStatus.REJECTED:
```

tests/factories.py
```diff
@@ -243,6 +248,145 @@ def create_release_request(workspace, user=None, **kwargs):
...
+        bll.set_status(request, RequestStatus.REJECTED, checker)
+        return refresh_release_request(request)
+
+    bll.set_status(request, RequestStatus.APPROVED, checker)
+    request = refresh_release_request(request)
+
+    if status == RequestStatus.APPROVED:
+        return request
+
+    elif status == RequestStatus.RELEASED:
```
madwort, line 371:
```suggestion
    if status == RequestStatus.RELEASED:
```

tests/factories.py
madwort: actually I'd like it more if the branches were flipped, but we're really into code golf now!
```suggestion
    if status in [RequestStatus.RETURNED, RequestStatus.WITHDRAWN]:
        bll.set_status(request, RequestStatus.RETURNED, checker)
        request = refresh_release_request(request)

        if not (
            status == RequestStatus.WITHDRAWN
            and withdrawn_after == RequestStatus.RETURNED
        ):
            return request

        bll.set_status(request, RequestStatus.WITHDRAWN, author)
        return refresh_release_request(request)
```

airlock/business_logic.py
rebkwok: I guess it is possible; if R1 completes their review and status is set to partially reviewed, and at the same time R2 completes their review and by the time it tries to set status to partially reviewed, it's already partially reviewed, we get an invalid state transition error. Given that the states for partially reviewed/reviewed can be determined by the current state of the completed reviews, we could check those in the event of an invalid state transition here, and set it to the right one.

madwort: oh yeah, that's a good point, we could gracefully handle it if we want to. let's not worry about it for now, there's already a lot going on.

airlock/business_logic.py
```diff
@@ -877,19 +879,20 @@ def all_files_approved(self):
             for request_file in filegroup.output_files
         )
 
-    def all_files_reviewed(self):
+    def all_files_reviewed_by_reviewer(self, reviewer: User) -> bool:
         return all(
-            request_file.reviewed()
-            for filegroup in self.filegroups.values()
-            for request_file in filegroup.output_files
+            rfile.get_status_for_user(reviewer) is not None
```
madwort, line 884: as noted by @rebkwok, if the status is UNDECIDED, this will probably do the wrong thing

airlock/business_logic.py
rebkwok: I've added some handling of the race condition for 2 checkers approving simultaneously, although at the same time it occurred to me that when a request is fully approved it can be returned/released/rejected, so potentially a 3rd reviewer could attempt to complete a review at the same time that another review returns/releases/rejects and puts it into an invalid state for reviewing.  In that case, it'll just re-raise the exception - which I think is OK, given how unlikely it is to happen.

madwort (approved):

PR #420, opened 2024-06-24
Handle bad csvs

### LINKED ISSUE #416: Airlock failing to show file contents
From @andrewscolm via tech-support:

> airlock is failing to show the added files from the *_descriptive folder of [this request](https://github.com/ebmdatalab/opensafely-output-review/issues/884#issuecomment-2185257886). After 6 mins it is still saying "Loading table data.."

 * [Source Slack thread](https://bennettoxford.slack.com/archives/C065F8P58AW/p1719220282904469)
 * [Discussion Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1719220616589069)

rebkwok:
Fixes #416

simple-datatables can only render tables with equal numbers of columns in each row. If a CSV output file has different columns per row (usually because it's not proper tabulated data), the datatable will fail to initialised. This was resulting in the page appearing to hang on the "loading ..." message.

This is addressed in a couple of ways:
1) Adds a `setTimeout` in the JS which waits for 5 seconds, and if the datatable element hasn't been found by then, it just unhides the table element anyway (which will show the plain html table if datatables failed to load).
2) Checks the column counts per row in the CSV renderer, and stops datatables being used at all if the counts aren't the same (which avoids having to wait 5 seconds when we know there's going to be a problem).

This does mean that we're assuming all tables will take at most 5s to load, and we're reading all CSV rows in in the renderer. Given that we've now got a 5000 row limit on L4 CSV files, this shouldn't cause any problems.

### DIFF
--- a/airlock/renderers.py
+++ b/airlock/renderers.py
@@ -122,7 +122,12 @@ class CSVRenderer(Renderer):
     def context(self):
         reader = csv.reader(self.stream)
         headers = next(reader, [])
-        return {"headers": headers, "rows": reader}
+        header_col_count = len(headers)
+        rows = list(reader)
+        ctx = {"headers": headers, "rows": rows, "use_datatables": True}
+        if any(len(row) != header_col_count for row in rows):
+            ctx["use_datatables"] = False
+        return ctx
 
 
 class TextRenderer(Renderer):
--- a/airlock/static/assets/file_browser/csv.js
+++ b/airlock/static/assets/file_browser/csv.js
@@ -6,6 +6,7 @@ const observer = new MutationObserver((mutations, obs) => {
     document.querySelector("#csvtable p.spinner").style.display = "none";
     document.querySelector("#csvtable table.datatable").style.display = "table";
     obs.disconnect();
+    clearTimeout();
     return;
   }
 });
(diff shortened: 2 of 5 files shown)


evansd (approved):
It's not _great_ that we have to rely on a timeout to work out whether datatables has thrown an error, and then do some extra server-side work to prevent it from loading in the first place to avoid the timeout. But I don't think there's a way round this given how things currently work.

It looks like this is the code which sets up the table:
https://github.com/opensafely-core/job-server/blob/ea56d8b8fca912abedd68281bc5394f34f1b8b35/assets/src/scripts/_datatable.js#L67

I don't know if there's someway of providing an error handler hoook/event here which we could attach to? That would be a simpler, more robust and more responsive solution if there's some way of doing that. Might need some input from Tom'O'as, though.

rebkwok:
If we weren't relying on the job-server datatables JS, it might be easier to figure out (I briefly tried, but gave up because there are other more pressing things to do).I think most CSVs will be displayable as long as they don't have different numbers of columns, so the timeout is just a fallback, as that'll only be for other weirdness that datatables can't handle.

PR #421, opened 2024-06-24
Reinstate context and controls

### LINKED ISSUE #404: Reinstate context + controls
Given the low volume of current users, if necessary we can tell users to ignore the context + controls textboxes until we're ready to ditch the Word doc.

rebkwok:
Fixes #404 
SHOW_C3 setting now controls only whether comments are visible.


airlock/templates/file_browser/group.html
```diff
@@ -1,22 +1,21 @@
 {% #card title=path_item.name container=True %}
-  {% if show_c3 %}
```
rebkwok, line 2: The diff looks like lots has changed - actually it's just this line that's moved to line 18 so it wraps just the comments form.

evansd (approved):

PR #429, opened 2024-06-27
Remove hx attributes from return button

rebkwok:
The return request isn't an htmx one, so these attributes - probably a copy/paste error - were causing it to load the response into the button element.


evansd (approved):

madwort:
verified locally!

PR #439, opened 2024-07-01
Only show "add file(s)" buttons when a request is in pending or returned statuses

### LINKED ISSUE #438: Disable add file button when request is under review
Currently we're showing the option to add a file when the request is submitted or reviewed

rebkwok:
Fixes #438 

A file can only be added from a workspace with a current request if that request is in pending or returned status (i.e. author-editable ones). We already have appropriate checks and error messages if a file can't be added due to request status, but this disables the button.

Also adds similar checks for the "add files" multiselect and disables it if there is an active request in non-editable status. It doesn't check there are add-able files, since the modal already shows the file state for each selected file.

image

### DIFF
--- a/tests/functional/test_e2e.py
+++ b/tests/functional/test_e2e.py
@@ -75,12 +75,18 @@ def login_as(live_server, page, username):
     expect(page.locator("body")).to_contain_text(f"Logged in as: {username}")
 
 
-def tree_element_is_selected(page, locator):
+def assert_tree_element_is_not_selected(page, locator):
     # We need to wait for a tiny amount to ensure the js that swaps the
-    # selected classes around has run. An arbitrary 20ms seems to be enough.
+    # selected classes around has had time to run before we assert that the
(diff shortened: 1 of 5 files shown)


tests/functional/test_e2e.py
```diff
@@ -75,12 +75,18 @@ def login_as(live_server, page, username):
...
-    # selected classes around has run. An arbitrary 20ms seems to be enough.
+    # selected classes around has had time to run before we assert that the
+    # class isn't present. An arbitrary 20ms seems to be enough.
     page.wait_for_timeout(20)
-    classes = locator.get_attribute("class")
-    return "selected" in classes
+    expect(locator).not_to_have_class(re.compile("selected"))
+
+
+def assert_tree_element_is_selected(locator):
```
rebkwok, line 86: Not sure why these checks started failing on this PR, and only in the docker tests, but they should use the playwright `expect` anyway, which will let them rely on the default wait

bloodearnest (approved):
Nice, thanks

PR #441, opened 2024-07-02
Show status icons in file browser

tomodwyer:
- Show the status icon next to the status name in the file browser
- Simplify the status for supporting and withdrawn files

### DIFF
--- a/airlock/templates/file_browser/file.html
+++ b/airlock/templates/file_browser/file.html
@@ -51,7 +51,7 @@
           </form>
         {% endif %}
       {% elif is_output_checker %}
-        {% if not path_item.is_supporting and not path_item.is_withdrawn %}
+        {% if path_item.is_output %}
           <div>
             {% if not file_reset_review_url %}
               <span class="file-status file-status--unknown">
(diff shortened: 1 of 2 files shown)


airlock/templates/file_browser/file.html
```diff
@@ -43,15 +51,26 @@
...
-        <div>
-          {% if not file_reset_review_url %}
-            <span class="file-status file-status--warning">No status</span>
-          {% elif file_approve_url %}
-            <span class="file-status file-status--danger">Changes requested</span>
-          {% elif file_reject_url %}
-            <span class="file-status file-status--success">Approved</span>
-          {% endif %}
-        </div>
+        {% if not path_item.is_supporting and not path_item.is_withdrawn %}
```
rebkwok, line 54: This `if` shouldn't be needed; the file approve/reset/reject urls are all None if the filetype is supporting or withdrawn
https://github.com/opensafely-core/airlock/blob/85083f408980679472759a0bad397971ad044a73/airlock/views/request.py#L210

tomodwyer: When I remove that if statement, I get the "no status" showing:

tomodwyer: Lovely, that's worked.

rebkwok (approved):

PR #452, opened 2024-07-05
Hide votes and comments from output-checkers/users as needed

### LINKED ISSUE #327: Hide decisions on files from other users until the end of a review round
During a review round (ie while a request is in the `SUBMITTED` state), any of a reviewer's decisions (for now approving/rejecting a file, but in future any comments) in that round should be hidden from all other users (the researcher, and other output checkers).

Depends on #326.

- [x] model "Round"s
- [ ] hide comments based on the Round state

bloodearnest:
- **Remove old hack to make yourself the author in dev**
- **Add Visibility enum**
- **Add column to FileGroupComment model to track the visibility**
- **Record a comment's Visibility.**
- **Set comment visibility via UI.**
- **Only show the comments you are allowed for the group**
- **Unblind comments when independant review is completed.**
- **Apply visibility rules to file votes**
- **Show file votes based on visibility in the file view.**

Fixes #327 

This is not the final version, but it gets the basic structure for tracking vote/comment visibility in place.

There will be some UI finesse needed - exposing the icons in the tree to the author, more info and details about which output checker voted which way.

Also, we need to decide what the consolidation of comments look like, so currently all comments remain private, although output checkers can create a public comment if they wasnt to. But that's different issue really, so left TODO in this change.


bloodearnest:
Becky and I did an extensive walkthrough of this PR in a call together.

Here's my TODO list from that discussion:

- look at using a django migrate time default, rather than a model defualt
- rename get_comment_visibilities_for_user to get_writable_comment_visibilities_for_user
- add comment about why we re-audit-log comments when unblinding
- change dal set_comment_visibility to be a singale transactional that sets all comments *and* the status to REVIEW in the same transaction
- comments on each assertion set in the big unblinding test
- doc string for RequestFileStatus
- All the `_status` are confusing
   - instead adopt the term "decision" for the request file status that considers all votes.
   - s/.file_status/.decision
   - s/.user_review/.user_review
   - s/.get_status/.get_decision
- separate file-status/file-type css to avoid conflict and confusion

rebkwok (approved):

PR #453, opened 2024-07-05
Add file group to file review audit logs

### LINKED ISSUE #364: File approval review does not store group info in the audit log
as per https://github.com/opensafely-core/airlock/blob/main/airlock/business_logic.py#L1397-L1403

This makes some sense, as the groups only make sense in terms of the release request, not in terms of the outside world. However, it has the surprising consequence that file approvals appear in the release request audit log but not the file group audit log in the airlock UI & I think they should appear in both places.

Screenshot 2024-05-31 at 15 01 35
Screenshot 2024-05-31 at 15 01 09
ux

madwort:
Screenshot 2024-07-05 at 16 56 48

fixes #364

### DIFF
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -428,8 +428,11 @@ def file_approve(request, request_id, path: str):
     except bll.FileNotFound:
         raise Http404()
 
+    urlpath = UrlPath(path)
+    group_name = urlpath.parts[0]
+
     try:
-        bll.approve_file(release_request, relpath, request.user)
+        bll.approve_file(release_request, relpath, request.user, group_name)
     except bll.ApprovalPermissionDenied as exc:
         raise PermissionDenied(str(exc))
 
@@ -447,8 +450,11 @@ def file_reject(request, request_id, path: str):
     except bll.FileNotFound:
         raise Http404()
 
+    urlpath = UrlPath(path)
+    group_name = urlpath.parts[0]
+
     try:
-        bll.reject_file(release_request, relpath, request.user)
+        bll.reject_file(release_request, relpath, request.user, group_name)
     except bll.ApprovalPermissionDenied as exc:
         raise PermissionDenied(str(exc))
 
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -424,15 +424,12 @@ def file_approve(request, request_id, path: str):
     release_request = get_release_request_or_raise(request.user, request_id)
 
     try:
-        relpath = release_request.get_request_file_from_urlpath(path).relpath
+        request_file = release_request.get_request_file_from_urlpath(path)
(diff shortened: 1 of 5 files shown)


airlock/views/request.py
```diff
@@ -428,8 +428,11 @@ def file_approve(request, request_id, path: str):
     except bll.FileNotFound:
         raise Http404()
 
+    urlpath = UrlPath(path)
+    group_name = urlpath.parts[0]
```
rebkwok, line 432: If line 427 changed from:
```
relpath = release_request.get_request_file_from_urlpath(path).relpath
```
to:
```
request_file = release_request.get_request_file_from_urlpath
```
you can get the group name from the `request_file` instead of parsing the path.

rebkwok: In fact, you could pass the entire `request_file` to `approve_file`/`reject_file` instead of `relpath` and `group_name`

rebkwok (approved):

airlock/views/request.py
madwort: refactored, wdyt @rebkwok ?

tests/factories.py
```diff
@@ -436,6 +437,22 @@ def write_request_file(
         review_file(request, path, UserFileReviewStatus.REJECTED, *checkers)
 
 
+def create_request_file_bad(request_file, bad_path):
```
rebkwok, line 440: Could we name this `create_request_file_bad_path` so it's obvious what's bad about it?

tests/unit/test_business_logic.py
```diff
@@ -2190,7 +2200,30 @@ def test_approve_file(bll):
         release_request,
         AuditEventType.REQUEST_FILE_APPROVE,
         user=checker,
-        path=path,
+        path=request_file.relpath,
+        group="group",
+    )
+
+
+def test_approve_file_group(bll):
```
rebkwok, line 2208: `test_approve_file_group` (and `test_reject_file_group` below) sound like they're testing approving a filegroup, rather than testing that the group appears in the audit when a file is approved. Maybe something like `test_approve_file_audit_event`?

rebkwok (approved):
LGTM, just some naming things comments

tests/factories.py
madwort: yep done

tests/unit/test_business_logic.py
madwort: interesting - I've since modified the basic tests to verify the default group appears in the audit, this was testing that a manually specified group also appears in the audit log, which maybe made some sense when we were passing a group_name to the accept/reject methods - but now we're passing a RequestFile I actually don't think it adds anything. So I'm going to remove it entirely.

PR #506, opened 2024-07-10
Update css classes for combined file decision status

### LINKED ISSUE #489: Filetree showing incorrect status when two reviewers have completed their review
shows output checker's review status, and it should show the decision

rebkwok:
Fixes #489
Agreed reviews: 2 reviewers have approved:
Screenshot from 2024-07-10 16-54-40

Agreed reviews: 2 reviewers have requested changes:
Screenshot from 2024-07-10 16-55-16

Conflicted reviews: Current reviewer has approved, other reviewer has requested changes:
Screenshot from 2024-07-10 16-55-37

Author's view of conflicted file in request:
Screenshot from 2024-07-10 17-11-31

Author's view of under-review file in workspace (this is unchanged):
Screenshot from 2024-07-10 17-13-23

Author's view of updated file in workspace (this is unchanged):
image

After returning and resubmitting (with no changes, so previous request-changes decisions are now undecided) - shows the user review status:
image

Review has now approved and completed review (other reviewer has rejected and not yet completed):
image

Other reviewer has now completed, shows combined decision:
image

### DIFF
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -13,6 +13,7 @@ from airlock.business_logic import (
     ReleaseRequest,
     RequestFileReviewStatus,
     RequestFileType,
+    RequestStatus,
     UserFileReviewStatus,
     Workspace,
 )
@@ -215,14 +216,14 @@ class PathItem:
 
         if self.workspace_status:
             classes.append(f"workspace_{self.workspace_status.value.lower()}")
+        elif self.is_output():
+            if self.request_status:
+                classes.append(f"request_{self.request_status.value.lower()}")
 
-        if self.request_status:
-            classes.append(f"request_{self.request_status.value.lower()}")
-
-        if self.user_request_status:
-            classes.append(f"user_{self.user_request_status.value.lower()}")
-        else:
-            classes.append("user_incomplete")
+            if self.user_request_status:
+                classes.append(f"user_{self.user_request_status.value.lower()}")
+            else:
+                classes.append("user_incomplete")
 
         if self.request_filetype:
             classes.append(self.request_filetype.value.lower())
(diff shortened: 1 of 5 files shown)


airlock/static/assets/file_browser/tree.css
```diff
@@ -163,7 +163,9 @@
   --icon-url: url("/static/icons/attach_file.svg");
 }
 
-[data-output-checker][data-request]:not([data-author]) .tree__file.output {
+[data-request][data-request-independent] .tree__file.user_incomplete,
```
bloodearnest, line 166: I think we can maybe get rid of the data-request and data-request-independant attributes altogether?

The logic of what/whether a user can see a thing is done server side, all the css needs is the presence of user_* or request_* classes on the tree node?

bloodearnest (approved):

airlock/static/assets/file_browser/tree.css
rebkwok: Tried, didn't work (although I think maybe data-request could go)
The html_classes are added to a PathItem, based on `request_status` and `user_request_status`. (confusingly named, because `request_status` here is the combined `RequestFileReviewStatus` rather than the status of the release_request.  A PathItem can have both of these, and which one you actually show to a reviewer is dependent on the status of the request, and the PathItem doesn't know anything about that.

Although, thinking about it, maybe we can change the logic in the tree building before the PathItem is created, so it only gets one of user_request_status and request_status depending on what status the release request is in.  That would move the logic out of the template without PathItem having to know about requests.  I will investigate.

bloodearnest: Right - I think that was the conclusion I came to when talking it though with @tomodwyer, that we should have one status class, not two. Originially, I had though we might show both statuses in the tree, which is why both are added.

However, which we show would depend on the phase of the request, which we don't have yet (working on it!), so may need to wait till that lands.

rebkwok: OK, I have refactored the logic into the file browser API (haven't tackled turning it into a single status class yet), and removed it from the template, which is all working. However, I think I may have implemented a visibility bit that isn't quite there yet.
I ended up taking many screenshots, so I've put them in a [slideshow](https://docs.google.com/presentation/d/1w7r5mZGxIRO9qS-H6BHT6q6hfmevBRAFX_Ce27BDxwc/edit?usp=sharing) because it seemed the easiest way to walk through and check what things look like at each stage

airlock/file_browser_api.py
```diff
@@ -561,12 +562,29 @@ def build_path_tree(
...
-                        path, user
-                    )
+                if isinstance(container, ReleaseRequest):
+                    if container.status in [
+                        RequestStatus.SUBMITTED,
+                        RequestStatus.PARTIALLY_REVIEWED,
+                    ]:
+                        # if the request is in independent review phase, set the user review
+                        # status for the file
+                        if user:  # pragma: no cover
```
bloodearnest, line 573: Right, this is basically the same logic I have, but in a different place.

I can handle this when I rebase on this once its landed

bloodearnest (approved):

PR #509, opened 2024-07-12
Send notifications for status changes only

### LINKED ISSUE #461: Remove notifications for non-status transitions
Now that we have the concept of rounds, we only need to send notifications when a request changes status. We should remove notifications for:
- adding file
- withdrawing file
- adding context/controls
- editing context/controls

rebkwok:
fixes #461 
Remove notifications for updates. We now only send notification for status changes. I've kept the option to send updates with a notification but reworked it a little, so in future we can send additional update messages along with a status change notification. Job-server handles these in [this PR](https://github.com/opensafely-core/job-server/pull/4443).

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -2233,6 +2233,19 @@ class BusinessLogicLayer:
         )
         self._dal.audit_event(audit)
 
+    def validate_update_dicts(self, updates):
+        if updates is None:
+            return
+        allowed_keys = {"update", "group", "user"}
+        for update_dict in updates:
+            assert (
+                "update" in update_dict
+            ), "Notification updates must include an `update` key"
+            extra_keys = set(update_dict.keys()) - allowed_keys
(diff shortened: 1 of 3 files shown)


airlock/business_logic.py
```diff
@@ -131,15 +131,6 @@ class NotificationEventType(Enum):
     REQUEST_REJECTED = "request_rejected"
     REQUEST_RETURNED = "request_returned"
     REQUEST_RESUBMITTED = "request_resubmitted"
-    REQUEST_UPDATED = "request_updated"
-
-
-class NotificationUpdateType(Enum):
-    FILE_ADDED = "file added"
-    FILE_WITHDRAWN = "file withdrawn"
-    CONTEXT_EDITIED = "context edited"
```
madwort, line 140: `CONTEXT_EDITIED` 🧐

madwort (approved):

PR #517, opened 2024-07-12
Prevent submitting a release request with empty context & controls

### LINKED ISSUE #462: Submission should not be allowed without context + controls
On submission, if context or controls are missing, user should see a message saying which group is missing the context or controls.

We don't need anything fancier in the UI at this stage.

madwort:
* absolute bare minimum to make this correct
* lots of test updates
* fixes #462 

Screenshot 2024-07-12 at 15 24 45

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1559,6 +1559,16 @@ class BusinessLogicLayer:
                 f"cannot change status from {release_request.status.name} to {to_status.name}"
             )
 
+        if to_status == RequestStatus.SUBMITTED:
+            for filegroup in release_request.filegroups:
+                if (
+                    release_request.filegroups[filegroup].context == ""
+                    or release_request.filegroups[filegroup].controls == ""
+                ):
+                    raise self.RequestPermissionDenied(
+                        f"incomplete context and/or controls for filegroup '{filegroup}'"
+                    )
+
         # check permissions
         owner = release_request.status_owner()
         # author transitions
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1269,6 +1269,9 @@ class BusinessLogicLayer:
     class RequestPermissionDenied(APIException):
         pass
 
+    class IncompleteContextOrControls(RequestPermissionDenied):
+        pass
+
     class RequestReviewDenied(APIException):
         pass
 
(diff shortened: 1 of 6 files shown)


airlock/views/request.py
```diff
@@ -322,6 +322,9 @@ def request_submit(request, request_id):
     try:
         bll.submit_request(release_request, request.user)
     except bll.RequestPermissionDenied as exc:
+        if "incomplete context and/or controls" in str(exc):
```
madwort, line 325: maybe don't use string matching for this?

airlock/business_logic.py
```diff
@@ -1569,6 +1569,16 @@ def check_status(
                 f"cannot change status from {release_request.status.name} to {to_status.name}"
             )
 
+        if to_status == RequestStatus.SUBMITTED:
+            for filegroup in release_request.filegroups:
+                if (
+                    release_request.filegroups[filegroup].context == ""
+                    or release_request.filegroups[filegroup].controls == ""
+                ):
+                    raise self.RequestPermissionDenied(
```
madwort, line 1578: used `RequestPermissionDenied` here, because I think that's an appropriate response from `check_status()`, however that means we need to do string matching to catch it in `views/request.py` -> should I instead create a new Exception for this?

evansd: Yeah, I think the easiest thing would be to make a subclass of `RequestPermissionDenied`.

Another option would be to add an option `reason_code` attribute to the exception and then a new enum type to supply to it. If we start adding lots of different fine grained errors then that would probably be the way to do it. But I think it's too much plumbing when we've only got this one case.

evansd (approved):

airlock/business_logic.py
madwort: > Yeah, I think the easiest thing would be to make a subclass of RequestPermissionDenied

good shout, just done that

airlock/views/request.py
```diff
@@ -322,6 +322,9 @@ def request_submit(request, request_id):
     try:
         bll.submit_request(release_request, request.user)
     except bll.RequestPermissionDenied as exc:
+        if isinstance(exc, bll.IncompleteContextOrControls):
+            messages.error(request, str(exc))
+            return redirect(release_request.get_url())
         raise PermissionDenied(str(exc))
```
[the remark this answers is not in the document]
madwort: yes yes PR to follow

PR #524, opened 2024-07-12
Allow easy switching between users in local dev environments

rebkwok:
Gives us a menu option to switch between any of the available users in a `dev_users.json` file.

If `AIRLOCK_API_TOKEN` is set (i.e. in prod or prod-like envs - anything with a connected job-server), `AIRLOCK_DEV_USERS_FILE` is always None, there will be no DEV_USERS, and the menu item won't be visible.

No AIRLOCK_API_TOKEN in .env:
Screenshot from 2024-07-12 15-17-26

With AIRLOCK_API_TOKEN in .env:
image

### DIFF
--- a/airlock/nav.py
+++ b/airlock/nav.py
@@ -1,6 +1,7 @@
 from collections.abc import Callable
 from dataclasses import dataclass
 
+from django.conf import settings
 from django.http import HttpRequest
 from django.urls import reverse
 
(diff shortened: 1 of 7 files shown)


evansd (approved):
This is going to make life so much easier. Thanks for taking it on!

One small thought: given that we have the user tokens as well as the usernames at the point we render the switcher menu, could we just make each menu option a form which posts directly to the login view with the appropriate username and token? That way we don't need a separate switcher URL and switcher view.

These kinds of dev-mode auth bypass widgets are really useful to have but there's always a small worry about creating extra surface area for security vulnerabilities to sneak in. If we could handle this all via the existing login view then there's less extra stuff to worry about.

Not a major concern at this stage though, so happy to land this at it is.

evansd (approved):
Ah great, glad that worked out!

PR #533, opened 2024-07-15
Add user_can_review function

bloodearnest:
- **Factor our output-checker-but-not-author logic**
- **Drive by fix for functional test when there's local env vars**

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -871,7 +871,7 @@ class ReleaseRequest:
                 # already set - no one knows the current status
                 pass
             case ReviewTurnPhase.CONSOLIDATING:
-                # only output-checkers know the current status
+                # only users who can review this request know the current status
                 if can_review:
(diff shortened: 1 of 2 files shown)


airlock/business_logic.py
```diff
@@ -864,15 +864,15 @@ def get_request_file_status(
...
         decision = RequestFileDecision.INCOMPLETE
-        is_output_checker = user.output_checker and self.author != user.username
+        can_review = self.user_can_review(user)
 
         match phase:
             case ReviewTurnPhase.INDEPENDENT:
                 # already set - no one knows the current status
                 pass
             case ReviewTurnPhase.CONSOLIDATING:
                 # only output-checkers know the current status
```
madwort, line 874: should the comment match code? my wording is awful though
```suggestion
                # only users that can review can see the current status
```

rebkwok (approved):

PR #535, opened 2024-07-15
Show released files in workspace view

### LINKED ISSUE #381: Show in the workspace UI whether a file has already been released
Following on from #344 : 

Show (in the tree and in contents) that a file you are viewing has been already been released. 
Possibly also prevent a researcher from adding that file as an output file to a new request.

rebkwok:
Fixes #381 

Tree icon and colour for released files in workspace:
image

Tooltip for disabled button:
Screenshot from 2024-07-15 14-59-25

Multiselect add:
image


rebkwok:
Note: released files cannot be added as **either** output or supporting files. I could make it allow adding them as supporting files, but I haven't yet, given that we're planning to allow re-releasing, and it's unlikely to happen

local_db/data_access.py
```diff
@@ -99,6 +99,13 @@ def get_requests_for_workspace(self, workspace: str):
             )
         ]
 
+    def get_released_files_for_workspace(self, workspace: str):
+        return dict(
+            RequestFileMetadata.objects.filter(
+                request__workspace=workspace, released_at__isnull=False
+            ).values_list("relpath", "file_id")
```
bloodearnest, line 106: Question using relpath as the dict key here - what happens if the same relpath has been released twice?  Are we relying on the the default query order to mean the latest released file will win the key race?  I think this works, as the version on disk should be the latest?

If so, should we a) explicitly order the query to ensure this and b) have a comment explaining why?

Is there a way this could work if file_id was the key, which might avoid this particular issue? Or is that not actually what we want to look up on?

tests/factories.py
```diff
@@ -391,6 +391,15 @@ def create_request_at_status(
         return request
 
     if status == RequestStatus.RELEASED:
+        for relpath in request.output_files():
+            bll._dal.release_file(
```
bloodearnest, line 395: I get that we can't actually release files, so we can't call the BLL methods here.

I will try figure out a way in my current refactor.

bloodearnest (approved):
Left a comment about dict keys, probably just needs a comment?

local_db/data_access.py
rebkwok: Mmm true, it could be a list of previously released file ids instead. I suppose you could have released one version, changed it and changed it back again. At the time, I was thinking that since we don't allow releasing more than once, there'd only ever be one.
I think it could use the file_id as the key, that's probably the best way

rebkwok: Ah, if I use file_id it just needs to be a list of file ids, not relpaths. Then you also can't re-release the same file under a different path/name (which means a few more updates to tests).

rebkwok: I think this is what we want, because that's what job-server will refuse to release, right?

PR #538, opened 2024-07-16
Disable editing of context and controls for author when the request is under review

### LINKED ISSUE #537: Disable editing context/controls while request are under review
Context and controls should only be editable by authors when in pending or returned status

### LINKED ISSUE #539: Disable comments based on the current request owner
Currently anyone can add a comment, at any stage in the request process.
With the new turns, we want to limit all editing to only the current owner. The exception is during an author's turn, where we might want to allow a project collaborator who isn't the author to answer output-checkers' questions.

In author's turn:
- author can comment
- any user with access to the workspace can comment

In output-checker's turn:
- output-checker can comment

In final status:
- no-one can comment

rebkwok:
Disable editing of context/controls for author when the request is under review
Fixes #537

 Hide comments unless allowed to comment
 Fixes #539 
    Commenting is only allowed if a request is
    - not final
    - is under review, and the user is an output-checker
    - is in draft, and the user has permission to accesss the workspace

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1074,6 +1074,11 @@ class ReleaseRequest:
             BusinessLogicLayer.STATUS_OWNERS[self.status] == RequestStatusOwner.SYSTEM
         )
 
+    def is_in_draft(self):
+        return (
+            BusinessLogicLayer.STATUS_OWNERS[self.status] == RequestStatusOwner.AUTHOR
+        )
+
 
 def store_file(release_request: ReleaseRequest, abspath: Path) -> str:
     # Make a "staging" copy of the file under a temporary name so we know it can't be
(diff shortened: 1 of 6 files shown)


bloodearnest (approved):
LGTM

Minor nit: `is_in_draft` doesn't quite read right for me when the the request is RETURNED, but I don't have any better suggestions, really

rebkwok:
Yeah, I spent a while trying to come up with a decent name for it. `in_progress` isn't specific enough, `is_editing` and `is_under_edit` sounded weird, and didn't really feel like they covered the PENDING stage.  Is `is_editing` better that `is_in_draft`?

bloodearnest:
> Yeah, I spent a while trying to come up with a decent name for it. `in_progress` isn't specific enough, `is_editing` and `is_under_edit` sounded weird, and didn't really feel like they covered the PENDING stage. Is `is_editing` better that `is_in_draft`?

yeah, I think its a bit better if you feel like making the change.

PR #546, opened 2024-07-17
Alerts on the request pages when something needs to be actioned

### LINKED ISSUE #503: Tell user that they have to do something after completing second review
That _something_ can be returning to researcher or releasing files or rejecting request.

We can tell them in a message.

rebkwok:
Adds a new component for showing an alert message when an action is required on a request.

Currently these are:
1) Output checker has reviewed all files and needs to complete their review
Screenshot from 2024-07-17 10-58-01

2) Output checker completes the second review for a request and the request now needs to be progressed by returning, rejecting or releasing.

Releasing is available (all files approved twice):
Screenshot from 2024-07-17 10-57-21

Releasing not available (some files rejected by at one or both reviewers):
Screenshot from 2024-07-17 11-35-16

Fixes #503

### DIFF
--- a/airlock/templates/file_browser/index.html
+++ b/airlock/templates/file_browser/index.html
@@ -141,6 +141,15 @@
         {% endif %}
       {% endif %}
     {% /airlock_header %}
+
+    {% if request_action_required %}
+      <div class="mt-4">
+        {% #alert variant="warning" title="Action required" dismissible=True %}
+          {{ request_action_required }}
+        {% /alert %}
+      </div>
+    {% endif %}
+
   {% else %}
     {% #airlock_header context=context current_request=current_request title=title workspace=workspace return_url=return_url %}
     {% /airlock_header %}
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -199,22 +199,6 @@ def request_view(request, request_id: str, path: str = ""):
         user_has_completed_review = False
         user_has_reviewed_all_files = False
 
-    if (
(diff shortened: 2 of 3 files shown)


airlock/templates/file_browser/index.html
```diff
@@ -141,6 +141,17 @@
...
+
+    {% if release_request.is_under_review and user_has_reviewed_all_files and not user_has_completed_review %}
+      {% #airlock_todo_alert message="You have reviewed all files. You can now complete your review." %}{% /airlock_todo_alert %}
+    {% elif release_request.status.name == "REVIEWED" and is_output_checker and not is_author %}
+      {% if release_request.can_be_released %}
+        {% #airlock_todo_alert message="Two independent reviews have been completed. You can now return, reject or release this request." %}{% /airlock_todo_alert %}
+      {% else %}
+        {% #airlock_todo_alert message="Two independent reviews have been completed. You can now return or reject this request." %}{% /airlock_todo_alert %}
+      {% endif %}
+    {% endif %}
```
evansd, line 153: I wonder if it would be cleaner overall to pull this logic out into the view and have it just pass the message into the template?

You might not even need a separate component then, which would reduce the level of indirection. Just something like:
```html
{% if todo_message %}
  <div class="mt-4">
    {% #alert variant="warning" title="Action required" dismissible=True %}
      {{ todo_message }}
    {% /alert %}
  </div>
{% endif %}
```

evansd (approved):

PR #549, opened 2024-07-17
Handle invalid UTF-8 in plain text view

### LINKED ISSUE #473: make utf8 decoder safe for viewing any files with invalid characters as txt
For example PNGs.

We also need to handle eg CSV files with invalid encoding.

evansd:
This involves a bit of refactoring as there is now more than one argument to `open()` which we need to modify depending on whether we're in text or binary mode.

Closes #473

### DIFF
--- a/airlock/renderers.py
+++ b/airlock/renderers.py
@@ -53,12 +53,13 @@ class Renderer:
         if cache_id is None:
             cache_id = filesystem_key(stat)
 
+        if cls.is_text:
+            stream: IO[Any] = abspath.open("r", errors="replace")
+        else:
+            stream = abspath.open("rb")
+
         return cls(
-            stream=(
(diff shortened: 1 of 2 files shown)


airlock/renderers.py
```diff
@@ -53,12 +53,18 @@ def from_file(
...
+            stream = abspath.open("rb")
+
         return cls(
-            stream=(
-                abspath.open("r", errors="replace")
-                if cls.is_text
-                else abspath.open("rb")
-            ),
+            stream=stream,
+            # stream=(
```
rebkwok, line 63: Some leftover commented out things here

rebkwok (approved):

airlock/renderers.py
evansd: Gah, thanks!

PR #550, opened 2024-07-17
Only count votes from completed reviews in overall file decisions

### LINKED ISSUE #494: Only count votes from reviews that have been completed
This is important because disclosivity an only be assessed by considering all files in a release together.  We should make this clear in a comment to the code.

rebkwok:
Fixes #494 

It's trivial to only count votes in the current turn for completed reviews. However, when a request is returned to a user, we want the user to be able to see the decisions from the previous turn, so I needed to store the users who had completed a review at the end of each turn on the request.


airlock/business_logic.py
```diff
@@ -871,18 +879,27 @@ def get_request_file_status(
         phase = self.get_turn_phase()
         decision = RequestFileDecision.INCOMPLETE
         can_review = self.user_can_review(user)
+        completed_reviewers_this_turn = self.completed_reviews.keys()
 
+        # If we're in the AUTHOR phase of a turn (i.e. the request is being
+        # edited by the author, it's not in an under-review status), we need
+        # to show the decision from the previous turn, if there is one. For
+        # all other (reviewing) phases, we show the current decision based on
+        # te completed reviews in this turn.
```
madwort, line 888:
```suggestion
        # the completed reviews in this turn.
```

madwort:
At the moment, approval reviews persist across rounds & request for changes reviews are reset every round. Would this implementation mean effectively that a reviewer who has approved all files still has to hit the "complete" button every review round in order for their approval reviews to persist?

rebkwok:
> Would this implementation mean effectively that a reviewer who has approved all files still has to hit the "complete" button every review round in order for their approval reviews to persist?

This is already the case. The `start_new_turn` already set completed_reviews to an empty dict, so if you'd approved everything in the previous round, your file approvals remain, but you still have to click the complete button again.

madwort (approved):

PR #556, opened 2024-07-17
Refactor factories to use BLL for all state manipulation, rather than the DAL.

bloodearnest:
In a few places, we use the DAL directly from the test factories to manipulate state. This is convenient, but has two drawbacks:

1) It potentially allows invalid states to be created
2) It requires duplicating some logic, particularly the creation of event logs.

This change moves to using only BLL methods to manipulate state. This requires us to respect some limitations that previously we had ignored. Most of that is handled in the test factories, but in one case a test needed a fair bit more setup to get into the right state.

Note that we still use the dal in a couple of places in the test factories, for creating filegroups and audit events directly. However, the former is very simple, and the latter has no BLL equivalent.

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1901,7 +1901,7 @@ class BusinessLogicLayer:
         This currently uses the old api, and is shared amongst provider
         implementations, but that will likely change in future.
 
-        Passing upload=False will just performa the DAL actions, but not
+        Passing upload=False will just perform the DAL actions, but not
         actually attempt to upload the files. This defaults to True, and is
         primarily used for testing.
         """
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1895,11 +1895,15 @@ class BusinessLogicLayer:
         release_request.set_filegroups_from_dict(filegroup_data)
         return release_request
 
-    def release_files(self, release_request: ReleaseRequest, user: User):
+    def release_files(self, release_request: ReleaseRequest, user: User, upload=True):
         """Release all files from a release_request to job-server.
 
         This currently uses the old api, and is shared amongst provider
         implementations, but that will likely change in future.
+
+        Passing upload=False will just performa the DAL actions, but not
+        actually attempt to upload the files. This defaults to True, and is
+        primarily used for testing.
         """
 
         # we check this is valid status transition *before* releasing the files
@@ -1909,12 +1913,16 @@ class BusinessLogicLayer:
         self.validate_file_types(file_paths)
 
         filelist = old_api.create_filelist(file_paths, release_request)
-        jobserver_release_id = old_api.create_release(
-            release_request.workspace,
-            release_request.id,
-            filelist.json(),
-            user.username,
-        )
+
+        if upload:
+            jobserver_release_id = old_api.create_release(
+                release_request.workspace,
+                release_request.id,
+                filelist.json(),
+                user.username,
+            )
+        else:
+            jobserver_release_id = None
 
         for relpath, abspath in file_paths:
             audit = AuditEvent.from_request(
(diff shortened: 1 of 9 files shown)


tests/factories.py
```diff
@@ -454,6 +456,8 @@ def add_request_file(
     elif rejected:
         review_file(request, path, RequestFileVote.REJECTED, *checkers)
 
+    return refresh_release_request(request)
```
madwort, line 443: hells yes, this is a lot nicer

tests/factories.py
```diff
@@ -324,17 +325,18 @@ def create_request_at_status(
     )
 
     # add all files
-    for add_file in files:
-        add_file(request)
+    if files:
```
madwort, line 328: not sure why you need `if files:` here?

tests/factories.py
```diff
@@ -489,27 +477,89 @@ def review_file(request, relpath, status, *users):
...
         else:
             raise AssertionError(f"unrecognised status; {status}")  # pragma: no cover
 
 
+@dataclass
+class TestRequestFile:
+    """Placeholder containing file metadata.
+
+    Allows us to set up file states declaratively. Use the add() and vote()
+    methods can be called with a request in the right state.
```
madwort, line 500: this reads like it's been edited weirdly
```suggestion
    Allows us to set up file states declaratively. The add() and vote()
    methods can be called with a request in the right state.
```

tests/unit/test_business_logic.py
```diff
@@ -601,26 +615,35 @@ def test_provider_request_release_files(mock_old_api, mock_notifications, bll, f
...
     # - set status to submitted
     # - set status to partially reviewed
     # - set status to reviewed
+    # - set status to returned
+    # - set status to submitted
+    # - set status to partially reviewed
+    # - set status to reviewed
+    # - set status to returned
     # - set status to approved
     # - set status to released
```
madwort, line 635: do we really need this comment? `expected_notifications` looks quite self-explanatory?

tests/unit/test_business_logic.py
```diff
@@ -638,9 +661,20 @@ def test_provider_request_release_files(mock_old_api, mock_notifications, bll, f
...
         AuditEventType.REQUEST_SUBMIT,
-        # checker reviews
+        # initial reviews
+        AuditEventType.REQUEST_FILE_APPROVE,
+        AuditEventType.REQUEST_FILE_APPROVE,
+        AuditEventType.REQUEST_FILE_APPROVE,
+        AuditEventType.REQUEST_FILE_APPROVE,
+        AuditEventType.REQUEST_REVIEW,
+        AuditEventType.REQUEST_REVIEW,
+        # return reqjest
```
madwort, line 671:
```suggestion
        # return request
```

tests/unit/test_business_logic.py
```diff
@@ -638,9 +661,20 @@ def test_provider_request_release_files(mock_old_api, mock_notifications, bll, f
...
+        AuditEventType.REQUEST_RETURN,
+        # withdraw an resubmitt
```
madwort, line 673:
```suggestion
        # withdraw and resubmit
```

airlock/business_logic.py
```diff
@@ -1895,11 +1895,15 @@ def withdraw_file_from_request(
...
         return release_request
 
-    def release_files(self, release_request: ReleaseRequest, user: User):
+    def release_files(self, release_request: ReleaseRequest, user: User, upload=True):
         """Release all files from a release_request to job-server.
 
         This currently uses the old api, and is shared amongst provider
         implementations, but that will likely change in future.
+
+        Passing upload=False will just performa the DAL actions, but not
```
madwort, line 1904:
```suggestion
        Passing upload=False will just perform the DAL actions, but not
```

madwort (approved):
Yeah, some good improvements here, just a couple of typo nits

PR #557, opened 2024-07-17
Use a custom button for modal component buttons

rebkwok:
This means we can apply consistent styling to the non-modal buttons.


evansd (approved):
I wonder if it's worth making the button class customisable (and upstreaming that change to jobserver). I think it would just be a case of changing this line:
https://github.com/opensafely-core/airlock/blob/3a38428ef35050bc6a1cf224c643695752ed4306/assets/templates/_components/modal/modal.html#L5

To include:
```
class=button_class|default:"flex-shrink-0" 
```

That feels like it might be useful more generally, and would avoid adding complexity to the templates.

I'd be happy to merge this PR as is and then look at doing the simplification myself if you don't have time.

rebkwok:
> I wonder if it's worth making the button class customisable

Done, and at the same time realised we don't actually need it, we can use the button's small=True instead

PR #559, opened 2024-07-18
Throw exception instead of returning None if we have a bad path

madwort:


airlock/business_logic.py
```diff
@@ -415,11 +415,8 @@ def get_url(self, relpath: UrlPath = ROOT_PATH) -> str:
         return reverse("workspace_view", kwargs=kwargs)
 
     def get_workspace_file_status(self, relpath: UrlPath) -> WorkspaceFileStatus | None:
-        # defence in depth, we've been given a bad file path
-        try:
-            self.abspath(relpath)
-        except BusinessLogicLayer.FileNotFound:
-            return None
+        # this will throw FileNotFound if we have a bad file path
+        self.abspath(relpath)
```
inglesp, line 419: If we had a method called `validate_path_exists` which just called `abspath`, we could do away with the comment.

inglesp (approved):

airlock/business_logic.py
madwort: yes indeed, I think there's a few places we could use such a method

PR #562, opened 2024-07-19
Rename enum and method names for requesting changes to file

### LINKED ISSUE #380: Rename the enum & code from reject file to request changes to file
In #373 we changed the UI display of this functionality, which was on the critical path for the beta. We should tidy this up by migrating the ENUM to a matching value & renaming all methods etc.

inglesp:
Fixes #380


airlock/business_logic.py
```diff
@@ -170,7 +170,7 @@ class AuditEventType(Enum):
     REQUEST_FILE_UPDATE = "REQUEST_FILE_UPDATE"
     REQUEST_FILE_WITHDRAW = "REQUEST_FILE_WITHDRAW"
     REQUEST_FILE_APPROVE = "REQUEST_FILE_APPROVE"
-    REQUEST_FILE_REJECT = "REQUEST_FILE_REJECT"
+    REQUEST_FILE_REQUEST_CHANGES = "REQUEST_FILE_REQUEST_CHANGES"
```
madwort, line 173: am I right in thinking this old value is written into the database & we therefore need a data migration?

```
sqlite> select * from local_db_auditlog where type = 'REQUEST_FILE_REJECT'
   ...> ;
32|output_checker_2|example-workspace|01J2KN9GGT60GSEBZT7XB0989N|output/rows.csv|{"group": "default"}|2024-07-12 22:43:00.674452|REQUEST_FILE_REJECT
54|output_checker_2|example-workspace|01J2KN9GGT60GSEBZT7XB0989N|output/rows.csv|{"group": "default"}|2024-07-17 14:04:09.558955|REQUEST_FILE_REJECT
```

madwort (changes requested):
would like to check if this is safe to deploy

airlock/business_logic.py
inglesp: > am I right...

looks like it!

rebkwok (approved):
This LGTM now :+1:

madwort (approved):

PR #563, opened 2024-07-19
Support re-adding withdrawn files

### LINKED ISSUE #375: Withdrawn files cannot be re-added to a release request
I'm sure whether this is required, but it feels like it might be at some point (e.g. a researcher accidentally withdraws the wrong file). Originally implemented in https://github.com/opensafely-core/airlock/issues/138

evansd:
Closes #375


tests/functional/test_e2e.py
```diff
@@ -476,6 +476,40 @@ def test_e2e_update_file(page, live_server, dev_users):
...
+    # Change our mind: go to workspace page and re-add it
+    workspace = bll.get_workspace("test-workspace", author)
+    page.goto(live_server.url + workspace.get_url("subdir/"))
+    find_and_click(page.locator('input[name="selected"]'))
+    find_and_click(page.locator("button[value=add_files]"))
+    find_and_click(page.get_by_role("form").locator("#add-file-button"))
+
+    # Confirm it's been re-added
+    expect(page.locator("body")).to_contain_text(
+        "Output file has been updated in request"
```
madwort, line 509: very tiny nit that ideally this would say "added" rather than "updated", but...

evansd: No, good call – I was being lazy.

PR #564, opened 2024-07-19
Getting output_checker info as a non-output checker throws an exception

madwort:


tests/unit/test_business_logic.py
```diff
@@ -759,25 +759,29 @@ def test_provider_get_outstanding_requests_for_review(output_checker, expected,
...
         # they did not author
-        (True, ["r1"]),
+        True,
     ],
 )
-def test_provider_get_returned_requests(output_checker, expected, bll):
+def test_provider_get_returned_requests(output_checker, bll):
     user = factories.create_user("test", ["workspace"], output_checker)
     other_user = factories.create_user("other", ["workspace"], False)
-    output_checker = factories.create_user("other-checker", ["workspace"], True)
```
madwort, line 780: nice bit of unused variable shadowing, tripped me up for a minute

inglesp (approved):
An improvement.  Perhaps better would be a simple `assert False`, to indicate that we expect that these lines getting hit indicates an error in our logic rather than the user doing something strange.

That is: `XXXPermisionDenied` errors indicate that the user is trying to do something that they can't (say, manually fiddling with a URL) while `assert False` indicate things that should never be hit.

madwort:
Hmm, so you would `assert False` as well as `return []`? I know we run production with asserts on, and we've no plans to change that, but I find it hard not to do the defensive thing here & assume that asserts could be turned off one day. 

I was also assuming that this could be potentially hit by a user trying to do something they can't - i.e. they manually fiddle with a URL & then there's an error in our logic which means that this method is called - in which case I think `assert False` isn't really enough. 

I might be more keen on `assert False` if the topic wasn't permissions / information leakage, and/or if I had some guarantees about where this code being called from.

PR #567, opened 2024-07-22
Fix "Add File to Request" button for withdrawn files

evansd:
This was behaviour that should have been in #563


rebkwok (approved):
Are we missing a test somewhere that would have caught this?

evansd:
Yep! Thanks Becky. I shouldn't even try to code in fuzzy-brain state. Have extended the existing test to cover this.

Issue #512, opened 2024-07-12
public comments wrongly classified as private in the output checker view


madwort:

inglesp:
Simon thinks this is a styling bug

bloodearnest:
Narrator: it was not a styling bug.

PR #570, opened 2024-07-22
Filter out audit logs that a user cannot see

### LINKED ISSUE #534: Extend visibility logic to audit log
In #452 we introduced the concept of a "turn", that batches up a set of in-progress comments/votes, and only shows them to the appropriate other user's when the turn is completed.

We should extend this to the audit log. We should record the turn an audit log was recorded in, and use the same logic in filtering comments to filter out audit logs the current user should not yet be able to see. Then that turn is completed, then those audit logs will become visible.

Audit logs for *private* comments should never be shown the user.

bloodearnest:
This change ensures that users don't see audit messages that they shouldn't see.

It uses exactly same logic as previous work hiding comments/votes from users. So, this means

 - Request author and other non-output-checkers cannot see private comment audit logs from any turn
 - When in INDEPENDENT phase, output-checkers cannot see each others audit logs for votes/comments from the current turn
 - When in CONSOLIDATING phase, output-checkers can see each others audit logs for votes/comments from the current turn, but the author cannot
 - When in AUTHOR phase, authors can now see the audit logs votes and *public* comment from the previous turn.
 - When in COMPLETE phase, all items from all turns are visible, subject to private/public rules

Fixes #534

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -338,7 +338,10 @@ def filter_visible_items(
         match item.visibility:
             case Visibility.PUBLIC:
                 # can always see public items from previous rounds
-                if item.review_turn < current_turn:
+                if (
+                    item.review_turn < current_turn
+                    or current_phase == ReviewTurnPhase.COMPLETE
+                ):
                     yield item
                 # can see public items for other users if CONSOLIDATING and can review
                 elif current_phase == ReviewTurnPhase.CONSOLIDATING and user_can_review:
@@ -347,7 +350,10 @@ def filter_visible_items(
                 # have to be able to review this request to see *any* private items
                 if user_can_review:
                     # can always see private items from previous rounds
-                    if item.review_turn < current_turn:
+                    if (
+                        item.review_turn < current_turn
+                        or current_phase == ReviewTurnPhase.COMPLETE
+                    ):
                         yield item
                     # can only see private items from current round if we are not INDEPENDENTe
                     elif current_phase != ReviewTurnPhase.INDEPENDENT:
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -2286,19 +2286,16 @@ class BusinessLogicLayer:
             release_request.id, group, comment_id, user.username, audit
         )
 
-    def get_audit_log(
+    def get_request_audit_log(
         self,
-        user: str | None = None,
-        workspace: str | None = None,
-        request: str | None = None,
+        user: User,
+        request: ReleaseRequest,
         group: str | None = None,
         exclude_readonly: bool = False,
         size: int | None = None,
     ) -> list[AuditEvent]:
         return self._dal.get_audit_log(
-            user=user,
-            workspace=workspace,
-            request=request,
+            request=request.id,
             group=group,
             exclude=READONLY_EVENTS if exclude_readonly else set(),
             size=size,
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -251,6 +251,8 @@ class AuditEvent:
         **kwargs: str,
     ):
         # Note: kwargs go straight to extra
+        # set review_turn from request
+        kwargs["review_turn"] = str(request.review_turn)
         event = cls(
             type=type,
             user=user.username,
@@ -282,6 +284,21 @@ class AuditEvent:
     def description(self):
         return AUDIT_MSG_FORMATS[self.type]
 
+    @property
+    def review_turn(self) -> int:
+        return int(self.extra.get("review_turn", 0))
+
+    @property
+    def author(self) -> str:
+        return self.user
+
+    @property
+    def visibility(self) -> CommentVisibility:
+        v = self.extra.get("visibility")
+        if v:
+            return CommentVisibility[v.upper()]
+        else:
+            return CommentVisibility.PUBLIC
 
 
 class VisibleItem(Protocol):
@@ -2243,6 +2260,7 @@ class BusinessLogicLayer:
             group=group,
             comment=comment,
             review_turn=str(release_request.review_turn),
+            visibility=visibility.name,
         )
 
         self._dal.group_comment_create(
@@ -2294,11 +2312,23 @@ class BusinessLogicLayer:
         exclude_readonly: bool = False,
         size: int | None = None,
     ) -> list[AuditEvent]:
-        return self._dal.get_audit_log(
+        """Fetches the audit log for this request, filtering for what the user can see."""
+
+        audits = self._dal.get_audit_log(
             request=request.id,
             group=group,
             exclude=READONLY_EVENTS if exclude_readonly else set(),
             size=size,
+        )
+
+        return list(
+            filter_visible_items(
+                audits,
+                request.review_turn,
+                request.get_turn_phase(),
+                request.user_can_review(user),
+                user,
+            )
         )
 
     def audit_workspace_file_access(
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -131,7 +131,6 @@ class Visibility(Enum):
     def description(self):
         return self.choices()[self]
 
-    @cached_property
     def independent_description(self):
         return "Only visible to you until both reviews submitted"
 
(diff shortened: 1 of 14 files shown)


tests/unit/test_business_logic.py
```diff
@@ -2138,12 +2143,67 @@ def setup_empty_release_request():
     return release_request, path, author
 
 
-def test_get_visible_comments_for_group(bll):
+@dataclass
+class VisibleItems:
+    """Helper class to make assertions about visiblity of comments and audit logs."""
+
```
rebkwok, line 2149: This class made sense to me after I'd read its use in the tests, but not really before. It might just need some better comments to explain what comments/audits are here - the visible items for a specific group and user at a stage in a request

tests/unit/test_business_logic.py
```diff
@@ -2267,110 +2340,122 @@ def get_comments(user):
...
+    assert not visible.comment("turn 1 checker 0 private", checkers[0])
+    assert visible.vote(RequestFileVote.APPROVED, checkers[0])
+    assert not visible.comment("turn 1 checker 1 private", checkers[1])
+    assert visible.vote(RequestFileVote.APPROVED, checkers[1])
+    assert visible.comment("turn 1 checker 0 public", checkers[0])
+    assert visible.comment("turn 2 author public", author)
+    assert not visible.comment("turn 3 checker 0 private", checkers[0])
+    assert visible.vote(RequestFileVote.REJECTED, checkers[0])
+    assert not visible.comment("turn 3 checker 1 private", checkers[1])
+    assert visible.comment("turn 3 checker 0 public", checkers[0])
```
rebkwok, line 2458: This is missing a test of comments in a turn before the request moves into a final state.  i.e. comments made just before a request is rejected/released.

airlock/business_logic.py
```diff
@@ -283,6 +283,61 @@ def description(self):
...
+    @property
+    def visibility(self) -> CommentVisibility:
+        raise NotImplementedError()
+
+
+def filter_visible_items(
+    items: Sequence[VisibleItem],
+    current_turn: int,
+    current_phase: ReviewTurnPhase,
+    can_see_private: bool,
```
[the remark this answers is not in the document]
bloodearnest: Good call, will rename

airlock/business_logic.py
```diff
@@ -283,6 +283,61 @@ def description(self):
...
+            yield item
+            continue
+
+        match item.visibility:
+            case CommentVisibility.PUBLIC:
+                # can always see public comments from previous rounds
+                if item.review_turn < current_turn:
+                    yield item
+                elif current_phase == ReviewTurnPhase.CONSOLIDATING and can_see_private:
+                    yield item
```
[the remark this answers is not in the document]
bloodearnest: That's a good point, I think we might not be handling this correctly atm.

One option is to increment the turn when we approve/reject a request, which I don't think we do atm.  That would make all previous turn things visible.

Another is to actually use COMPLETE phase in this filtering logic.

Leaning towards the latter, I think.

airlock/business_logic.py
```diff
@@ -283,6 +283,61 @@ def description(self):
...
+        if item.author == user.username:
+            yield item
+            continue
+
+        match item.visibility:
+            case CommentVisibility.PUBLIC:
+                # can always see public comments from previous rounds
+                if item.review_turn < current_turn:
+                    yield item
+                elif current_phase == ReviewTurnPhase.CONSOLIDATING and can_see_private:
```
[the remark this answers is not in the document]
bloodearnest: yes, that is correct.

Output checkers can see public comments added by all output-checkers in the current turn, but researcher cannot

Does this read better with the s/can_see_private/user_can_review/` change you suggested?

I will also add a comment to clarify.

bloodearnest:
> I realise that some of the questions are about previous comment-visibility functionality that isn't specific to the audit log, but given that this is refactoring the previous visibility code, I think it's probably worth fixing here.

Right, there is separate ticket for this, think https://github.com/opensafely-core/airlock/issues/512

For the first 3 screen shots, I think the actual visiblity is correct (i.e. whether its shown or not)? But the text explaining the visibility is all kinds of wrong, which the issue above is tracking.

For the 4th screenshot, that is a bug, as noted in my reply to another comment: https://github.com/opensafely-core/airlock/pull/570#discussion_r1687881914

airlock/business_logic.py
```diff
@@ -282,6 +283,21 @@ def __str__(self):
...
+    @property
+    def review_turn(self) -> int:
+        return int(self.extra.get("review_turn", 0))
+
+    @property
+    def author(self) -> str:
+        return self.user
+
+    @property
+    def visibility(self) -> CommentVisibility:
```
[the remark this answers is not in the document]
bloodearnest: Sure

airlock/business_logic.py
```diff
@@ -251,6 +251,7 @@ def from_request(
         **kwargs: str,
     ):
         # Note: kwargs go straight to extra
+        kwargs.setdefault("review_turn", str(request.review_turn))
```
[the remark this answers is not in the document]
bloodearnest: I was thinking that in test scenarios we might want control it.

But I ended up adding all the test coverage in the big integration test, so didn't need it. I will just force set it.

tests/unit/test_business_logic.py
bloodearnest: Good catch, it is, which is why we've not caught the bug about rejected requests visibility.

airlock/business_logic.py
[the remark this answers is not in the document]
rebkwok: Yes, I think it's better if it's called `user_can_review`

airlock/business_logic.py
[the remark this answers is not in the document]
rebkwok: Agree, I think using COMPLETE to control the visibility is better - that change of status (to a final status) isn't really a new turn as we've been thinking about it

airlock/business_logic.py
```diff
@@ -282,6 +285,83 @@ def __str__(self):
...
+        # you can always see things you've authored. Doing this first
+        # simplifies later logic, and avoids potential bugs with users adding
+        # items but then they can not see the item they just added
+        if item.author == user.username:
+            yield item
+            continue
+
+        match item.visibility:
+            case Visibility.PUBLIC:
+                # can always see public items from previous rounds
```
rebkwok, line 341:
```suggestion
                # can always see public items from previous turns and completed turns
```

airlock/business_logic.py
```diff
@@ -282,6 +285,83 @@ def __str__(self):
...
+            case Visibility.PRIVATE:
+                # have to be able to review this request to see *any* private items
+                if user_can_review:
+                    # can always see private items from previous rounds
+                    if (
+                        item.review_turn < current_turn
+                        or current_phase == ReviewTurnPhase.COMPLETE
+                    ):
+                        yield item
+                    # can only see private items from current round if we are not INDEPENDENTe
```
rebkwok, line 359:
```suggestion
                    # can only see private items from current turn if we are not INDEPENDENT
```

airlock/business_logic.py
```diff
@@ -282,6 +285,83 @@ def __str__(self):
...
+                    or current_phase == ReviewTurnPhase.COMPLETE
+                ):
+                    yield item
+                # can see public items for other users if CONSOLIDATING and can review
+                elif current_phase == ReviewTurnPhase.CONSOLIDATING and user_can_review:
+                    yield item
+            case Visibility.PRIVATE:
+                # have to be able to review this request to see *any* private items
+                if user_can_review:
+                    # can always see private items from previous rounds
```
rebkwok, line 353:
```suggestion
                    # can always see private items from previous turns
```

airlock/business_logic.py
```diff
@@ -124,16 +124,17 @@ class CommentVisibility(Enum):
...
         }
 
     def description(self):
         return self.choices()[self]
 
-    @cached_property
     def independent_description(self):
-        return "Only visible to you until both reviews completed"
+        return (
+            self.description() + " (Only visible to you until both reviews completed)"
```
rebkwok, line 136: I think this wording needs some thought. It makes the visibility pill very long, and it always stays visible, so at smaller screen sizes it covers up the username. Would " (pending)"  be enough, if users have an understanding already of how the independent review process works? Or some sort of icon?  

image

rebkwok (approved):
A few nits, that's all

airlock/business_logic.py
bloodearnest: Yeah, I wasn't happy with this, but I didn't try reducing width, which is clearly not going to work.

I've used Pending is to mean "won't be visible until the turn is complete", and Hidden to mean "hidden from other output checkers until 2 review complete", aka blinded.

Maybe the way forward here is to have pill that says either "Private" or "Public", and another pill that says "Hidden" (or perhaps "Blinded"), but with tooltips to have enough space to clearly explain what we mean, rather than trying to cram the text in a pill?

rebkwok: Yeah, I think one(ish) word pills with tooltips would be better

bloodearnest: Given we haven't got tooltips sorted yet, I'll change this to one word, and add ticket for adding tooltips

PR #583, opened 2024-07-26
Filter audit logs for visibility

bloodearnest:
This is an identical resubmission of https://github.com/opensafely-core/airlock/pull/570

I accidentally merged that then reverted. This is that with clean history, and rebased on all the main change since then.

- **Factor out the core visibility logic to be reusable**
- **Remove activity log from workspace.**
- **Switch tests that assert against audit logs to use the DAL**
- **Rename get_audit_log to get_request_audit_log**
- **Filter audit logs by visibility**
- **Cover the new filtered audit logs in tests**
- **Clarify filter_visible_items logic**
- **Rename CommentVisibility to Visibility**
- **Fix visible logic when phase is COMPLETE**
- **Fix bug in comment descriptions and class**
- **Refactor the comment visiblity pills**
- **Fix comment typos**

### DIFF
--- a/tests/unit/test_business_logic.py
+++ b/tests/unit/test_business_logic.py
@@ -1,6 +1,7 @@
 import hashlib
 import inspect
 import json
+from dataclasses import dataclass
 from hashlib import file_digest
 from io import BytesIO
 from pathlib import Path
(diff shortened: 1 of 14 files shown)


tests/unit/test_business_logic.py
```diff
@@ -2148,12 +2155,71 @@ def setup_empty_release_request():
     return release_request, path, author
 
 
-def test_get_visible_comments_for_group(bll):
+@dataclass
+class VisibleItemsHelper:
+    """Helper class to make assertions about visiblity of comments and audit logs.
+
+    It will fetch comments and audits that are visible a specific request and
```
rebkwok, line 2162:
```suggestion
    It will fetch comments and audits that are visible for a specific request and
```

rebkwok (approved):
One tiny comment nit, otherwise all looks good

PR #584, opened 2024-07-26
Prevent resetting of file votes after a review has been submitted

rebkwok:
After a user submits a review, the request changes status (to PARTIALLY_REVIEWED or REVIEWED). In these statuses, users can still change their votes. However, we want to prevent them from re-setting votes to undecided, to avoid the possibility of returning a request with incomplete file reviews.

Also includes some refactoring of the voting button logic in the view to make the template logic simpler.

### DIFF
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -270,6 +270,10 @@ def request_view(request, request_id: str, path: str = ""):
                 assert False, "Invalid RequestFileVote value"
         else:
             file_reset_review_url = None
+        
+        # Disable reset button for submitted reviews
+        if user_has_submitted_review:
+            file_reset_review_url = None
 
     if (
         release_request.is_under_review()
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -1,3 +1,5 @@
+from typing import Any, Dict
+
 import requests
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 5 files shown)


tests/unit/test_business_logic.py
```diff
@@ -2797,6 +2797,41 @@ def test_reset_review_file_no_reviews(bll):
...
 
+def test_reset_review_file_after_review_submitted(bll):
+    path = Path("path/file1.txt")
+    release_request = factories.create_request_at_status(
+        "workspace",
+        status=RequestStatus.SUBMITTED,
+        files=[factories.request_file(path=path)],
+    )
+    checker = factories.create_user(output_checker=True)
+    # request_file = release_request.get_request_file_from_output_path(path)
```
madwort, line 2808:
```suggestion
```

airlock/views/request.py
```diff
@@ -235,45 +237,49 @@ def request_view(request, request_id: str, path: str = ""):
...
+        existing_review_status = existing_review.status if existing_review else None
+        match existing_review_status:
+            case RequestFileVote.APPROVED:
+                voting_buttons["approve"]["disabled"] = True
+            case RequestFileVote.REJECTED:
+                voting_buttons["reject"]["disabled"] = True
+            case RequestFileVote.UNDECIDED | None:
+                voting_buttons["reset_review"]["disabled"] = True
+            case _:  # pragma: no cover
                 assert False, "Invalid RequestFileVote value"
```
madwort, line 278: random question about this block - if there's no existing review, and `existing_review_status` is `None`, then match existing_review_status will hit the case _: branch & raise an exception? In which case, do we need the `existing_review.status if existing else None` clause at all?

madwort (approved):
I like the refactor of the voting button logic!

airlock/views/request.py
rebkwok: If there's no existing review, `existing_review_status` will always be None, and it'll hit the `case RequestFileVote.UNDECIDED | None:` branch. 

The `case _:` branch shouldn't ever be hit, because we've checked all possible values of `RequestFileVote`, but we have a bunch of places where we have that catchall just in case. I guess we might add another RequestFileVote value in future and this would catch it.

madwort: ah I mis-read it!

rebkwok:
> I like the refactor of the voting button logic!

Good! I'm planning to have a look for some more candidates for this sort of thing next week along with refactoring the permissions out :)

PR #585, opened 2024-07-26
Change file "rejected" to "changes requested"

rebkwok:
This should do the rest of #380 
We've previously changed the voting wording on the frontend from "reject" to "request changes", and we've updates some of the method names, and the AuditEventType Enum. This does the RequestFileVote and RequestFileDecision enums, the view/url names, tests and all comments I could find that refer to rejecting files.


tests/factories.py
```diff
@@ -292,7 +292,7 @@ def create_request_at_status(
     output-checker. If not provided, a default output checker is used.
 
     Optionally, `request_file` can be given `checkers`, a list of users who will
-    review (approve/reject) the file. If not provided, default output checkers
+    review (approve/request changes tp) the file. If not provided, default output checkers
```
madwort, line 295:
```suggestion
    review (approve/request changes to) the file. If not provided, default output checkers
```

madwort (approved):

PR #586, opened 2024-07-26
Do not allow changing group & filetype when updating file

madwort:


airlock/business_logic.py
```diff
@@ -1769,6 +1769,31 @@ def update_file_in_request(
         release_request: ReleaseRequest,
         relpath: UrlPath,
         user: User,
+    ) -> ReleaseRequest:
+        request_file = release_request.get_request_file_from_output_path(relpath)
+        return self._delete_from_then_add_file_to_request(
```
rebkwok, line 1774: I think it makes sense to re-use this code, but I don't like the name of the method. `replace_file_in_request`?  Or maybe it just needs some comments to explain why we have to delete and then re-add a file.

airlock/business_logic.py
```diff
@@ -1785,15 +1810,17 @@ def update_file_in_request(
         ), "File hash does not match manifest.json"
 
         request_file = release_request.get_request_file_from_output_path(relpath)
+        old_group = request_file.group
+        old_filetype = request_file.filetype
 
         for reviewer_username in request_file.reviews:
```
rebkwok, line 1816: When a file is withdrawn, does it keep all its existing reviews? Again, maybe this just needs some more explanation about why we have to do it

airlock/business_logic.py
```diff
@@ -1808,8 +1835,8 @@ def update_file_in_request(
             type=AuditEventType.REQUEST_FILE_WITHDRAW,
             user=user,
             path=relpath,
-            group=request_file.group,
-            filetype=request_file.filetype.name,
+            group=old_group,
+            filetype=old_filetype.name,
         )
         filegroup_data = self._dal.delete_file_from_request(
```
rebkwok, line 1841: For withdrawing, haven't we already done this? Is it going to be confusing in the audit log that we withdraw a withdrawn file for the second time? And then the next audit log is to update it, but we're actually adding it, aren't we? Can you withdraw a file and then change your mind and add it again, without it having been changed?

rebkwok (commented):
Some questions about what happens when we re-add a withdrawn file. If we do need to do all those steps for withdrawn files, then i think this is fine, maybe just some additional comments.

airlock/business_logic.py
madwort: The behaviour prior to this PR (which is from https://github.com/opensafely-core/airlock/pull/563 ) is that when a file is withdrawn it keeps all its reviews, then if it is subsequently re-added, it loses those reviews. Maybe we should be keeping those reviews?

airlock/business_logic.py
madwort: > For withdrawing, haven't we already done this?

Not in this case. 

If the request is PENDING, and you withdraw a file, it *is* deleted. But then if you try to add again the file we would run the add handler (not the add_withdrawn handler), because it doesn't exist on the request.

This is dealing with a request in RETURNED state - when you withdraw the file in that state it is marked as withdrawn, *not* deleted. Then if you add the file back again, we run the add_withdrawn handler which deletes it & then adds it.

> Is it going to be confusing in the audit log that we withdraw a withdrawn file for the second time?

Yes, although it does reflect what's happening in the code - in that there are actually two withdraw events in this scenario as described above (first marks as withdrawn, second deletes). The only way around this is to not emit an audit event at all for the second withdraw event in this scenario, which means holding the audit log a bit wrong, or pushing this code out of BLL & into DAL, which I think means holding that separation a bit wrong. 

> And then the next audit log is to update it, but we're actually adding it, aren't we?

Yes, that's true from the end user's point of view, I could change?

> Can you withdraw a file and then change your mind and add it again, without it having been changed?

Yep, that's what this code is trying to do.

airlock/business_logic.py
madwort: `replace_file_in_request` is a much better name!

airlock/business_logic.py
rebkwok: I think current behaviour is OK, but maybe a comment to say so?

airlock/business_logic.py
rebkwok: Probably OK for now - I think we had some bigger questions about what and how much we put in audit logs/activity logs

rebkwok (approved):

PR #592, opened 2024-07-31
Refactoring permissions, part 1/N

rebkwok:
This is the start of refactoring permissions into their own module.

As suggested by @inglesp, permissions are in pairs of fucntions, `user_can_do_x()` which returns a boolean and `check_user_can_do_x()` which raises exceptions. I chose to have `user_can_do_x()` call `check_user_can_do_x()` and turn the caught exception raised rather than the inverse, because that means that `check_user_can_do_x()` can raise the exception with different useful error messages, depending on what raised it.

So far this pulls out the permission checking that was in users.py and most of the permission checking from business_logic_layer.py. It doesn't yet deal with file voting or check_status, because a) this PR was getting a bit big, and b) there's some other notes about refactoring check_status to do more specific checks alongside the status changes.


airlock/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+        raise exceptions.RequestPermissionDenied(
+            f"you do not have permission to author requests for {workspace_name}"
+        )
+
+
+def user_has_role_on_workspace(user: User | None, workspace_name: str):
+    try:
+        return check_user_has_role_on_workspace(user, workspace_name) is None
+    except exceptions.RequestPermissionDenied:
+        return False
```
madwort, line 56: this is fine, although personally I would have been tempted by
```
def user_has_role_on_workspace(user: User | None, workspace_name: str):
    try:
        check_user_has_role_on_workspace(user, workspace_name)
    except exceptions.RequestPermissionDenied:
        return False
    return True
```

madwort: (I only mention this because it's the beginning of setting a pattern)

airlock/business_logic.py
```diff
@@ -1489,52 +1489,35 @@ def get_requests_authored_by_user(self, user: User) -> list[ReleaseRequest]:
...
-            for attrs in self._dal.get_requests_by_status(
+        return list(
+            self._get_reviewable_requests_by_status(
+                user,
                 RequestStatus.SUBMITTED,
                 RequestStatus.PARTIALLY_REVIEWED,
                 RequestStatus.REVIEWED,
             )
-            # Do not show output_checker their own requests
-            if attrs["author"] != user.username
```
madwort, line 1524: really good to get rid of this line

madwort (approved):

airlock/permissions.py
rebkwok: This is what I did first, then I decided I didn't like the extra `return True`. But if you'd have done it this way too, maybe my first instinct was the right one

PR #594, opened 2024-07-31
Refactor permissions.py into policies and permissions

rebkwok:
We want to extract to a common location all of our checks on whether a user can do a thing. Some of these checks are permissions (e.g. a user never has permission to review a request if they are not an output checker). However, some are dependent on the state of the request or file that the user is trying to act on. These change, and so aren't solely user- dependent permissions. E.g. a user may have permission to review a request, but can't actually review at this stage because the request is in RETURNED status, and can only be worked on by the author.

This splits permissions and policies, where policies are checks on whether a permitted user can peform an action on a request/file at this particular time/status.

### DIFF
--- a/airlock/policies.py
+++ b/airlock/policies.py
@@ -1,0 +1,70 @@
+"""
+Check if a user is allowed to perform a particular action on a request or
+request file. This is dependent on our policies regarding changes that can
+be made:
+- while requests are in particular statuses
+- for types of file
+- at a certain phase in a request turn
+
+For example, a user has permission to review a particular request and vote on
+a file if they are an output checker and they are not the author of this request.
+However, in addition, the file must be of type OUTPUT, and the request must be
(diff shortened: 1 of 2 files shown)


airlock/policies.py
```diff
@@ -0,0 +1,125 @@
...
+    editable state
+    """
+    if user.username != request.author:
+        raise exceptions.RequestPermissionDenied(
+            f"only author {request.author} can modify the files in this request"
+        )
+    if not request.is_editing():
+        raise exceptions.RequestPermissionDenied(
+            f"cannot modify files in request that is in state {request.status.name}"
+        )
```
madwort, line 46: see discussion in slack - I guess I was thinking that half of this would go in policies & half in permissions, maybe that doesn't make life any easier though.

madwort (approved):

PR #604, opened 2024-08-08
Amend policy for adding a file to a request & reuse

madwort:
* the UI would not present the option to add a file in the state `UNDER_REVIEW`. Although the BLL function would accept it, I think the DAL would correctly reject it
* I believe the existing function in `policies.py` was derived from the BLL function

### DIFF
--- a/airlock/policies.py
+++ b/airlock/policies.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from airlock import exceptions
+from airlock.enums import WorkspaceFileStatus
 from airlock.types import UrlPath
 from airlock.utils import is_valid_file_type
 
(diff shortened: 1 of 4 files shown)


airlock/policies.py
```diff
@@ -53,6 +69,12 @@ def check_can_add_file_to_request(workspace: "Workspace", relpath: UrlPath):
     if workspace.file_has_been_released(relpath):
         raise exceptions.RequestPermissionDenied("Cannot add released file to request")
 
+    workspace_status = workspace.get_workspace_file_status(relpath)
+    if workspace_status not in VALID_WORKSPACEFILE_STATES_TO_ADD:
+        raise exceptions.RequestPermissionDenied(
+            f"Cannot add file in status {workspace_status}"
+        )
```
madwort, line 76: this block transferred as-is from its previous location, I'm doing more work on this in my other branch as I tease apart add_file from update_file, so I don't think it's worth grooming this right now.

madwort:
changed my mind about this a little, moving this to the workspace class

PR #605, opened 2024-08-08
Show a sensible dialogue when updating a file

madwort:
* before this PR, when updating a file we were showing the add file dialogue
  * this included filetype & group selectors which were being ignored 
* create (& use) a new multiselect view & form to use for updating files
* refactor the code for re-adding a withdrawn file & updating a file using the new policies/permissions files
* at the moment updating a file is limited to one file, but this lays a lot of the groundwork to implement a multiselect update files button on the workspace view

Screenshot 2024-08-09 at 16 38 50

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -458,20 +458,6 @@ class Workspace:
     def request_filetype(self, relpath: UrlPath) -> None:
         return None
 
-    def file_has_been_released(self, relpath: UrlPath) -> bool:
-        return self.get_workspace_file_status(relpath) == WorkspaceFileStatus.RELEASED
-
-    def file_can_be_added(self, relpath: UrlPath) -> bool:
-        return self.get_workspace_file_status(relpath) in [
-            WorkspaceFileStatus.UNRELEASED,
-            WorkspaceFileStatus.WITHDRAWN,
-        ]
-
-    def file_can_be_updated(self, relpath: UrlPath) -> bool:
-        return self.get_workspace_file_status(relpath) in [
-            WorkspaceFileStatus.CONTENT_UPDATED,
-        ]
-
 
 @dataclass(frozen=True)
 class CodeRepo:
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -461,17 +461,15 @@ class Workspace:
     def file_has_been_released(self, relpath: UrlPath) -> bool:
         return self.get_workspace_file_status(relpath) == WorkspaceFileStatus.RELEASED
 
-    def file_can_be_updated(self, relpath: UrlPath) -> bool:
-        return self.get_workspace_file_status(relpath) in [
-            WorkspaceFileStatus.CONTENT_UPDATED,
-            WorkspaceFileStatus.WITHDRAWN,
-        ]
-
     def file_can_be_added(self, relpath: UrlPath) -> bool:
         return self.get_workspace_file_status(relpath) in [
             WorkspaceFileStatus.UNRELEASED,
-            WorkspaceFileStatus.CONTENT_UPDATED,
             WorkspaceFileStatus.WITHDRAWN,
+        ]
+
+    def file_can_be_updated(self, relpath: UrlPath) -> bool:
+        return self.get_workspace_file_status(relpath) in [
+            WorkspaceFileStatus.CONTENT_UPDATED,
         ]
 
 
@@ -1640,6 +1638,12 @@ class BusinessLogicLayer:
         relpath: UrlPath,
         user: User,
     ) -> ReleaseRequest:
+        relpath = UrlPath(relpath)
+        workspace = self.get_workspace(release_request.workspace, user)
+        permissions.check_user_can_update_file_on_request(
+            user, release_request, workspace, relpath
+        )
+
         request_file = release_request.get_request_file_from_output_path(relpath)
         return self.replace_file_in_request(
             release_request, relpath, user, request_file.group, request_file.filetype
@@ -1653,6 +1657,12 @@ class BusinessLogicLayer:
         group_name: str = "default",
         filetype: RequestFileType = RequestFileType.OUTPUT,
     ) -> ReleaseRequest:
+        relpath = UrlPath(relpath)
+        workspace = self.get_workspace(release_request.workspace, user)
+        permissions.check_user_can_add_file_to_request(
+            user, release_request, workspace, relpath
+        )
+
         return self.replace_file_in_request(
             release_request, relpath, user, group_name, filetype
         )
@@ -1667,7 +1677,7 @@ class BusinessLogicLayer:
     ) -> ReleaseRequest:
         relpath = UrlPath(relpath)
         workspace = self.get_workspace(release_request.workspace, user)
-        permissions.check_user_can_update_file_on_request(
+        permissions.check_user_can_replace_file_in_request(
             user, release_request, workspace, relpath
         )
 
--- a/airlock/forms.py
+++ b/airlock/forms.py
@@ -81,9 +81,6 @@ class FileForm(forms.Form):
     )
 
 
-FileFormSet = formset_factory(FileForm, extra=0)
-
-
 class FileTypeForm(forms.Form):
     FILETYPE_CHOICES = [
         (RequestFileType.OUTPUT.name, RequestFileType.OUTPUT.name.title()),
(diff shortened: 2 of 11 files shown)


airlock/views/workspace.py
```diff
@@ -252,25 +251,51 @@
...
+    )
+
+    filetype_formset = FileFormSet(
+        initial=[{"file": f} for f in files_to_add],
+    )
 
-    # default redirect in case of error
-    next_url = workspace.get_url()
+    return TemplateResponse(
+        request,
```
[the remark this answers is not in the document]
madwort: I think we've ignored this warning in the past?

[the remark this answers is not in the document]
evansd: Yeah, this whole code scanning thing is Not Very Useful.

airlock/templates/update_files.html
```diff
@@ -0,0 +1,41 @@
+{# hide default button, as we are going to autoshow the modal anyway #}
+{% fragment as button %}
+{% endfragment %}
+<style>
+  div.filetype-radio div { display: hidden; }
+  div.filetype-radio label { display: hidden; }
+</style>
```
rebkwok, line 7: I think everything above here can be removed; we don't need to hide the button, or set `custom_button=button` in the line below. We don't define any custom `button_text`, so the default modal button is hidden anyway.
And the formset that's used in this template doesn't have a filetype-radio, so there's no need to hide it.

airlock/views/workspace.py
```diff
@@ -245,6 +250,45 @@ def multiselect_add_files(request, multiform, workspace):
...
+    files_ignored = {}
+
+    # validate which files can be added
+    for f in multiform.cleaned_data["selected"]:
+        workspace.abspath(f)  # validate path
+
+        if policies.can_update_file_on_request(workspace, UrlPath(f)):
+            files_to_add.append(f)
+        else:
+            files_ignored[f] = "cannot add using the update dialogue"
```
rebkwok, line 264: Should this check `can_add_file_to_request`?  If the file is already released, it won't be updatable, and this message implies it can be added, just not with this dialogue.

airlock/business_logic.py
```diff
@@ -461,17 +461,15 @@ def request_filetype(self, relpath: UrlPath) -> None:
     def file_has_been_released(self, relpath: UrlPath) -> bool:
         return self.get_workspace_file_status(relpath) == WorkspaceFileStatus.RELEASED
 
-    def file_can_be_updated(self, relpath: UrlPath) -> bool:
+    def file_can_be_added(self, relpath: UrlPath) -> bool:
```
rebkwok, line 464: We should be able to get rid of these methods on `Workspace` now that the enums are in their own module, and move the logic here into `policies.py` (they were only here before to avoid circular imports when importing `WorkspaceFileStatus` in `policies.py`)

airlock/permissions.py
```diff
@@ -151,6 +151,23 @@ def user_can_add_file_to_request(
...
+):
+    check_user_can_edit_request(user, request)
+    policies.check_can_replace_file_in_request(workspace, relpath)
+
+
+def user_can_replace_file_in_request(
+    user: User, request: "ReleaseRequest", workspace: "Workspace", relpath: UrlPath
+):  # pragma: no cover; not currently used
+    try:
+        check_user_can_update_file_on_request(user, request, workspace, relpath)
```
rebkwok, line 165: should be `check_user_can_replace_file_on_request`?

airlock/policies.py
```diff
@@ -68,6 +68,37 @@ def check_can_add_file_to_request(workspace: "Workspace", relpath: UrlPath):
...
+    try:
+        check_can_replace_file_in_request(workspace, relpath)
+    except exceptions.RequestPermissionDenied:
+        return False
+    return True
+
+
+def check_can_replace_file_in_request(workspace: "Workspace", relpath: UrlPath):
+    """
+    This file can replace an existing file in the request.
```
rebkwok, line 81: Can you expand this comment to explain how it's different to `can_update...`; it wasn't immediately obvious to me why we need to check `can_be_added` or `can_be_updated` when replacing a file (it's the withdrawn ones, right? They're still technically on the request and need to be replaced if they're being re-added)

airlock/business_logic.py
madwort: hah! doh! I did that originally, and then changed my mind & followed what you'd done elsewhere & moved them onto Workspace! I'll change.

airlock/templates/update_files.html
madwort: absolutely right, removed.

airlock/views/workspace.py
```diff
@@ -244,25 +244,14 @@ def multiselect_add_files(request, multiform, workspace):
...
-def workspace_add_file_to_request(request, workspace_name):
-    workspace = get_workspace_or_raise(request.user, workspace_name)
-    release_request = bll.get_or_create_current_request(workspace_name, request.user)
-    form = AddFileForm(request.POST, release_request=release_request)
-    formset = FileTypeFormSet(request.POST)
-
-    # default redirect in case of error
-    next_url = workspace.get_url()
+# also displays errors if present
+def add_or_update_forms_are_valid(request, form, formset):
```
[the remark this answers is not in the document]
madwort: ah, right, yeah, I had the collective noun for a form and a formset containing forms as "forms", but I see what you mean, it is actually be "form". changed!

airlock/views/workspace.py
madwort: I've done what I suggested above

airlock/templates/update_files.html
madwort: it seems we do need `custom_button=button` & some other lines, otherwise we get an artefact on the parent page, as per this screenshot:

Screenshot 2024-08-14 at 11 44 56

airlock/permissions.py
madwort: yikes, yes, thanks for spotting!

airlock/forms.py
```diff
@@ -74,6 +74,16 @@ def clean_new_filegroup(self):
...
 
 
+class FileForm(forms.Form):
+    file = forms.CharField(
+        required=True,
+        widget=forms.HiddenInput(),
+    )
+
+
+FileFormSet = formset_factory(FileForm, extra=0)
```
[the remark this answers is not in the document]
madwort: yep, I've re-used `RequiredOneBaseFormSet`, although I don't understand this bit

> then we don't need an additional formset.

are you suggesting that I should refactor `FileTypeForm` to inherit from `FileForm` in order for them to share a Formset somehow?

airlock/policies.py
madwort: added a clarifying comment, does that make sense now?

airlock/business_logic.py
madwort: refactored all of these into the policies file

airlock/policies.py
```diff
@@ -79,7 +130,11 @@ def check_can_update_file_on_request(workspace: "Workspace", relpath: UrlPath):
             f"Cannot update file of type {relpath.suffix} in request"
         )
 
-    if not workspace.file_can_be_updated(relpath):
+    status = workspace.get_workspace_file_status(relpath)
+
+    if status not in [
+        WorkspaceFileStatus.CONTENT_UPDATED,
+    ]:
```
rebkwok, line 137: change to `status !=  WorkspaceFileStatus.CONTENT_UPDATED`?

airlock/templates/update_files.html
```diff
@@ -0,0 +1,37 @@
+{# hide default button, as we are going to autoshow the modal anyway #}
+{% fragment as button %}
+{% endfragment %}
```
rebkwok, line 3: Pretty sure that the above lines can be deleted, and the line below can change to 
```
{% #modal id="updateRequestFile" autoshow=True %}
```

rebkwok (approved):
couple of minor nits but otherwise :+1:

airlock/templates/update_files.html
madwort: I think this is because it's still sharing some template stuff with the add_files dialogue - maybe I should do a further refactor of that. I'll probably need to touch that anyway when re-enabling updating multiple files.

madwort: I'm going to merge & do this in my next PR, hope that's ok.

PR #606, opened 2024-08-09
Small add file permission refactor

madwort:
* the UI would not present the option to add a file in the state UNDER_REVIEW. Although the BLL function would accept a file in that state, the BLL's call to the DAL would then correctly reject it
* I believe the existing function in policies.py was derived from the BLL function

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -467,6 +467,13 @@ class Workspace:
             WorkspaceFileStatus.WITHDRAWN,
         ]
 
+    def file_can_be_added(self, relpath: UrlPath) -> bool:
+        return self.get_workspace_file_status(relpath) in [
+            WorkspaceFileStatus.UNRELEASED,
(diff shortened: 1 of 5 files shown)


inglesp (approved):
Why do `file_can_be_updated` and `file_can_be_added` belong on `Workspace` and not in `policies.py`?

madwort:
> Why do `file_can_be_updated` and `file_can_be_added` belong on `Workspace` and not in `policies.py`?

That's a great question & it's a real judgement call - my first version of this PR had the function body in policies & then I changed my mind this morning & moved it into `Workspace` this morning. 
Part of my reasoning is "because it's what Becky did" in https://github.com/opensafely-core/airlock/pull/592/commits/af9135826e5596263925e4327d6a1eff36adbd75 , and part of my reasoning is that the details of the `WorkspaceFileStatus` enum do belong alongside the rest of the code that works with it in `Workspace` - even though we're using policies.py as a global router for code that needs to reason about files in workspaces.

PR #607, opened 2024-08-09
Allow withdrawing files from the file list view

### LINKED ISSUE #247: Withdraw files from the list view
https://bennettoxford.slack.com/archives/C065F8P58AW/p1712755051357289?thread_ts=1712753890.166379&cid=C065F8P58AW

> To drop a file from the request you have to do it from the individual #file view. Whilst i'm preparing a request, I'd like to be able to drop them from the list view.

evansd:
Closes #247 

UI preview:
image


airlock/views/request.py
```diff
@@ -475,6 +483,39 @@ def file_withdraw(request, request_id, path: str):
...
+            try:
+                bll.withdraw_file_from_request(release_request, path, request.user)
+                successes.append(f"The file {path} has been withdrawn from the request")
+            except exceptions.RequestPermissionDenied as exc:
+                errors.append(str(exc))
+
+        if errors:
+            display_multiple_messages(request, errors, "error")
+        if successes:
+            display_multiple_messages(request, successes, "success")
```
madwort, line 513: for interest, I was just working near [the equivalent place in workspace.py](https://github.com/opensafely-core/airlock/blob/main/airlock/views/workspace.py#L330-L332), which has a slightly different behaviour - I think I prefer this though.

airlock/templates/file_browser/_includes/directory_table/request.html
```diff
@@ -24,6 +29,13 @@
         <td>{{ path.request_filetype.name|title }}</td>
         <td data-order="{{ path.size }}">{{ path.size_mb }}</td>
         <td>{{ path.modified_at|date:"Y-m-d H:i" }}</td>
+        {% if multiselect_withdraw %}
+          <td>
+            {% if True %}
```
madwort, line 34: why `if True`?

madwort (approved):

airlock/views/request.py
evansd: Ah yeah, I hadn't seen that. I think it's better to keep them separate though.

airlock/templates/file_browser/_includes/directory_table/request.html
evansd: :facepalm: Leftover debugging! Thanks for catching

airlock/views/request.py
madwort: once this has merged I think I'll change workspace to match

PR #608, opened 2024-08-09
Always show workspaces msgs with the correct level

madwort:
* following the way it's now done in request.py
* also bump the empty array test into the helper function
* remove the empty test from request.py

### DIFF
--- a/airlock/views/helpers.py
+++ b/airlock/views/helpers.py
@@ -77,6 +77,8 @@ def get_path_item_from_tree_or_404(tree: PathItem, path: UrlPath | str):
 
 
 def display_multiple_messages(request, msgs, level="success"):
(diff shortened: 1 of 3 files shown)


evansd (approved):
Moving the conditional inside the function is a nice tidy-up :+1:

airlock/views/helpers.py
```diff
@@ -77,6 +77,8 @@ def get_path_item_from_tree_or_404(tree: PathItem, path: UrlPath | str):
 
 
 def display_multiple_messages(request, msgs, level="success"):
+    if not msgs:
+        return
```
madwort, line 81: nb. I'm not wedded to this, lmk if you'd prefer to remove

PR #616, opened 2024-08-12
Only show request actions on request overview page

### LINKED ISSUE #472: Only show submit & withdraw on the top-level request view
requires a "back to top-level request view" button in its place

evansd:
This moves the various request action buttons out of the header and onto the top-level request overview page. The area in the header previously occupied by the buttons is replaced by a link to the request overview.

This results in some more hoping between pages during the end-to-end tests, but nothing too complicated.

image

Closes #472


rebkwok (approved):
"Back to request" feels a bit odd in the context, since you're already in the request. Especially under "back to workspace" which takes you to a different bit of airlock entirely. Not sure I can think of a very good alternative though. "Go to request root/home"?

PR #617, opened 2024-08-13
Enforce consistent requirements between dev and prod

evansd:


bloodearnest (approved):
it maybe uv may help here in future, I hope


How about also adding a justfile command `compile` that always does both, in the right order? Document that rather than `pip-compile` (and hide that from the help list), and it will likely make the scenario even less likely?

evansd:
> How about also adding a justfile command compile that always does both, in the right order?

Yes! We discussed exactly this in the thread with Becky where the issue originally came up. I just wanted to quickly make it impossible to do the wrong thing first. But it would be good to make it easy to do the right thing also.

PR #619, opened 2024-08-13
Remove factories request creation workaround

bloodearnest:
- **Remove ability to customise the id of a request**
- **Remove bll._create_release_request**
- **Separate factories.create_user use cases**

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1308,14 +1308,11 @@ class BusinessLogicLayer:
         # and is not an output-cheker
         request = self.get_current_request(workspace, user)
 
-        # requests for output-checkers, and for archived workspaces and inactive
-        # projects are still viewable, check if user has permission to create one
-        # Note: we check this here, as this function is only used when we are
-        # about to modify a request, so do a look ahead check to make sure we can
-        permissions.check_user_can_action_request_for_workspace(user, workspace)
-
         if request is not None:
             return request
+
+        # check if user has permission to create one
+        permissions.check_user_can_action_request_for_workspace(user, workspace)
 
         audit = AuditEvent(
             type=AuditEventType.REQUEST_CREATE,
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -1077,7 +1077,6 @@ class DataAccessLayerProtocol(Protocol):
         author: str,
         status: RequestStatus,
         audit: AuditEvent,
-        id: str | None = None,  # noqa: A002
     ):
         raise NotImplementedError()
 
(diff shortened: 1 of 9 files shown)


airlock/business_logic.py
```diff
@@ -1337,11 +1310,28 @@ def get_or_create_current_request(
 
         # requests for output-checkers, and for archived workspaces and inactive
         # projects are still viewable, check if user has permission to create one
+        # Note: we check this here, as this function is only used when we are
+        # about to modify a request, so do a look ahead check to make sure we can
         permissions.check_user_can_action_request_for_workspace(user, workspace)
```
inglesp, line 1315: I see there's been a bit of discussion in Slack about this function, and I share the confusion about why we'd check for these permissions when _getting_ a request.  The comment helps, but what happens if we move the check to after the `return request` line?

tests/factories.py
```diff
@@ -30,32 +30,42 @@
...
+        for workspace in workspaces
+    }
+
+    return create_user_from_dict(
+        username, workspaces_dict, output_checker, last_refresh
+    )
+
 
+def create_user_from_dict(
+    username, workspaces_dict, output_checker=False, last_refresh=None
```
inglesp, line 64: Should these parameters be typed?

inglesp (approved):
:+1:

airlock/business_logic.py
bloodearnest: If we do that, the test mentioned in the discussion fails. I discovered this by doing exactly as you suggest.

I'm not sure if this behaviour is load bearing on not. I can investigate further?

tests/factories.py
bloodearnest: So, we don't type check tests/ so types annotations are not guaranteed to be 100% accurate, so I've tended not to. I added the list[str] one as documenation to clarify the current confusion.

And the type for workspace_dict is messy, something like `dict[str, bool | dict[str, str | bool]]`

Also, I think this dict mess will go away with proper user model, so I'm inclined to leave it

madwort: I would like to see this done, but as a separate ticket that enables mypy checking for the entire `tests/` directory & all current typing issues fixed...

airlock/business_logic.py
inglesp: The failing tests are testing `get_or_create_current_request` directly.  `get_or_create_current_request` is only used by `workspace_add_file_to_request`, and tests for `workspace_add_file_to_request` continue to pass if we move the permissions check.

Additionally, we've got extra permission checks later on in the BLL functions called by `workspace_add_file_to_request`.  As such, the permissions check that protects _getting_ an existing request is not load bearing.

So I'm pretty confident we can move the permissions check, and remove the failing tests.  I think the failing tests are too closely coupled to our current implementation, and don't test anything about the actual behaviour of the system.

bloodearnest: So, there's another bll functions that are "protected" by this: `add_withdrawn_file_to_request`, but that seems fine also, so I agree.

bloodearnest: done

tests/factories.py
bloodearnest: Might be worth trying. I suspect that it might be quite a bit of work, but may be useful

Resolving this comment for now

airlock/business_logic.py
inglesp: great

PR #620, opened 2024-08-14
Revert "Hide multiselect for updated files (#595)"

madwort:
This reverts commit 91e1cb4858d19629b1e67b7c1d9ec3da13ce5b42.

### DIFF
--- a/airlock/file_browser_api.py
+++ b/airlock/file_browser_api.py
@@ -190,9 +190,6 @@ class PathItem:
     def is_valid(self):
         return is_valid_file_type(Path(self.relpath))
 
-    def is_updated(self):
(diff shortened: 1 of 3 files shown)


rebkwok (approved):
Looks good, except I don't like the "Cannot be updated using the add dialogue" message when you try to add an updated file. I'm not sure a user would understand what we mean. Maybe something like "File already added to request, select Update Files to update it" (but probably that's too wordy).

madwort:
fair point - I've removed that message, so now it will show the default "already in group x" message if you try to add an updated file. It's slightly less helpful, but I think people should be able to work it out, and it looks much tidier.

PR #621, opened 2024-08-14
Add group context modal to file view

### LINKED ISSUE #483: request view - file page - "go to group" button
when viewing a file, it's difficult to know where to put a comment for that file

bloodearnest:
- **Refactor request group template context**
- **Add view context modal to contents view**
- **Add small functional test for context modal**
Button here: image

Modal: 
image

Fixes #483

### DIFF
--- a/airlock/templates/file_browser/group.html
+++ b/airlock/templates/file_browser/group.html
@@ -8,26 +8,26 @@
   }
 </style>
 {% #card title=path_item.name container=True %}
-  <form action="{{ group_edit_url }}" method="POST" aria-label="group-edit-form">
+  <form action="{{ group.c2_edit_url }}" method="POST" aria-label="group-edit-form">
     {% csrf_token %}
     <div style="display: flex; flex-direction: row; width: 100%; flex-wrap: wrap; gap: 1rem;">
       <div style="display: flex: flex-direction: column; flex-basis: 100%; flex: 1;">
-        {% form_textarea field=group_edit_form.context show_placeholder=True resize=True placeholder="Describe the data to be released in this group of files" class="w-full max-w-lg mx-auto" rows=6 disabled=group_readonly readonly=group_readonly %}
+        {% form_textarea field=group.c2_edit_form.context show_placeholder=True resize=True placeholder="Describe the data to be released in this group of files" class="w-full max-w-lg mx-auto" rows=6 disabled=group.c2_readonly readonly=group.c2_readonly %}
       </div>
       <div style="display: flex: flex-direction: column; flex-basis: 100%; flex: 1; gap: 1rem;">
-        {% form_textarea field=group_edit_form.controls resize=True show_placeholder=True placeholder="Describe the disclosure controls that have been applied to these files" class="w-full max-w-lg mx-auto" rows=6 disabled=group_readonly readonly=group_readonly %}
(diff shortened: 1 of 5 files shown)


airlock/views/request.py
```diff
@@ -287,13 +285,23 @@ def group_presenter(release_request, relpath, request):
     visibilities = release_request.get_writable_comment_visibilities_for_user(
         request.user
     )
+    # are we on the group page?
+    if len(relpath.parts) == 1:
+        inline = False
+    else:
+        inline = True
```
madwort, line 292: it's semi-tempting to do a refactor to add some convenience methods to `UrlPath` for things like this, although from a quick skim of the codebase we've not hit the 3-4 repeated usages I'd want to see to make me think it's worthwhile. (I'm thinking `relpath.is_group()` for here and also an accessor-type method `relpath.group` in place if `relpath.parts[0]`.

airlock/views/request.py
```diff
@@ -323,12 +270,68 @@ def request_view(request, request_id: str, path: str = ""):
...
+
+    # are comments readonly?
+    c2_readonly = inline or not permissions.user_can_edit_request(
+        request.user, release_request
+    )
```
madwort, line 297: looks like c2 is context & controls, not comments?

```suggestion
    # are context and controls readonly?
    c2_readonly = inline or not permissions.user_can_edit_request(
        request.user, release_request
    )
```

madwort: (this is why I like extra-long variable names...)

airlock/templates/file_browser/group.html
```diff
@@ -7,7 +7,7 @@
     background-color: darkgrey;
   }
 </style>
-{% #card title=path_item.name container=True %}
+{% #card title=group.title|default_if_none:path_item.name container=True class="group_modal" %}
```
madwort, line 10: in what context could `group.title` be none?

madwort (approved):

airlock/views/request.py
bloodearnest: Well, that's a whole thing

There are times when a UrlPath means "path to file in a workspace" and times when it means "actual url" and times where it means "request url that will include group". cf `bll.get_request_file_from_output_path and bll.get_request_file_from_urlpath`.

Previously, I tried to split those into separate types: OutputPath, UrlPath, RequestUrlPath (which had a `.group` as you suggest and a `.output_path` w/o the group), and use the right types to help us catch these distintions.

But at the time, I couldn't make it work cleanly.  It's probably worth trying again, and definitely would be a good refactor.

airlock/templates/file_browser/group.html
bloodearnest: ah, it was possibly none at an earlier point in the branch, when I added this. Will remove.

bloodearnest: fixed

airlock/views/request.py
madwort: > Curious - you would seriously prefer context_and_controls_readonly here?

yes I would! I know I'm an outlier...

airlock/views/request.py
madwort: yeah, I think it would be a really valuable refactor, if we have time.

madwort: put a ticket on the triage pile https://github.com/opensafely-core/airlock/issues/622

PR #623, opened 2024-08-15
Madwort/fix some mypy tests

madwort:


inglesp (approved):
Approving, but am a bit ambivalent about this.  It seems to add a fair bit of noise (especially the type narrowing) and unless we enforce type checking, I don't know how much benefit we get.

How close are we to being able to enforce type checking across the whole tests package?

Issue #622, opened 2024-08-15
Refactor UrlPath


madwort:
> There are times when a UrlPath means "path to file in a workspace" and times when it means "actual url" and times where it means "request url that will include group". cf bll.get_request_file_from_output_path and bll.get_request_file_from_urlpath.
> 
> Previously, I tried to split those into separate types: OutputPath, UrlPath, RequestUrlPath (which had a .group as you suggest and a .output_path w/o the group), and use the right types to help us catch these distintions.
> 
> But at the time, I couldn't make it work cleanly. It's probably worth trying again, and definitely would be a good refactor.
> 

https://github.com/opensafely-core/airlock/pull/621#discussion_r1718240397

madwort:
I think I've fought this a bit in the past (having to keep in mind whether a path contains a group name or not) - would be great to have it really obvious (& enforced by the type system).

bloodearnest:
One idea I wanted to try here was have the types enforced/converted in our django urls.py patterns, so that the views got typed instances of UrlPath/GroupPath, rather than strings.

You can do this with custom url type converters easily enough: https://docs.djangoproject.com/en/5.0/topics/http/urls/#registering-custom-path-converters

madwort:
I'm not sure we can do it with custom path converters if we need access to other data, e.g. `request.user`, `release_request` etc - we could do it with middleware though?

bloodearnest:
I don't think we need anything except path components for a converter to UrlPath/GroupPath?

madwort:
actually you're right - I thought this was going to verify filenames/groupnames against a request but for these purposes it doesn't. this has given me the idea to do a similar thing to other params, though, which do need these things - e.g. release_request (see slack thread https://bennettoxford.slack.com/archives/C069YDR4NCA/p1741949840761249 )

rw251:
When I was working on allowing groups to be renamed, I did [some preliminary work](https://github.com/opensafely-core/airlock/compare/main...rw/separate-group-from-path) that separated the `group_name` field from the `path` in the URLs (the next step would have been `group_name` -> `group_id`. This may or may not be useful depending on your approach, but linking it here so at least you know about it.

PR #625, opened 2024-08-15
Enable mypy for tests/

madwort:
* fix or ignore all remaining mypy issues with `tests/`
* add `tests/` to the list of directories checked as standard by mypy
* add some type annotations to `test/factories.py` and fix mypy issues as they arise

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -590,7 +590,7 @@ def create_filegroup(release_request, group_name, filepaths=None):
     user = create_user(release_request.author, [release_request.workspace])
     for filepath in filepaths or []:  # pragma: nocover
         bll.add_file_to_request(release_request, filepath, user, group_name)
-    return bll._dal._get_or_create_filegroupmetadata(release_request.id, group_name)
+    return bll._dal._get_or_create_filegroupmetadata(release_request.id, group_name)  # type: ignore
 
 
 def refresh_release_request(release_request, user=None) -> ReleaseRequest:
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -30,11 +30,11 @@ from airlock.users import User
 
 
 def create_user(
-    username="testuser",
+    username: str="testuser",
     workspaces: list[str] | None = None,
-    output_checker=False,
+    output_checker: bool=False,
     last_refresh=None,
-):
+) -> User:
     """Factory to create a user.
 
     For ease of use, workspaces is just a flat list of workspace name, which is
@@ -62,7 +62,7 @@ def create_user(
 
 def create_user_from_dict(
     username, workspaces_dict, output_checker=False, last_refresh=None
-):
+) -> User:
     if last_refresh is None:
         last_refresh = time.time()
 
@@ -262,7 +262,9 @@ def create_repo(workspace, files=None, temporary=True):
     return CodeRepo.from_workspace(workspace, commit)
 
 
-def create_release_request(workspace, user=None, status=None, **kwargs) -> ReleaseRequest:
+def create_release_request(
+    workspace, user=None, status=None, **kwargs
+) -> ReleaseRequest:
     if status:
         assert (
             status == RequestStatus.PENDING
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -484,7 +484,9 @@ def get_default_output_checkers():
     ]
 
 
-def review_file(request: ReleaseRequest, relpath: UrlPath, status, *users):
+def review_file(
+    request: ReleaseRequest, relpath: UrlPath, status: RequestFileVote, *users
+):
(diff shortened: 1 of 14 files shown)


tests/unit/test_business_logic.py
```diff
@@ -501,7 +501,9 @@ def test_provider_get_workspaces_for_user(bll, output_checker):
             "archived": False,
         },
     }
-    user = factories.create_user(workspaces=workspaces, output_checker=output_checker)
+    user = factories.create_user_from_dict(
+        username="testuser", workspaces_dict=workspaces, output_checker=output_checker
+    )
```
madwort, line 506: this one is actually kind of a bug

inglesp (approved):
All looks fine (I've not reviewed with a fine-toothed comb) but again, am a bit ambivalent about this.  It seems to add a lot of noise for limited benefit.

Would be interested in what the rest of the team think.

tests/local_db/test_data_access.py
```diff
@@ -218,7 +218,9 @@ def test_group_comment_delete_bad_params():
         comment="author comment",
     )
     with pytest.raises(exceptions.APIException):
-        dal.group_comment_delete(release_request.id, "badgroup", "1", author, audit)
+        dal.group_comment_delete(
+            release_request.id, "badgroup", "1", author.username, audit
+        )
```
madwort, line 223: this DAL method should be called with a string, but was being called with a BLL User object

tests/local_db/test_data_access.py
```diff
@@ -228,7 +230,9 @@ def test_group_comment_delete_bad_params():
         comment="other comment",
     )
     with pytest.raises(models.FileGroupComment.DoesNotExist):
-        dal.group_comment_delete(release_request.id, "group", "50", author, audit)
+        dal.group_comment_delete(
+            release_request.id, "group", "50", author.username, audit
```
madwort, line 234: (same)

tests/local_db/test_data_access.py
```diff
@@ -238,4 +242,6 @@ def test_group_comment_delete_bad_params():
         comment="author comment",
     )
     with pytest.raises(exceptions.APIException):
-        dal.group_comment_delete(release_request.id, "group", "1", other, audit)
+        dal.group_comment_delete(
+            release_request.id, "group", "1", other.username, audit
```
madwort, line 246: (same)

tests/factories.py
```diff
@@ -69,7 +69,7 @@ def create_user_from_dict(
     return User(username, workspaces_dict, output_checker, last_refresh)
 
 
-def ensure_workspace(workspace_or_name) -> Workspace:
+def ensure_workspace(workspace_or_name: Workspace | str) -> Workspace:
```
madwort, line 72: Workspace or string -> `workspace_or_name`

tests/factories.py
```diff
@@ -175,7 +175,7 @@ def update_manifest(workspace: Workspace | str, files=None):
         workspace.manifest = manifest
 
 
-def create_workspace(name, user=None) -> Workspace:
+def create_workspace(name: str, user=None) -> Workspace:
```
madwort, line 178: string -> `name`

tests/factories.py
```diff
@@ -188,7 +188,7 @@ def create_workspace(name, user=None) -> Workspace:
     return bll.get_workspace(name, user)
 
 
-def write_workspace_file(workspace, path, contents="", manifest=True):
+def write_workspace_file(workspace: Workspace | str, path, contents="", manifest=True):
```
madwort, line 191: Workspace or string -> `workspace`

tests/factories.py
```diff
@@ -606,7 +606,7 @@ def refresh_release_request(release_request, user=None) -> ReleaseRequest:
 def create_audit_event(
     type_,
     user="user",
-    workspace="workspace",
+    workspace: str = "workspace",
```
madwort, line 609: string -> 'workspace'

tests/integration/views/test_workspace.py
```diff
@@ -544,7 +544,7 @@ def test_workspace_multiselect_add_files_all_valid(airlock_client, bll):
                 "test/path1.txt",
                 "test/path2.txt",
             ],
-            "next_url": workspace.get_url("test/path.txt"),
+            "next_url": workspace.get_url(UrlPath("test/path.txt")),
```
madwort, line 547: adding UrlPath to all of these paths does add a lot of noise, but I think it would make it easier to do [this refactor](https://github.com/opensafely-core/airlock/issues/622) in future.

I also feel that - `Workspace.get_url()` asks for a type, the tests should provide that type (rather than a string that happens to work).

PR #628, opened 2024-08-16
Create button to make private comments public

### LINKED ISSUE #514: Support changing visibility of existing comments (requested by Will)
possibly add an option to change the visibility of an existing comment - so if the output checkers have previously left private comments about a request, they could just make them public rather than having to write new comments. 

(they can just copy-paste into a new comment though, as a workaround)

You should only be able to change the visibility of comments in the current round.

madwort:
fixes #514 

Screenshot 2024-08-16 at 16 52 26


airlock/views/request.py
```diff
@@ -723,3 +727,38 @@ def group_comment_delete(request, request_id, group):
...
+                user=request.user,
+            )
+        except exceptions.RequestPermissionDenied as exc:  # pragma: nocover
+            # currently, we can't hit this because of get_release_request_or_raise above.
+            # However, that may change, so handle it anyway.
+            raise PermissionDenied(str(exc))
+        except exceptions.FileNotFound:
+            raise Http404(
+                request, f"Comment not found in group {group} with id {comment_id}"
+            )
```
inglesp, line 755: Under what circumstances could this be raised?  As far as I can tell this would only happen if the comment had been deleted.  In which case I think the error message should be something like "Could not make comment public as it has been deleted".

inglesp (approved):

PR #631, opened 2024-08-22
update jobserver and cover in ci

bloodearnest:
- **add new required env var to local job-server configuration**
- **use ubuntu-22.04 in CI, not latest, as that moves under us**
- **Add scheduled CI task to validate local job-server setup**

### DIFF
--- a/.github/workflows/job-server.yml
+++ b/.github/workflows/job-server.yml
@@ -1,0 +1,38 @@
+---
+name: Local job-server setup CI
+
+on:
+  schedule: 
+    # every day at 2:37am
+    - cron: "37 2 * * *"
+
+jobs:
+  test-local-job-server:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: "opensafely-core/setup-action@v1"
+        with:
+          python-version: "3.11"
+          install-just: true
+ 
+      - name: "Test local job server setup"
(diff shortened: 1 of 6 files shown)


bloodearnest:
Testing the action on `push:` trigger, it works, have removed `push:` trigger now.

https://github.com/opensafely-core/airlock/actions/runs/10507260217/job/29108631503?pr=631

rebkwok (approved):
We're going to need more job-server changes when https://github.com/opensafely-core/job-server/pull/4552 lands, aren't we?  I guess that's a good test of the workflow :)

.github/workflows/job-server.yml
```diff
@@ -0,0 +1,38 @@
...
+          just job-server/create-workspace test-workspace
+
+      - name: "Notify Slack on Failure"
+        # TODO: 2024-08-27 this does not work, gives the error "not_in_channel", despite all our bots being in the channel
+        if: failure() && github.ref_name == 'main'
+        uses: zuplo/github-action-slack-notify-build@cf8e7e66a21d76a8125ea9648979c30920195552 # v2
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+        with:
+          channel_id: "C069YDR4NCA"
```
rebkwok, line 35: Have you tried this without the quotes?  The other workflows we have that use this don't wrap the channel id in quotes

PR #632, opened 2024-08-22
Remove the SHOW_C3 flags

rebkwok:
We don't need the option to hide context, controls or comments any more.


bloodearnest (approved):
Huh, pleasingly small diff

PR #633, opened 2024-08-22
Delete comment visibility

rebkwok:
- change button type for Delete Comment button from cancel to submit (fixes bug closing Context modal when there's a delete comment button as well as a Cancel button)
- only show Delete comment button in the template if the user_can_comment flag is also present. 
- Fix the "blinded" pill to remove duplicate text and only show it for comments in the current turn

### DIFF
--- a/airlock/templates/file_browser/group.html
+++ b/airlock/templates/file_browser/group.html
@@ -64,13 +64,15 @@
                 </form>
               </div>
             {% endif %}
-            <div>
-              <form action="{{ group.comment_delete_url }}" method="POST">
-                {% csrf_token %}
-                <input type="hidden" name="comment_id" value="{{ comment.id }}">
-                {% #button variant="danger" type="submit" %}Delete comment{% /button %}
-              </form>
-            </div>
+            {% if group.user_can_comment %}
+              <div>
+                <form action="{{ group.comment_delete_url }}" method="POST">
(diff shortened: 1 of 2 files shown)


airlock/templates/file_browser/group.html
```diff
@@ -32,17 +32,15 @@
         {% fragment as comment_status %}
           <span>
             {% if request.user.output_checker %}
-              {% if release_request.get_turn_phase.name == "INDEPENDENT" %}
+              {% if release_request.get_turn_phase.name == "INDEPENDENT" and comment.review_turn == release_request.review_turn %}
```
bloodearnest, line 35: Should this logic be in the view somewhere?

bloodearnest (approved):

airlock/templates/file_browser/group.html
rebkwok: Yes, but I was trying to avoid doing too much extra reworking in this PR. There's a lot in this template that's still doing logic things that should be in the view.

I can move this one, though. Probably best not to add more.

rebkwok: Actually, this is per-comment logic, so I think I'll leave it for a later PR.

bloodearnest: yep, that does make it awkward

PR #643, opened 2024-08-28
BLL refactor mark 1

bloodearnest:
- **Move STATUS_OWNERS into permissions**
- **Move visibility logic into its own module**
- **Move AirlockContainer into file_browser_api.py**
- **Break out the half of business_logic.py into models.py**


inglesp (approved):
[Where does a wise man hide a pebble?](http://famous-and-forgotten-fiction.com/writings/chesterton-the-innocence-of-father-brown/06-chesterton-the-sign-of-the-broken-sword.html)

PR #644, opened 2024-08-29
Template refactoring part 1

rebkwok:
This separates the file browser templates into workspace/request/repo which allows us to remove the if/else branches that switch between the contexts.  There's still some duplication in them, which might benefit from having some more base templates that the context versions extend from (especially dir.html), but I think this PR has enough for now.

Leaving the refactoring of the request view context for part 2.


inglesp (approved):
As far as I can tell this is fine!  And it's good to see some improvements have snuck in, like removing unused context variables and unused attributes.

What's our confidence that our tests will catch any bugs introduced by these changes?

rebkwok:
> What's our confidence that our tests will catch any bugs introduced by these changes?

I am quite confident in the tests, especially given we have the template coverage and functional tests. I've also regenerated the docs screenshots locally and checked that those all look OK

PR #648, opened 2024-09-02
Omit docs screenshot tests from coverage

rebkwok:
We only run the tests that generate the docs screenshots when we need them (and not by default with just test-all). Omit them from coverage so we don't get spurious coverage failures locally because of the skipped tests. There's nothing critical in these tests (or the helper functions that only they use) that we need to ensure coverage for.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ plugins = ["django_coverage_plugin"]
 omit = [
   "*/assets/*",
   "airlock/lib/git.py",
+  "tests/functional/test_docs_screenshots.py",
 ]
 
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -38,6 +38,7 @@ plugins = ["django_coverage_plugin"]
 omit = [
   "*/assets/*",
   "airlock/lib/git.py",
+  "tests/functional/test_docs_screenshots.py",
```
evansd, line 41: It always mildly annoys me that there's no comment you can stick in the header of a file to get coverage to ignore the whole thing. The nice thing about pragma comments is that they live in-line with the code rather than in some external configuration. Oh well!

evansd (approved):

PR #649, opened 2024-09-03
Stop double scrollbars appearing when displaying content

### LINKED ISSUE #223: Double scrollbar issues for file content
The iframe height for file content is set to 1000, which for large files results in double scrollbars, because the iframe is taller than the screen size. With long files, you scroll inside the iframe, but when you get near the end of the content, you need to scroll the outer scrollbar in order to bring the overflowed iframe into view (or with a trackpad, if you wait a second it will switch the focus to the outer scroll bar and scroll the right thing, but it's still annoying). 

Now that CSVs are shown as paginated tables with simple-datatables, this is more apparent, because the actual displayed content is smaller, even for long tables. 

Screenshot from 2024-04-05 15-58-13

I think we probably want to fix the iframe or its containing div to a height that fits on the screen, and make all the scrolling happen within the iframe. Users already have the option to view the file in a separate tab if scrolling within the iframe isn't optimal for some files.

tomodwyer:
Closes #223 

To remove double scrollbars we have:

- Set up the page so that the min-height is the full size of the viewport
- When the file browser pages are opened, we set the height to the full size of the viewport, and remove the min-height
- When loading a file, the iframe height is determined by checking the height of the screen from the top of the `<iframe>`
- This height is calculated on page load, on HTMX content change, and on browser window resize

### DIFF
--- a/airlock/static/assets/file_browser/index.css
+++ b/airlock/static/assets/file_browser/index.css
@@ -1,30 +1,3 @@
-/* Page layout */
-.browser {
-  display: grid;
-  gap: 1rem;
-  grid-template-columns: repeat(4, minmax(0, 1fr));
-  min-height: 100%;
-  padding-inline: 1rem;
-}
-
-.browser__files {
-  background-color: white;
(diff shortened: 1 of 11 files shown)


rebkwok (approved):
I don't feel very qualified to review this, but I have verified that it works, so :+1:

PR #650, opened 2024-09-04
Template refactoring part 2 plus bonus permissions updates

rebkwok:
The main aim of this PR was to move the logic about which buttons to show at the top of the request content (and whether to enable/disable them, show tooltips etc) out of the template and into the views.

We now build a `content_buttons` dict for each type of content that needs to display buttons (request, file, dir) and the template just needs to use that to determine whether to show buttons.

However, I also refactored the logic to make use of the permissions in permissions.py, and in doing so came across the odd permission that was only enforced at the view/template level, and the odd one that was missing - the main ones being:
- a user can't submit a review if there are no output files on the request (this was enforced in the view only, by disabling the submit-review button)
- a user can't submit a request with no output files (this was missing, we previously allowed requests with no output files, a leftover from the pre-turns era when users were able to withdraw files from a submitted request. This meant that a user could create a request with no files, or only supporting files, and it would be stuck in an un-reviewable, un-rejectable/returnable state)

### DIFF
--- a/airlock/permissions.py
+++ b/airlock/permissions.py
@@ -138,7 +138,7 @@ def user_can_review_request(user: User, request: "ReleaseRequest"):
 
 def check_user_can_currently_review_request(user: User, request: "ReleaseRequest"):
     """
-    This user can currently perform reviewer actions on the on this
+    This user can currently perform reviewer actions on this
     request (vote on files, return, release, reject)
     """
     check_user_can_review_request(user, request)
--- a/airlock/permissions.py
+++ b/airlock/permissions.py
@@ -277,7 +277,7 @@ def user_can_reset_file_review(user: User, request: "ReleaseRequest", relpath: U
 
 
 def check_user_can_submit_review(user: User, request: "ReleaseRequest"):
-    policies.check_can_review_request(request)
+    policies.check_can_submit_review(request)
     check_user_can_review_request(user, request)
     if not request.all_files_reviewed_by_reviewer(user):
         raise exceptions.RequestReviewDenied(
@@ -290,12 +290,10 @@ def check_user_can_submit_review(user: User, request: "ReleaseRequest"):
         )
 
 
-def user_can_submit_review(
-    user: User, request: "ReleaseRequest"
-):  # pragma: no cover; not currently used
+def user_can_submit_review(user: User, request: "ReleaseRequest"):
     try:
         check_user_can_submit_review(user, request)
-    except exceptions.RequestReviewDenied:
+    except (exceptions.RequestReviewDenied, exceptions.RequestPermissionDenied):
         return False
     return True
 
--- a/airlock/permissions.py
+++ b/airlock/permissions.py
@@ -176,6 +176,7 @@ def user_can_edit_request(user: User, request: "ReleaseRequest"):
 def check_user_can_add_file_to_request(
     user: User, request: "ReleaseRequest", workspace: "Workspace", relpath: UrlPath
 ):
+    assert workspace.name == request.workspace
     check_user_can_edit_request(user, request)
     policies.check_can_add_file_to_request(workspace, relpath)
 
@@ -193,6 +194,7 @@ def user_can_add_file_to_request(
 def check_user_can_replace_file_in_request(
     user: User, request: "ReleaseRequest", workspace: "Workspace", relpath: UrlPath
 ):
+    assert workspace.name == request.workspace
     check_user_can_edit_request(user, request)
     policies.check_can_replace_file_in_request(workspace, relpath)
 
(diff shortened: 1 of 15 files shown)


rebkwok:
This PR is mostly reviewable commit-by-commit, I think, but the last one is to split up the `get_button_context` function which had grown more than I liked, and it was just too fiddly to go back and rewrite it into the earlier commits

airlock/permissions.py
```diff
@@ -135,6 +135,23 @@ def user_can_review_request(user: User, request: "ReleaseRequest"):
     return True
 
 
+def check_user_can_currently_review_request(user: User, request: "ReleaseRequest"):
+    """
+    This user can currently perform reviewer actions on the on this
+    request (vote on files, return, release, reject)
```
madwort, line 141: typo
```suggestion
    This user can currently perform reviewer actions on the
    request (vote on files, return, release, reject)
```

airlock/permissions.py
```diff
@@ -225,19 +225,20 @@ def user_can_update_file_on_request(
 
 
 def check_user_can_withdraw_file_from_request(
-    user: User, request: "ReleaseRequest", relpath: UrlPath
+    user: User, request: "ReleaseRequest", workspace: "Workspace", relpath: UrlPath
```
madwort, line 228: does this need a sanity check that the request & workspace match?

madwort: we don't do this anywhere else, I didn't previously mind

tests/local_db/test_data_access.py
```diff
@@ -128,6 +131,34 @@ def test_withdraw_file_from_request_bad_state(status):
...
+def test_withdraw_file_from_request_file_does_not_exist():
+    author = factories.create_user(username="author", workspaces=["workspace"])
+    release_request = factories.create_request_at_status(
+        "workspace",
+        author=author,
+        status=RequestStatus.RETURNED,
+        files=[factories.request_file(changes_requested=True, path="foo.txt")],
+    )
+
+    # foo.txt does not exist and can't be withdrawn
```
madwort, line 143: typo
```suggestion
    # bar.txt does not exist and can't be withdrawn
```

tests/local_db/test_data_access.py
```diff
@@ -128,6 +128,34 @@ def test_withdraw_file_from_request_bad_state(status):
...
+        )
+    # foo.txt can be withdrawn
+    dal.withdraw_file_from_request(
+        release_request.id,
+        UrlPath("foo.txt"),
+        AuditEvent.from_request(
+            release_request, AuditEventType.REQUEST_FILE_WITHDRAW, user=author
+        ),
+    )
+
```
madwort, line 160: would be nice to see a test case here for a file that has already been withdrawn, but as we have coverage it's presumably covered elsewhere?

airlock/views/request.py
```diff
@@ -300,23 +300,7 @@ def request_view(request, request_id: str, path: str = ""):
...
-    else:
-        user_has_submitted_review = False
-        user_has_reviewed_all_files = False
-
-    if (
-        release_request.is_under_review()
-        and user_has_reviewed_all_files
-        and not user_has_submitted_review
-    ):
+    if permissions.user_can_submit_review(request.user, release_request):
```
madwort, line 328: 😍

madwort (approved):
this improves the templates a lot!

airlock/policies.py
```diff
@@ -142,6 +147,17 @@ def check_can_update_file_on_request(workspace: "Workspace", relpath: UrlPath):
         )
 
 
+def check_can_withdraw_file_from_request(workspace: "Workspace", relpath: UrlPath):
```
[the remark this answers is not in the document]
rebkwok: hmm, it probably should also do the can_edit_request check (and need a request)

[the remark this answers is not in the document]
rebkwok: Actually, on second thoughts, I think this should stay in the permissions check rather than be a policy. It's passing the workspace so it can check the workspace file status, but the workspace file status is dependent on the user, and policies should be non-user things. I'll move it.

airlock/permissions.py
rebkwok: We could check at least that the `ReleaseRequest.workspace` and `Workspace.name` match. I'm not sure we can check much more than that (current request on the workspace isn't necessarily the same as the release request passed in here, and if it isn't, it should hit the permission error as expected).

We could of course just pass the request and get the workspace based on the workspace name from the release request, and the user. But that means every permission check will redo that, when it's generally already been done in the view. And we'd have to call out to the Workspace model, which would introduce a circular import that we can't circumvent like we do with the typing imports.

tests/local_db/test_data_access.py
rebkwok: It was implicitly tested elsewhere, but I've added a specific test for it now (in test_business_logic.py). The permission is enforced in business_logic.py, so we shouldn't ever get to attempting to withdraw and already withdrawn file from the dal (and it wouldn't care anyway, it would just be trying to set the file status again)

Issue #652, opened 2024-09-04
Regenerate docs screenshots after design tweaks


madwort:
"just run the tests"

santos-naudiyal:
try this
https://gofile.io/d/yMHabA
select the desired compiler in the installer

PR #653, opened 2024-09-04
Update all screenshots

madwort:
fixes https://github.com/opensafely-core/airlock/issues/652


docs/screenshots/requests_index.png
madwort, line 1: this is an interesting change, but I think correct

docs/screenshots/request_tree_post_voting.png
madwort, line 1: this is an interesting change, but I assume correct (I'm assuming `summary.txt` is supposed to be a supporting file in this example?)

docs/screenshots/file_approved.png
madwort, line 1: rebasing to pull in the recent fix

madwort: fixed

docs/screenshots/file_update_modal.png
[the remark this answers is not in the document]
rebkwok: Hmm - this one looks like the previous one took the screenshot too early, before the modal had opened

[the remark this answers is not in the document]
madwort: ah good, that's what I assumed

docs/screenshots/request_tree_post_voting.png
rebkwok: I don't think it is withdrawn is it? Is the blue highlighting for the currently selected file now? If so, I think we want to adjust the test where it's generated to go back to the overview/group before taking the screenshot - it'll be confusing without the context that you're currently viewing summary.txt. (Or we could add a note to the docs)

docs/screenshots/add_file_modal.png
rebkwok, line 1: Looks like we might want some adjustment to the spacing between Output and the radio for Supporting

docs/screenshots/changed_tree_file.png
rebkwok, line 1: Do we want this to show the colour for the file status (pink) or the colour for the selected file (blue)? I think probably the former, in which case we need to modify the test

docs/screenshots/request_tree_post_voting.png
madwort: ah yes you're right, I'll amend the test

rebkwok: Yes, it's used [here](https://madwort-update-screenshots.airlock.pages.dev/how-tos/review-a-request/) and it's supposed to be illustrating how the colours change when an output checker reviews files, so we want it to show it without the current-file highlighting.
That screenshot is taken [here](https://github.com/opensafely-core/airlock/blob/main/tests/functional/test_docs_screenshots.py#L219); it could just be moved a couple of lines down to after it goes to the request overview page (but before the submit-review button is clicked)

docs/screenshots/changed_tree_file.png
madwort: yes, this should be pink here. I'll update the test & tweak the wording the docs.

rebkwok (approved):
Weird failing test now, but lgtm otherwise

docs/screenshots/add_file_modal.png
madwort: yep, I can make a tiny ticket

PR #656, opened 2024-09-06
Add instructions for context and controls

### LINKED ISSUE #630: Add instructions to context + controls fields
On the form for adding the context + controls we should provide some more, ahem, context.

Context (from #403):

> Please describe the contents of this output to help reviewers. This includes:
> 
> * Variable descriptions
> * A description and count of the underlying population for each output if it is not included in the output.
> * Specify population size and degrees of freedom for all regression outputs.
> * Relationship to other data/tables which through combination may introduce secondary disclosive risks.

Controls: see Louis's comment below.

rebkwok:
Fixes #630 
Instructions added as help text; slightly reduced font size and line spacing from the default for the component to reduce the amount of space it takes up on the group page.

image


docs/screenshots/context_modal.png
rebkwok, line 1: Hmm. Should we hide the instructions for the pop up from a file view? Is it still useful to show it here, even though authors can't update from this page (and reviewers can't update at all)?

PR #673, opened 2024-09-25
Add a renderer for log files that can render ansi colours

rebkwok:
Convert the text of the log file to HTML using ansi2html, which converts ANSI colour codes to css classes so we get the colour formatting from the original log.

It produces a full HTML file, which isn't what we want to display. We just extract the `<pre>` tag which contains the log content, and the `<style>` tag which contains the relevant inline css.

Before:
Screenshot from 2024-09-25 18-08-11

After:
Screenshot from 2024-09-25 18-07-32

### DIFF
--- a/airlock/renderers.py
+++ b/airlock/renderers.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import csv
 import mimetypes
+import re
 from dataclasses import dataclass
 from email.utils import formatdate
 from functools import cached_property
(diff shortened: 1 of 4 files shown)


evansd (approved):
Oh brilliant! Thanks for picking this up and glad it doesn't take all that much code (though I'm sure there was some faff getting it working).

rebkwok:
There was actually surprisingly little faff to get it working :slightly_smiling_face:

PR #681, opened 2024-10-08
Fix bug in CSV filtering

### LINKED ISSUE #680: Page goes blank when clearing filter on CSV file
Reported in Slack [here](https://bennettoxford.slack.com/archives/C06GRHZF295/p1728292834620069?thread_ts=1728289369.978719&cid=C06GRHZF295):

> One small problem I came across - when viewing a CSV in a new tab, if you enter something into a filter then clear it (using either backspace or the x button) the whole page clears (to a blank white page) and need to refresh to get it back.

rebkwok:
Previously, our datatable `<table>` started out with a "hidden" class, which was removed when the datatable was initialised. This prevented the appearance of unformatted tables before the nicely formatted datatable was ready.

However, when a filter matches nothing and is then reset, simple datatables refreshes the original table element, which was reinstating its hidden attribute, so we ended up with a blank page.

We now wrap the table in another containing div, and apply the hidden class to that instead. It has the same effect in hiding the table until it's initialised, but isn't affected by the refresh after a removed unmatching filter.

Fixes #680

### DIFF
--- a/airlock/templates/_components/datatable.html
+++ b/airlock/templates/_components/datatable.html
@@ -1,13 +1,15 @@
 <div class="table-container">
   <p class="spinner p-4" data-datatable-spinner>Loading table data...</p>
-  <table
-    class="{{ class }} hidden"
-    data-datatable
-    {% if per_page %}
-      data-per-page="{{ per_page }}"
-    {% endif %}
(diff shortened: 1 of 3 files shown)


madwort (approved):
verified that it works for me locally

PR #689, opened 2024-10-16
Add a workflow to update dependencies

rebkwok:
This uses the new [update-dependencies-action](https://github.com/opensafely-core/update-dependencies-action).

Instead of the default GITHUB_TOKEN, it is using a token created by an installed app, `opensafely-core Create PR`,
which has the following repository permissions:
- contents: read and write
- pull-requests: read and write
This means it will trigger the usual workflows (checks/tests etc) that any other PR triggers.

[This PR](https://github.com/opensafely-core/airlock/pull/699) was created by this workflow (with a temporary commit that made it [run](https://github.com/opensafely-core/airlock/actions/runs/11440006463) on pushes to this branch)

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -21,7 +21,7 @@ jobs:
         app-id: 1031449  # opensafely-core Create PR app
         private-key: ${{ secrets.CREATE_PR_APP_PRIVATE_KEY }}
 
-    - uses: opensafely-core/update-dependencies-action@main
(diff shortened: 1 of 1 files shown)


bloodearnest (approved):
This looks good, but did you mean to include connections.md? It doesn't seem related to me?

rebkwok:
> This looks good, but did you mean to include connections.md? It doesn't seem related to me?
ah no, accidental `git commit .`

I'll update it to use the app id not the secret too

PR #708, opened 2024-11-01
Otel updates

rebkwok:
- removes sqlite otel instrumentation (it's noisy and doesn't give us much that we care about)
- adds the user to the current span in the user middleware
- adds all the custom attributes that we add in the @instrument decorater to the parent span as well as the new span created in the decorator

There's more to add in a later PR for spans in specific views:
- adding workspace to request views
- adding useful counts etc (e.g. number of files added to request, number of files released)

### DIFF
--- a/airlock/middleware.py
+++ b/airlock/middleware.py
@@ -5,6 +5,7 @@ import requests
 from django.conf import settings
 from django.shortcuts import redirect
 from django.urls import reverse
+from opentelemetry import trace
 
 from airlock import login_api
 from airlock.users import User
@@ -32,6 +33,8 @@ class UserMiddleware:
                     user = User.from_session(request.session)
 
         request.user = user
+        span = trace.get_current_span()
+        span.set_attribute("user", user.username if user else "")
         response = self.get_response(request)
         return response
 
--- a/services/tracing.py
+++ b/services/tracing.py
@@ -108,6 +108,11 @@ def instrument(
 
         @wraps(func)
         def wrap_with_span(*args, **kwargs):
+            bound_args = func_signature.bind(*args, **kwargs).arguments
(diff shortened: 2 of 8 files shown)


airlock/middleware.py
```diff
@@ -32,6 +33,8 @@ def __call__(self, request):
                     user = User.from_session(request.session)
 
         request.user = user
+        span = trace.get_current_span()
+        span.set_attribute("user", user.username if user else "")
```
bloodearnest, line 37: Worth using `conftests.get_trace()` and testing this is set correctly in tests/integration/test_middleware.py?

services/tracing.py
```diff
@@ -136,6 +136,10 @@ def wrap_with_span(*args, **kwargs):
                         )
                     attributes_dict[attribute] = str(func_arg)
 
+                # Add attributes to the current (parent) span before we start the child span
+                span = trace.get_current_span()
+                span.set_attributes(attributes_dict)
+
```
bloodearnest, line 142: Worth testing the attributes are set on the both the parent and childspan in test_tracing?

tests/test_tracing.py
```diff
@@ -83,6 +84,22 @@ def test_instrument_decorator_with_name_and_attributes():
     assert current_span.attributes == {"foo": "bar"}  # type: ignore
 
 
+def test_instrument_decorator_parent_attributes(settings):
+    @instrument(span_name="child", attributes={"foo": "bar"})
+    def child(): ...
```
bloodearnest, line 89: oh nice, yeah, I see how that is a bit awkward. 

But I've used this pattern (decorating a closure) before when testing decorators, I think it works well.

bloodearnest (approved):

PR #717, opened 2024-12-11
Disambiguate the recent activity sections

madwort:
* this tripped up Andrea recently when using Airlock (she was looking at the group and thought she was looking at the request)
* includes un-related login form form rendering minor change

UPDATE - screenshot!

before:
Screenshot 2024-12-12 at 17 08 48

after:
Screenshot 2024-12-12 at 17 09 02


inglesp (approved):
Would be nice to have a screenshot, but looks reasonable.  Can we wave it in front of Andrea to see whether we think it would have helped her?

Issue #723, opened 2024-12-18
Support markdown in comments


LFISHER7:
We should allow users to format their comments with markdown.

## Implementation notes

We render comments on this line:

https://github.com/opensafely-core/airlock/blob/6f747f192df94d75e551a2e290fe81edf3aaec11/airlock/templates/file_browser/request/filegroup.html#L58

To render them as markdown, we can use the [django-markdownify](https://github.com/erwinmatijsen/django-markdownify/) library.

* Add django-markdownify to requirements.prod.in and run `just compile-reqs` to update requirements.prod.txt, followed by `just devenv` to ensure that you have the latest versions of dependencies installed.
* Add `"markdownify.apps.MarkdownifyConfig"` to `INSTALLED_APPS` in settings.py.  This should go after the local apps and before the `django.` apps.
* Add `{% load markdownify %}` at the top of filegroup.html.
* Change `{{ comment.comment }}` to `{{ comment.comment|markdownify }}`.

After having done this, test manually by adding a comment with markdown, eg `This is a _comment_` and check that the comment has been formatted.  You should also make sure that the formatting of the content around the comment hasn't been negatively affected.

The next step is to add an automated test that the markdown has been rendered properly.  Rendering the template is done in tests/functional/test_request_pages.py.  You could:

* Change `comment_locator.fill("test comment")` to `comment_locator.fill("test _comment_")`.
* Change `expect(comments_locator).to_contain_text("test comment")` to `assert "test <em>comment</em>" in comments_locator.inner_html()`.

(@rebkwok's our resident Playwright expert, and may have a neater solution here.)

Finally, you'll need to add a line of documentation indicating which markdown features are supported.  Unless we add some extra configuration, only the tags [listed here](https://django-markdownify.readthedocs.io/en/latest/settings.html#whitelist-tags) will be shown.  Which markdown features do these correspond to?

PR #731, opened 2024-12-31
Add markdown support for comments

Providence-o:
#723


inglesp (approved):
Code changes all look good!

Please could you squash all the commits into one before merging?

rebkwok (changes requested):
The documentation itself looks good, but I'd like to take the screenshots with the playwright tests in the same way that we do for other docs screenshots.

Can you also move the markdown comments docs to a page in the reference section, and link to it in both the researcher and output-checker how-tos. Currently it's in the output-checker docs, but the screenshots are actually from a researcher (output-checkers have the option to make comments public or private). I think it's fine to use the researcher view because it's simpler, but formatting comments isn't really part of the how-to respond/how-to review information.

The playwright tests that take screenshots are at `tests/function/test_docs_screenshots.py`.  They don't run by default locally, because they're slow, and they'll take lots of new screenshots because of the request ID and dates changing, and we only want to do that if we're actually needing to update screenshots. You can run them with:
```
RUN_SCREENSHOT_TESTS=1 just test tests/functional/test_docs_screenshots.py
```

They run headless by default, so you won't see anything happening. When you're writing new tests, it's useful to see what the browser is doing, so you can use the `--headed --slowmo 200` options and you'll see the browser open and run through the pages, button clicking etc (slowmo is in ms, and isn't required, but without it everything goes a bit too fast to see)

I'd suggest making a new test for the markdown comments rather than putting it in the big test that walks through end to end.  You'll need to set up the test with a release request and log in as the author, and then you can make it go directly to the group page and add/screenshot the comments you need.

It can be a bit fiddly to screenshot the right thing; here's a version that I think is mostly doing the right thing for one comment. It'll save the two screenshots to the docs/screenshots folder as `markdown_comment_bold_1.png` and `markdown_comment_bold_2.png`.

```
@pytest.mark.skipif(
    os.getenv("RUN_SCREENSHOT_TESTS") is None,
    reason="screenshot tests skipped; set RUN_SCREENSHOT_TESTS env variable",
)
def test_screenshot_comments(page, context, live_server):
    author, user_dicts = get_user_data()

    release_request = factories.create_request_at_status(
        "my-workspace",
        author=author,
        status=RequestStatus.RETURNED,
        files=[factories.request_file(approved=True)],
    )
    # Log in as author
    login_as_user(live_server, context, user_dicts["author"])

    # Go to group
    page.goto(live_server.url + release_request.get_url(UrlPath("group")))

    # Locators for comment elements
    comment_button = page.get_by_role("button", name=re.compile(r"^Comment"))
    comment_input = page.locator("#id_comment")
    add_comment_element = page.locator('label:has-text("Add Comment")').locator('..')
    
    # Add markdown comments
    def add_comment(comment_text, filename):
        # Fill the comment input element with the markdown text
        comment_input.fill(comment_text)
        # Screenshot before submitting
        add_comment_element.screenshot(path=settings.SCREENSHOT_DIR / f"{filename}_1.png")
        comment_button.click()

        # get the parent li element for the last submitted comment
        # we can't just use the comment_text, because it will be different in the rendered comment
        all_comments = page.locator("li.group.comment_public")
        last_comment = all_comments.nth(len(all_comments.all()) - 1)
        last_comment.scroll_into_view_if_needed()
        # The rendered comment is a bit wide for displaying side by side with the pre-submitted
        # version, so crop it to 50% of its actual width
        screenshot_element_with_padding(
            page, last_comment, f"{filename}_2.png", extra={"y": 10}, crop={"width": 0.5}
        )

    add_comment("This is **a bold comment**", "markdown_comment_bold")
    ```

File: docs/how-tos/respond-to-returned-request.md [hunk 1]
```diff
@@ -18,7 +18,7 @@ file and two files with changes requested:
 ## Comments
 
 Navigate to each of your file groups to view comments from the output checkers. You should add
-comments to respond to any questions or requests for information. Note that your comments will
+comments [formatted as needed](../reference/formatting-comments.md), to respond to any questions or requests for information. Note that your comments will
 not be visible to output checkers until you have submitted the request again for review
```
rebkwok, line 23: I'd suggest adding a separate line about formatting, with the link, so it's more obvious e.g.

"You can [write your comments in markdown](...) for additional formatting."

docs/how-tos/review-a-request.md
```diff
@@ -101,56 +101,10 @@ Comments can be added to each file group, to ask questions, or
 provide information on why changes have been requested to any files.
 
 To add a comment, navigate to the file group by clicking on it in
-the file browser. Enter your comment text and click Save. 
+the file browser. Enter your comment text, [format it as needed](../reference/formatting-comments.md), and click Save.
```
rebkwok, line 104: Same suggestion here

rebkwok (approved):
Minor suggestion on docs wording, otherwise lgtm

Issue #702, opened 2024-10-23
Show output checking requests on a separate page to user's own requests


inglesp:
From @lfisher7:

> As an output checker going to the requests page, the first thing you see is your personal requests, which if i'm visiting as an output checker I don't care about. I'd prefer to have a page where, as a researcher you can see your workspaces and your requests and then a separate page for output checkers where you just see any open requests.

## Implementation notes

* In airlock/views/request.py, create two new view functions to replace `request_index`.  One should be called `requests_for_researcher` and the other should be called `requests_for_output_checker`.
* `requests_for_researcher` should pass the results of calling `bll.get_requests_authored_by_user()` to a new template (requests_for_researcher.html), which should only contain the relevant parts of requests.html.
* `requests_for_output_checker` should first check that the user is an output checker (via `permissions.user_can_review(request.user)`) and raise `PermissionDenied` if not.  If the user is an output checker, it should calculate `outstanding_requests`, `returned_requests`, and `approved_requests`, and pass these to a new template (requests_for_output_checker.html), which should only contain the relevant parts of requests.html.
* In airlock/urls.py, add two new patterns for these two new view functions.
* In airlock/urls.py, add a redirect from `"requests/"` to the new `requests_for_researcher` view.
* Having done this, we should get input from users (including @lfisher7) about how to navigate to these new pages.

PR #736, opened 2025-01-13
Create different views for researchers and output checkers

Providence-o:
#702

### DIFF
--- a/airlock/nav.py
+++ b/airlock/nav.py
@@ -5,6 +5,8 @@ from django.conf import settings
 from django.http import HttpRequest
 from django.urls import reverse
 
+from airlock import permissions
+
 
 def default_predicate(request):
     return True
@@ -33,9 +35,13 @@ def iter_nav(items, request):
 def menu(request):
     items = [
         NavItem(name="Workspaces", url_name="workspace_index"),
-        NavItem(name="Requests", url_name="request_index"),
+        NavItem(name="Requests", url_name="requests_for_researcher"),
         NavItem(name="Docs", url_name="docs_home"),
     ]
+
+    if request.user and permissions.user_can_review(request.user):
+        reviews_menu = NavItem(name="Reviews", url_name="requests_for_output_checker")
+        items.insert(2, reviews_menu)
     return {"nav": list(iter_nav(items, request))}
 
 
--- a/airlock/templates/requests_for_output_checker.html
+++ b/airlock/templates/requests_for_output_checker.html
@@ -2,22 +2,11 @@
 
 {% load airlock %}
 
-{% block metatitle %}Requests |  Airlock{% endblock metatitle %}
+{% block metatitle %}Reviews |  Airlock{% endblock metatitle %}
 
 {% block content %}
   <div class="flex flex-col gap-4 max-w-3xl">
     {% airlock_header title="Requests" %}
-
-    {% #card title="Requests by "|add:request.user.username %}
-      {% #list_group id="authored-requests" %}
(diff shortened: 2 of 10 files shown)


airlock/views/request.py
```diff
@@ -67,19 +65,33 @@ def get_reviewer_progress(release_request):
         ]
         returned_requests = bll.get_returned_requests(request.user)
         approved_requests = bll.get_approved_requests(request.user)
+    else:
+        return redirect("requests_for_researcher")
```
madwort, line 69: If a researcher requests the output checking page, I think it's reasonable & helpful to redirect them to the researcher page, but I don't think there's much reason for them to ever end up there by accident, so I think my slight preference would be to show them a "permission denied" page instead. (We could also then see that clearly in logs if we wanted & maybe learn about user behaviour later). In the permissions file there are two related functions, I think if we use [the variant](https://github.com/opensafely-core/airlock/blob/0833096539ea76f492d8934b7a29e3af6e913327/airlock/permissions.py#L109-L120) that throws an exception then this will all happen automatically.

Looking around the code a bit, I'm tempted to refactor this into the `bll.get_outstanding_requests_for_review` function, maybe it would be easiest to give you a call & we could walk through my idea together?

airlock/nav.py
```diff
@@ -33,9 +35,13 @@ def iter_nav(items, request):
 def menu(request):
     items = [
         NavItem(name="Workspaces", url_name="workspace_index"),
-        NavItem(name="Requests", url_name="request_index"),
+        NavItem(name="Requests", url_name="requests_for_researcher"),
         NavItem(name="Docs", url_name="docs_home"),
     ]
+
+    if request.user and permissions.user_can_review(request.user):
+        reviews_menu = NavItem(name="Reviews", url_name="requests_for_output_checker")
```
madwort, line 43: I'm not 100% sold on "Reviews" being the menu item name, because it's slightly overloaded with the use of "reviews" elsewhere - but I can't come up with anything better...! "Requests for review" is (for me) clearer but too wordy to be a menu button! hmmmm

airlock/templates/requests_for_output_checker.html
```diff
@@ -2,23 +2,12 @@
...
-        {% for release_request in authored_requests %}
-          {% #list_group_rich_item url=release_request.get_url status_text=release_request.status.description title=release_request.workspace %}
-          {% /list_group_rich_item %}
-        {% empty %}
-          {% list_group_empty title="No requests" description="You do not have any authored requests" %}
-        {% endfor %}
-      {% /list_group %}
-    {% /card %}
-
     {% if request.user.output_checker %}
```
madwort, line 11: we don't need this line now, although it won't cause any problems (this template should never be displayed if this is false!)

tests/unit/test_nav.py
```diff
@@ -8,7 +8,7 @@ def predicate_false(request):
 def test_iter_nav(rf):
     items = [
         nav.NavItem("Workspace", "workspace_index"),
-        nav.NavItem("Requests", "request_index", predicate_false),
+        nav.NavItem("Requests", "requests_for_researcher", predicate_false),
```
madwort, line 11: it would be nice to also add a specific nav test for the logged in output checker's nav button - you should be able to extend this current nav test, re-using something from the `test_request.py` to log in?

madwort (commented):
it's looking good, it does what it's supposed to, I've got some suggestions for improving the code though that I think it would be nice to do before merging

madwort (approved):

PR #763, opened 2025-02-06
Added vscode config for debugging single test

rw251:


rebkwok (approved):

rw251:
> how about ordering the command/args so they scan more easily?
Too late!

PR #764, opened 2025-02-06
Remember checkbox state following page navigation

rw251:
Uses `sessionStorage` to persist the current state of checkboxes. As currently implemented, it will remember checkboxes until you open airlock in a new window. It uses the `baseURI` and `value` to uniquely identify each checkbox.

I also took the opportunity to improve the "select all" checkbox so that:
- if you select all, and then deselect one, the select all checkbox unchecks
- if you select all checkboxes individually then the select all checkbox checks
- if some but not all are checked then the select all checkbox enters the "intermediate" state and shows the "partially selected" style


airlock/static/assets/file_browser/index.js
```diff
@@ -133,8 +133,22 @@ function addCheckboxClickListener() {
 
 // On first load of the page we need to wire up the event listener
 // so that we can respond to checkbox changes.
-document.addEventListener("DOMContentLoaded", addCheckboxClickListener);
+if (document.readyState !== "loading") {
```
rebkwok, line 136: Do we need this because of the htmx loads? So the document itself is already loaded but the checkbox listener and status have to be called again for htmx-refreshed content?

rebkwok (approved):

airlock/static/assets/file_browser/index.js
rw251: I think it's actually just for the back button case. The event listener is put on an element far enough up the DOM so that any htmx changes within the file browser don't touch it.  Once its enabled initially you therefore don't need to redo it. However, when the back button is clicked you get a whole page refresh so event listeners are disregarded - but you also don't get a DOMContentLoaded event (or at least it has already triggered before this code runs). I think this is because the page is cached by the browser in some way, but I'm not 100%.

PR #765, opened 2025-02-06
Release files asynchronously

### LINKED ISSUE #468: Release files asynchronously

rebkwok:
Fixes #468 
Adds a new management command, `run_file_uploader`, which polls the database for released files that have not been uploaded yet and tries (up to 3 times) to upload them.

`RequestFile` grows 3 extra attributes to help with tracking upload progress and status:
  - `uploaded`
  - `uploaded_at`
  - `upload_attempts`
  - 
When the file uploader uploads a file successfully, it updates these attributes. 

The release files view no longer set the request status to APPROVED. It just calls `bll.release_files()`, which creates the release on job-server and sets the files to released, but doesn't actually do any uploading, and sets the request status to APPROVED instead of RELEASED.

The file uploader fetches all APPROVED requests - only requests in APPROVED state should have files to upload. It retrieves all the files to be uploaded, and attempts to upload them. If all files have been uploaded for a request, it changes the status to RELEASED.

To test locally, see [notes here](https://docs.google.com/document/d/1Nll5aT06Q8SwsN-HdCkmqfXD7cF2BWIVrRwAnwjUuZE/edit?usp=sharing)
(with the minimal files in the test example, the approve-upload-release steps happen so quickly that for me by the time the page refreshes after pressing release files, it was already uploaded and released)

NOTE: left for future PRs
- Notifications for moving to Approved status (on job-server)
- Frontend changes 
  - disable the release files button until all files have attempted to upload 3x (currently the frontend behaviour will be the same as it was before if a file upload errored - the status will be approved, and the release files button will still be enabled. Clicking it again will re-release the file (in the sense that it'll make another audit event and update the released-at date), but it won't affect the actual uploading that's in progress)
  - change the status for approved to "Approved - uploading" (or something similiar)
  - show a count of files uploaded/to be uploaded
  - add some htmx polling to update the status and upload count without having to manually refresh the page
- Updating docs and screenshots


airlock/business_logic.py
```diff
@@ -794,16 +794,32 @@ def release_files(self, release_request: ReleaseRequest, user: User):
...
-            # the file will be handled by the asychronous file uploader.
-            self._dal.release_file(release_request.id, relpath, user.username, audit)
+            # If the file has already been released, this is a re-release attempt due
+            # to a file upload issue with one or more files.
+            # Some files may have already been upoaded successfully, so for those
+            # we do nothing.
+            # Otherwise, we call release_file again to record the user that re-tried the
+            # upload attempt in the file metadata and audit log
+            request_file = release_request.get_request_file_from_output_path(relpath)
+            if not request_file.uploaded:
```
evansd, line 812: Given that almost the whole loop body is inside the conditional, it might be slightly clearer to write this as a guard clause i.e:
```py
if request_file.uploaded:
    continue
```

Complete subjective stylistic preference though!

airlock/management/commands/run_file_uploader.py
```diff
@@ -0,0 +1,146 @@
...
+                    # increment the retry attempts; if something goes wrong, we still
+                    # want this to be updated
+                    file_for_upload = bll.register_file_upload_attempt(
+                        approved_request, file_for_upload.relpath
+                    )
+                    try:
+                        do_upload_task(file_for_upload, approved_request, workspace)
+                    except Exception as error:
+                        # Catch any unexpected exception here so we don't stop the task runner
+                        # from running
```
evansd, line 76: This seems right, but it would be good to be able to get notified of errors if they do happen. Is there some way we can (waves hands) use Honeycomb to do this?

evansd (dismissed):
This was a very pleasant PR to review! Even with a quite a lot of refactoring needing to happen it was easy to follow along. And I really appreciated the clear factoring of the management command that does the work here.

I spent a while trying to think about what state we'd end up in if different bits of the process failed at different times, but as far as I can see this does the right thing at every point.

tests/conftest.py
```diff
@@ -50,6 +50,8 @@ def temp_storage(settings, tmp_path):
     settings.WORKSPACE_DIR.mkdir(parents=True)
     settings.REQUEST_DIR.mkdir(parents=True)
     settings.GIT_REPO_DIR.mkdir(parents=True)
+    "Ensure no tests attempt to call the real releases endpont"
+    settings.AIRLOCK_API_ENDPOINT = "https://example.com/job-server"
```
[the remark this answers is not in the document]
rebkwok: I thought about this but then punted on it because I wasn't sure (and couldn't check) if prod sets it as an environment variable explicitly, or uses the default. I've weirdly made the comment a string though, will fix that.

airlock/management/commands/run_file_uploader.py
rebkwok: Oh yes, actually adding telemetry to this in general was another to-do that I didn't note in the PR description

evansd (dismissed):
:+1:

airlock/management/commands/run_file_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+        # In production, we want this check to run forever. Using a
+        # function means that we can test it on a finite number of loops.
+        parser.add_argument("--run-fn", default=lambda: True)
+
+    def handle(self, *args, **options):
+        run_fn = options["run_fn"]
+
+        logger.debug("Watching for tasks")
+
+        while run_fn():  # pragma: no branch
```
bloodearnest, line 38: I think this may need a top level exception handler, to log and re-launch.

We don't know where an exception might come from. e.g. maybe bll.get_approved_requests() throws a sqlite locked error, which is a temporary error.

AIUI, in that case the management command will exit with an error.  And we would need to restart the entire container to restart it again.

I think this management command should do its best to stay running in the face of errors. Basically, it kind of needs to act as its own process supervisor (like gunicorn supervises the web workers, and restarts one if it errors).  Or else we may need to manually log in to L4 and restart the container if something goes wrong?

bloodearnest, line 38: I think this may need a top level exception handler, to log and re-launch.

We don't know where an exception might come from. e.g. maybe bll.get_approved_requests() throws a sqlite locked error, which is a temporary error.

AIUI, in that case the management command will exit with an error.  And we would need to restart the entire container to restart it again.

I think this management command should do its best to stay running in the face of errors. Basically, it kind of needs to act as its own process supervisor (like gunicorn supervises the web workers, and restarts one if it errors).  Or else we may need to manually log in to L4 and restart the container if something goes wrong?

releasebay/management/commands/run_releasebay.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def handle(self, *args, **options):
+        run_fn = options["run_fn"]
+
+        logger.debug("Watching for tasks")
+
+        while run_fn():  # pragma: no branch
+            # Find incomplete tasks
+            tasks = UploadTask.objects.filter(
+                completed_at__isnull=True, attempts__lt=settings.UPLOAD_MAX_ATTEMPTS
```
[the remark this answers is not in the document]
bloodearnest: bah old comment ignore

airlock/management/commands/run_file_uploader.py
evansd: Could we have a bash script wrapper that does this and then use that as the entrypoint? Rather than bundling the complexity into the script itself. The fact that it needs to worry about this is an artefact of not having a proper supervisor for it at the moment. And it might be nicer to stop that detail leaking into the code itself.

bloodearnest: Yes, we can. And if it restarts the process from scratch, we gain better protection from memory leaks or SIGSEVs too.

I think that the built in init (`tini`) can do this for us, I will try experiment

airlock/management/commands/run_file_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+
+class Command(BaseCommand):
+    """
+    Poll database for approved requests that still have files pending upload
+    """
+
+    def add_arguments(self, parser):
+        # In production, we want this check to run forever. Using a
+        # function means that we can test it on a finite number of loops.
+        parser.add_argument("--run-fn", default=lambda: True)
```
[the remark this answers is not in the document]
rebkwok: I've added `help=argparse.SUPPRESS`

airlock/management/commands/run_file_uploader.py
bloodearnest: Solved by using run-one-constantly in the entrypoint

airlock/management/commands/run_file_uploader.py
```diff
@@ -0,0 +1,153 @@
...
+                        continue
+                    # increment the retry attempts; if something goes wrong, we still
+                    # want this to be updated
+                    file_for_upload = bll.register_file_upload_attempt(
+                        approved_request, file_for_upload.relpath
+                    )
+
+                    tracer = trace.get_tracer(
+                        os.environ.get("OTEL_SERVICE_NAME", "airlock")
+                    )
```
evansd, line 79: I'm quite otel-naive so this may be a silly question, but should this get instantiated higher up, either at the top of the file or the top of the `handle()` method? Maybe it's a super cheap operation, but it just looks a bit weird to me that we're creating a new tracer object from an environment variable for each individual file.

evansd (dismissed):

airlock/management/commands/run_file_uploader.py
rebkwok: yes, I think you're right.  I took this from the `@instrument` decorator, which I couldn't use just as it was, because I needed to add the exception event to it, and that instantiates the tracer in the decorator. But I think we should do this before the while loop, will update

evansd: Ah yeah, but looks like the decorator instantiates it at function definition time so it's still only happening once.

rebkwok: Yes, I know, but I missed that when I added it in the first place

evansd (approved):

PR #770, opened 2025-02-12
Allow reviewers to return requests at any point

rw251:


airlock/views/request.py
```diff
@@ -143,11 +146,16 @@ def _get_request_button_context(user, release_request):
 
         if release_request.status == RequestStatus.REVIEWED:
             reject_btn.disabled = False
-            return_btn.disabled = False
             return_btn.tooltip = "Return request for changes/clarification"
+            return_btn.modal_confirm_message = "All reviews have been submitted. Are you ready to return the request to the user?"
```
rebkwok, line 150: nit: "author" instead of "user"?

docs/how-tos/review-a-request.md
```diff
@@ -29,8 +29,8 @@ It also shows a log of recent activity.
 
 This page also contains buttons for various actions that can be performed
 on the request as a whole (submitting your review, returning the request to
-the researcher etc.).  In the example below, all buttons are disabled, as
-this request has not yet been reviewed.
+the researcher etc.).  In the example below, all buttons (except "Return request")
+are disabled, as this request has not yet been reviewed.
```
rebkwok, line 33: Is it worth a comment somewhere in the for-researcher's docs on how to create/submit a request to say that if you've made a mistake, you can ask an output-checker to return the request to you in order to make further changes?

rebkwok (approved):
This was much more straightforward than I'd imagined it would be. One minor wording nit and a docs question, nothing blocking

docs/how-tos/review-a-request.md
rw251: That's a very good idea.

rebkwok (approved):

PR #777, opened 2025-02-14
Refactory

bloodearnest:
Refactor test factories for creating users.

This PR has been extracted from work to use a db User class. It tries to draw a clearer distinction between test users in API json format, and test airlock users, which look almost the same atm, but are going to change shortly.

It also enables the linter to find typos, and will make it easier to change the API user format in future if needed.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -29,52 +29,80 @@ from airlock.types import UrlPath
 from airlock.users import User
 
 
-def create_user(
-    username: str = "testuser",
-    workspaces: list[str] | None = None,
-    output_checker: bool = False,
-    last_refresh=None,
-) -> User:
-    """Factory to create a user.
+def create_api_project(project="project", ongoing=True):
+    """Test factory for project details"""
+    return dict(name=project, ongoing=ongoing)
 
-    For ease of use, workspaces is just a flat list of workspace name, which is
-    converted into an appropriate dict with all the right keys.
 
-    If you need to create a more complex workspace dict, you can call
-    create_user_from_dict directly.
-    """
-    if workspaces is None:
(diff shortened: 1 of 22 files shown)


tests/local_db/test_data_access.py
```diff
@@ -112,7 +112,7 @@ def test_delete_file_from_request_bad_state():
     ],
 )
 def test_withdraw_file_from_request_bad_state(status, mock_old_api):
-    author = factories.create_user(username="author", workspaces=["workspace"])
+    author = factories.create_airlock_user(username="author", workspaces=["workspace"])
```
rw251, line 115: Am I right that `workspaces=["workspace"]` is now the default so that omitting it (here and elsewhere) would have the same behaviour? Though perhaps it's more readable in the tests to see it explicitly named.

tests/factories.py
```diff
@@ -29,52 +29,79 @@
...
-    last_refresh=None,
-) -> User:
-    """Factory to create a user.
+):
+    """Test factory to create a user from the Auth API
 
-    For ease of use, workspaces is just a flat list of workspace name, which is
-    converted into an appropriate dict with all the right keys.
+    For convenience, the factory accepts the following values for workspaces:
+     - None:      create a default workspace called "workspace
```
rw251, line 57: A nit so small I can't even see it, but this is missing a closing "

rw251 (dismissed):

tests/local_db/test_data_access.py
bloodearnest: Yeah, the test author chose to be explicit, so I let their decision stand

rw251 (approved):

PR #779, opened 2025-02-14
Update docs to reflect the asynchronous uploads

rebkwok:
Plus a drive-by fix for the screenshots with superfluous tooltips visible


docs/explanation/workflow-and-permissions.md
```diff
@@ -62,10 +67,11 @@ flowchart TD
...
         E -- Updates and responds to questions --> B
-        D -- All files approved ----> F(RELEASED)
+        D -- All files approved ----> F(APPROVED)
         D -- Not approved ----> G(REJECTED)
-        A --> H(WITHDRAWN)
-        E --> H
+        F(APPROVED) -- Files uploaded ----> H(RELEASED)
+        A --> I(WITHDRAWN)
+        E --> I
     end
```
madwort, line 75: this change looks good. whilst looking at it I realise that `RETURNED` currently only leads to `WITHDRAWN` but I can do that nit separately

rebkwok: Returned also leads back to submitted (with "updates and responds to questions"  - i.e. resubmitting). I don't think there's anywhere else it should go.  That's consistent with the auto-generated [state diagram](https://async-release-docs.airlock.pages.dev/reference/request-states/)

madwort: oh sorry, I misread it!! as you were

PR #781, opened 2025-02-14
Update the uploaded files count on the request page

rebkwok:
When a request is approved and has files uploading in the background, we want to be able to keep the user updated as the files upload.  This adds a view that renders just the uploaded file count snippet, and uses htmx to poll it for updates.  If the request isn't in an upload-in-progress state, we don't include the htmx attributes, so we don't poll when nothing will change.

If the request moves out of uploading state (either because it complete, or because it fails and won't re-try anything automatically anymore), we reload the page so we update other relevant elements as well (request status, release button etc).

Note that the audit log on the overview page should technically be also updated along with the file count, but I think it's probably acceptable that it does update until the page is reloaded. If users are staying on the page to wait for the uploads to complete, they probably aren't looking at the audit log until the request either completes or fails, and then it'll be reloaded and updated anyway.

### DIFF
--- a/tests/functional/test_request_pages.py
+++ b/tests/functional/test_request_pages.py
@@ -698,3 +698,89 @@ def test_request_uploaded_files_status(
             expect(release_files_button).to_be_enabled()
         else:
             expect(release_files_button).not_to_be_enabled()
+
+
+def test_request_uploaded_files_counts(
+    live_server,
+    context,
+    page,
+    bll,
+    mock_old_api,
+    mock_notifications,
+):
+    # make a release request in APPROVED status
+    release_request = factories.create_request_at_status(
+        "workspace",
+        status=RequestStatus.APPROVED,
+        files=[
+            factories.request_file(
+                group="group", contents="1", path="file1.txt", approved=True
+            ),
+            factories.request_file(
+                group="group", contents="2", path="file2.txt", approved=True
+            ),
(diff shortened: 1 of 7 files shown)


tests/functional/test_request_pages.py
```diff
@@ -671,3 +671,96 @@ def test_request_uploaded_files_status(
...
+            ),
+            factories.request_file(
+                group="group", contents="2", path="file2.txt", approved=True
+            ),
+            factories.request_file(
+                group="group", contents="3", path="file3.txt", approved=True
+            ),
+        ],
+    )
+    output_checker = login_as_user(
```
rw251, line 700: Have you seen the `create_api_user()` factory function Simon recently added. Example in the tests here:
https://github.com/opensafely-core/airlock/blob/1c371d914380d68d01dfa2dc0bce3052bde69055/tests/functional/test_request_pages.py#L151-L154

rw251 (dismissed):

tests/functional/test_request_pages.py
rebkwok: Oh yes, I have, I just haven't got used to using it yet, and i think I possibly wrote this test (i.e. copied/pasted the setup from another test) before it went it. I'll update this one, thanks.

rw251 (approved):

PR #802, opened 2025-02-17
Refactor login api and separate last_refresh from user data

bloodearnest:
This is in preparation for switching to a db user object.  Separating out last_refresh keeps things cleaner, making the user object purely a representation of the auth API data.

Includes fixture refactoring to make stubbing the auth api easier.

### DIFF
--- a/airlock/login_api.py
+++ b/airlock/login_api.py
@@ -20,31 +20,18 @@ def get_user_data(user: str, token: str):
         return get_user_data_prod(user, token)
 
 
-def get_user_data_prod(user: str, token: str):
-    response = session.post(
-        f"{settings.AIRLOCK_API_ENDPOINT}/releases/authenticate",
-        headers={"Authorization": settings.AIRLOCK_API_TOKEN},
-        json={"user": user, "token": token},
+def get_user_data_prod(username: str, token: str):
+    api_user = auth_api_call(
+        "/releases/authenticate", {"user": username, "token": token}
     )
-    # We don't currently get any more detail about failures than this
-    if response.status_code == requests.codes.forbidden:
-        raise LoginError("Invalid user or token")
-    response.raise_for_status()
-    details = response.json()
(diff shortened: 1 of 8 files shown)


airlock/login_api.py
```diff
@@ -58,7 +56,38 @@ def get_user_data_dev(dev_users_file: Path, user: str, token: str):
...
         raise LoginError("Invalid user or token")
     else:
-        details = dev_users[user]["details"]
-        # ensure that we never try refresh this user with job-server
-        details["last_refresh"] = time.time() + (365 * 24 * 60 * 60)
-        return details
+        return dev_users[user]["details"]
+
+
+def level4_auth_api_call(path, json):
```
rebkwok, line 62: I have a not-very-strong objection to calling this `level4_...`; I know it reflects theview names in the jobserver endpoints that it's calling, but presumably at some point we might have Airlock in a non-TPP backend where it doesn't have a thing called level 4 that it's calling from. Is just `auth_api_call` too easilty confused with this current api? `external_auth_api_call`?

rebkwok (approved):
Just a naming-things comment

airlock/login_api.py
bloodearnest: So, I would expect that there would likely be an entirely different get_user_data/get_user_authz implementation in that case, and we would have a module level abstraction for it, and this function would be internal.

That said, level4_ is not a great name so I will changed it

bloodearnest:
Having now fleshed out a fuller auth implementation using django.contrib.auth, I have realised that there are some awkward problems with storing last_refresh in the session and its interaction with our dev user switch, where we do the unusual thing of POSTing to /login when we are already logged in. This breaks some assumptions about sessions use, as the session is (correctly) cleared in this case.

Also, there was a bit of an inconsistancy/layering issue with saving the state user model and session, so I think I may revert the 3rd commit here, as it turns out not to have been helpful.

PR #813, opened 2025-02-20
Add meta data to requests view

### LINKED ISSUE #756: Add descriptive metadata to 'requests' view
The requests page doesn't provide identifying metadata for users (which aids navigation). It would be helpful to make the following changes: 
- Modify the request names i.e reduce the length of the generated IDs
- Add date created displayed in reverse chronological order
- Display the workspace the request belongs to. 

Current request page: 

Image

Proposed request page: 

Image

Providence-o:
Fixes #756

### DIFF
--- a/assets/templates/_components/list-group/list-group-rich-item.html
+++ b/assets/templates/_components/list-group/list-group-rich-item.html
@@ -23,6 +23,8 @@
         <span class="{{ title_class }}">{{ title }}</span>
         {% endif %}
       </dd>
+
+      
       {% if status_text or custom_status %}
         <dt class="sr-only">Status:</dt>
         <dd class="ml-auto flex-shrink-0">
--- a/airlock/models.py
+++ b/airlock/models.py
@@ -651,6 +651,9 @@ class ReleaseRequest:
     def get_id(self):
         return self.id
 
+    def get_short_id(self):
+        return f"{self.id[:3]}...{self.id[-6:]}"
+
     def get_url(self, relpath=""):
         return reverse(
             "request_view",
--- a/airlock/templates/requests_for_output_checker.html
+++ b/airlock/templates/requests_for_output_checker.html
@@ -11,11 +11,21 @@
     {% #card title="Outstanding requests awaiting review" %}
       {% #list_group id="outstanding-requests" %}
         {% for release_request, request_progress in outstanding_requests %}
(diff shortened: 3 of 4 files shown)


assets/templates/_components/date-time.html
tomodwyer, line 1: This file can now be removed

assets/templates/_components/list-group/list-group-rich-item.html
```diff
@@ -23,6 +23,15 @@
...
       </dd>
+
+      
+      {% if date %}
+        <dt class="sr-only">Date:</dt>
+        <dd class="flex-shrink-0">
+          {% date_time variant="info" text=date %}
+        </dd>
+      {% endif %}
+
```
tomodwyer, line 34: This change can now be reverted

assets/templates/components.yaml
```diff
@@ -8,6 +8,7 @@ components:
   card: "_components/card/card.html"
   card_footer: "_components/card/card-footer.html"
   code: "_components/code.html"
+  date_time: "_components/date-time.html"
```
tomodwyer, line 11: This change can be reverted

rebkwok (dismissed):
The code changes here all look fine to me, and the extra workspace and created at info looks good. There's something that's slightly worrying me about the truncated request though. On the requests and reviews pages, everything there is a request. Do we need the "Request: " prefix? I know the plan is to merge workspaces and requests, so it probably works there, but it maybe isn't needed on the Reviews page?

Do you think we need to give users some advance warning of the truncated request IDs?  They're used to seeing the long ID everywhere at the moment.

Providence-o:
We would definitely need to let users be aware of the truncated ID's so they don't get confused when trying to identify requests/reviews. 

I had figured having the 'request' prefix follow the format on the request details page would be helpful in realising the IDs are truncated.

Providence-o:
We have decided to implement a temporary solution where the requests will be displayed like this: `01J...RCVWF2` Full discussion [here](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1740402057728219)

rebkwok (approved):
This looks good now, thanks. Could you also update the `requests_for_workspace.html` to match the `requests_for_researcher.html` one (as a separate PR, if you want to get this one in first)

PR #814, opened 2025-02-20
Fix upload bug when file content has changed

rebkwok:
The uploader was incorrectly uploading the workspace file content instead of the request file content. This fixes the bug and adds a test for it.

### DIFF
--- a/tests/integration/management/commands/test_run_file_uploader.py
+++ b/tests/integration/management/commands/test_run_file_uploader.py
@@ -153,6 +153,7 @@ def test_run_file_uploader_command(upload_files_stubber, bll):
     assert last_trace.attributes == {
         "release_request": release_request.id,
         "workspace": "workspace",
+        "group": "group",
         "file": "test/file2.txt",
     }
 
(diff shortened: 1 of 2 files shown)


evansd (approved):
Good catch! I totally missed this when reviewing the first time around.

tests/integration/management/commands/test_run_file_uploader.py
```diff
@@ -77,15 +77,17 @@ def test_do_upload_task(upload_files_stubber, bll, freezer):
...
 
     relpath = UrlPath("test/file.txt")
     # modify workspace file content
-    factories.write_workspace_file(release_request.workspace, relpath, contents="changed")
+    factories.write_workspace_file(
+        release_request.workspace, relpath, contents="changed"
+    )
     # refresh workspace
-    workspace = bll.get_workspace(
+    bll.get_workspace(
```
madwort, line 90: can we drop this `get_workspace` call entirely?

rebkwok: We could, but I left it in there after the failing test was passing just in case. I have another PR about to go, so I might remove it there, but leaving it here to avoid the re-approve faff

PR #815, opened 2025-02-20
Retry file uploads forever

### LINKED ISSUE #812: Retry failed file uploads forever
Mostly we expect failed file uploads to be due to some transient error, and to succeed on a retry. We currently retry 3x and then require a user to go an click the button again. It would make more sense to keep retrying, with a delay after 3x (or whatever)
[slack](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1739880018331649) 🧵

rebkwok:
Fixes #812 

Simplifies the file uploading, and removes the need for users to go and press a re-release button, by retrying file uploads until they succeed (or until we manually intervene because there's something more broken than a transient connection error that re-trying would fix).

Instead of trying up to 3x, we now keep track of the time the last attempt was made, and retry again after 60s. For transient errors, this delay might be enough for the next attempt to succeed, whereas retrying without a delay (other than the length of one loop through the file uploader's while loop) would be more likely to hit the same thing again.

I did wonder about keeping a max upload attempts but making it a very big number, just in case we leave something retrying over and over until the retry attempts count gets too big for sqlite, but then I did the maths and realised that at 1 attempt per minute, we've got several billion years until that happens, so we're probably good.


evansd (approved):
I do love a net-negative line count PR! Looks like a good simplification all round.

> we've got several billion years until that happens, so we're probably good

:rofl:

Issue #797, opened 2025-02-14
Better signposting for output checkers when all files are reviewed


rw251:
After reviewing the last file in a request, an alert appears with “Action required. You have reviewed all files. You can now submit your review.” But you stay on the same (file view) page, and there’s no information about how or where to submit the review. We could include a link to the request overview page in the alert so you can easily navigate to the right place.

Though we need to consider that after reviewing the last file they might want to first go to the "group" page to add comments.

Issue #718, opened 2024-12-11
File group view: add the file group name to the top details bar


madwort:
When viewing a filegroup in a request, although the sidebar highlighting changes, we just had a user not realise they were looking at a group instead of the top-level request. Another possible way we could highlight that would be to add the group name to the details above the main content section, comme ça:

<img width="1217" alt="Image" src="https://github.com/user-attachments/assets/18906856-c328-4d87-ad54-9f80965bc038" />

rebkwok:
There's slightly more complexity to this than it might appear (or at least, to adding the group to the header). If you click from one group to another in the tree, the page content (but not the header) is updated via htmx. So it will need an [hx-oob-swap](https://htmx.org/attributes/hx-swap-oob/) or something to update it in the header. Or perhaps we have clicks on groups in the tree reload the whole page and not just the content.

PR #826, opened 2025-02-21
Only show "back to request" link if not already on that page

rw251:
Previously, when viewing a request you would always see:
image
Even if you were on the page it linked to.

Now, if you're on that page already you get:
image


airlock/views/request.py
```diff
@@ -372,6 +372,8 @@ def request_view(request, request_id: str, path: str = ""):
         "include_code": code_url is not None,
         "include_download": not is_author,
         "upload_in_progress": release_request.upload_in_progress(),
+        "is_htmx_request": request.htmx,
```
rebkwok, line 375: We have access to `request` in the template, is there a benefit to making `is_htmx_request` a separate variable? Does the template linter complain?

airlock/templates/file_browser/request/back_link_or_title.html
```diff
@@ -0,0 +1,9 @@
+<div hx-swap-oob="innerHTML:.header__actions">
+  {% if is_request_root %}
+    <h2 class="font-semibold tracking-tight text-slate-900">Request overview</h2>
+  {% else %}
+    <a class="header__link header__nav-link" href="{{ release_request.get_url }}" id="request-home-button">
+      &larr; Request overview
+    </a>
+  {% endif %}
+</div>
```
rebkwok, line 9: There is a [ticket](#718) for adding the group name into the header that will also need an hx-swap-oob when users navigate with the tree. Is it worth thinking about how to do that at the same time? We don't want to end up with lots of bits of oob swaps, so would it make sense to replace the whole header rather than just the title/back link?

rebkwok: I'm also looking at updates to the action-required alert (https://github.com/opensafely-core/airlock/issues/797), which I think also want an oob swap for, so it can have a link to the overview page but not if you're already on that page. So that's basically everything in `templates/file_browser/request/index.html`.

airlock/views/request.py
rw251: Good point. I'll change.

airlock/templates/file_browser/request/back_link_or_title.html
rw251: Good idea. I'll see if I can rework it a bit.

rw251: The whole "header" now updates whenever the file browser contents change via htmx request - but only for the file browser for "requests".

I considered renaming the "content" block to something more meaningful e.g. "header", but that pattern appears in lots of places so I just left it as is.

rebkwok (approved):
Thanks for making those changes - makes my ticket at least 100% easier :slightly_smiling_face:

PR #828, opened 2025-02-21
Better alerts for action required and group name in request header

### LINKED ISSUE #718: File group view: add the file group name to the top details bar

### LINKED ISSUE #797: Better signposting for output checkers when all files are reviewed

rebkwok:
Fixes #718 
Fixes #797 

Include links back to the request overview page in the action required alerts (visible when a user completes their file reviews but hasn't submitted the review yet, or when a request is ready to return/reject/release). For the submit-review, i considered putting a link to actually submit the review in the alert, rather than directing the user back to the overview page, but it requires a form and a post, and would move more logic out to the templates.

Also adds the group into the header, because both of these elements are different if you're on the overview page, and use the same htmx oob to update when the user navigates around via tree links. 

Example overview page: No group in header, no links in alert
Screenshot from 2025-02-21 16-51-41

Example content page (a file, same header for group and folder views): Group in header, alter has link to the overview page.
Screenshot from 2025-02-21 16-51-31

### DIFF
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -362,7 +362,10 @@ def request_view(request, request_id: str, path: str = ""):
                 "Two independent reviews have been submitted. You can now "
             )
         else:
-            request_action_required = f"Two independent reviews have been submitted. go to the <a class='text-oxford-600' href='{release_request.get_url()}'>request overview</a> to "
+            request_action_required = (
+                "Two independent reviews have been submitted. Go to the "
+                f"<a class='text-oxford-600' href='{release_request.get_url()}'>request overview</a> to "
+            )
 
         if release_request.can_be_released():
             request_action_required += "return, reject or release this request."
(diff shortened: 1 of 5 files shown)


airlock/views/request.py
```diff
@@ -343,17 +343,33 @@ def request_view(request, request_id: str, path: str = ""):
...
     elif (
         release_request.status.name == "REVIEWED"
         and permissions.user_can_review_request(request.user, release_request)
     ):
+        if relpath == ROOT_PATH:
+            request_action_required = (
+                "Two independent reviews have been submitted. You can now "
+            )
+        else:
+            request_action_required = f"Two independent reviews have been submitted. go to the <a class='text-oxford-600' href='{release_request.get_url()}'>request overview</a> to "
```
rw251, line 365: go -> Go

rw251 (dismissed):

rw251 (approved):

PR #831, opened 2025-02-24
Add user app with db User model and backfill

bloodearnest:
~This switches airlock to using normal django.contrib.auth and a database User model.~

Update: this now does the first part of this change, creating and populating the User db table. #837 will make the actual switch to using the new table.

This is all contained within a new `users` django app.

This change is designed to be the minimal change to switch. The following are deliberately *not* part of this PR, and are expected to be done separately:

 - does not change the User object API, just the implementation
 - does not add a foreign key, at the db or ORM level, between local_db and the new User model.
 - does not add support for `copiloted_workspaces`, even though the api data contains it
 - does not add support for `fullname`, even though the api data contains it, UI continues to use the username.
 - does not change the BLL API, it still consumes full User models (we may or may not want to change this).


airlock/management/commands/run_file_uploader.py
```diff
@@ -11,15 +11,17 @@
...
 
 
 logger = logging.getLogger(__name__)
 
 # We need a user with the output-checker role to access some
 # bll methods
-system_user = User("system", output_checker=True)
+system_user = User(
+    user_id="system", api_data={"username": "system", "output_checker": True}
+)
```
rebkwok, line 24: Now that this is a db user rather than an ephemeral thing, would it be better to `get_or_create` it within the `handle()` method instead of making it a global (it's only used in one place now anyway)

airlock/management/commands/run_file_uploader.py
```diff
@@ -151,4 +153,4 @@ def get_upload_files_and_update_request_status(release_request):
 
 
 def get_user_for_file(request_file):
-    return User(username=request_file.released_by, output_checker=True)
+    return User.objects.get(pk=request_file.released_by)
```
rebkwok, line 156: ~~If users are only going to be created in the db when they log in, there's a chance this could break things if an upload happens to be in progress when we switch to using the db users, and whoever released the files hasn't yet logged in. The uploader will keep trying forever, so once they do log in, it'll upload, so maybe the risk is not worth worrying about.~~
Nvm, got to the last commit now

tests/functional/conftest.py
```diff
@@ -46,14 +46,23 @@ def login_as_user(live_server, context, user_dict):
     sets the session cookie.
     """
     user = factories.create_airlock_user(**user_dict)
-    session_store = Session.get_session_store_class()()  # type: ignore
-    session_store["user"] = user.to_dict()
-    session_store.save()
+
+    from django.contrib import auth
+    from django.test import RequestFactory
```
rebkwok, line 51: do these imports need to be inside the function?

rebkwok (approved):
LGTM, just one question
It's failed on another functional test, which I'm hoping #832 might prevent

airlock/management/commands/run_file_uploader.py
bloodearnest: Possibly.

It isn't actually persisted here. We instantiate it directly, and we don't save() it. So it doesn't exist in the db.

I doon't think there's any audit log generated for the bll call we need it for, so the username is not even stored anywhere.

However, having it as a global smells wrong.  I will move it into the handle() call, and add a comment about not actually being persisted.

airlock/management/commands/run_file_uploader.py
bloodearnest: It's a good point though, as it could still happen before the manual back fill runs.

I may manually deploy this, just to avoid that.

rebkwok (approved):

airlock/management/commands/run_file_uploader.py
bloodearnest: Thinking about it more, I may split this PR into two. One with the db models/migrations and the backfill, which we can deploy and run, and an another with switching to actually use it.  I think that way we can ensure nothing goes wrong?

bloodearnest: Ok, I have done this - this PR has been rebased and does not perform the switch, and I added a live backfill.

#837 will do the switch, once this is live and users have been populated.

airlock/management/commands/run_file_uploader.py
bloodearnest: The fix for this is now part of #837

tests/functional/conftest.py
bloodearnest: The fix for this is now part of #837

PR #832, opened 2025-02-24
Replace wait_for_timeout with wait for htmx classes

rebkwok:
Playwright docs warn that `wait_for_timeout` is flaky https://playwright.dev/python/docs/api/class-frame#frame-wait-for-timeout

Replace it with a wait for the htmx class that's added and then removed during an htmx operation.
https://htmx.org/docs/#request-operations

Also wait for htmx to be done in the `find_and_click` helper function. Not all of these clicks will involve htmx, but lots of them do. Note that we could wait for `htmx-added` counts of 1 and then 0, which would verify that an htmx swap happened - but it doesn't always pick up the 1 count, even when it's definitely an htmx operation, so I think sometimes it's too fast for playwright; in any case, this means that we can add it in the `find_and_click` function without being too bothered about whether it's htmx or not, we just make sure there's no htmx-swapping still going on.

### DIFF
--- a/tests/functional/test_e2e.py
+++ b/tests/functional/test_e2e.py
@@ -8,8 +8,10 @@ from airlock.enums import RequestStatus
 from airlock.types import UrlPath
 from tests import factories
 
+from .conftest import wait_for_htmx
 
-def find_and_click(locator):
+
+def find_and_click(page, locator):
     """
     Helper function to find a locator element and click on it.
     Asserts that the element is visible before trying to click.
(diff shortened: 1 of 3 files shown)


bloodearnest (approved):

tests/functional/test_e2e.py
```diff
@@ -19,29 +21,25 @@ def find_and_click(locator):
     """
     expect(locator).to_be_visible()
```
rw251, line 22: The wait_for_htmx is a good idea, but I think the existing code was actually redundant. [This](https://playwright.dev/docs/actionability) says that playwright auto waits for a locator to be visible, stable, enabled and receiving events before it attempts the click.

tests/functional/test_e2e.py
```diff
@@ -19,29 +21,25 @@ def find_and_click(locator):
...
     locator.click()
```
rw251, line 23: The wait_for_htmx is a good idea, but I think the existing code was actually redundant. [This](https://playwright.dev/docs/actionability) says that playwright auto waits for a locator to be visible, stable, enabled and receiving events before it attempts the click.

PR #835, opened 2025-02-25
Enabled drop down toggle on project names

### LINKED ISSUE #757: Make 'projects' on workspace view collapsible
Some of our users work on many projects which have many work-spaces. To improve their experience, we are introducing indented collapsible elements in the workspace view. 

Current workspaces page: 

Image

Proposed workspaces page:

Image

Providence-o:
Fixes #757

### DIFF
--- a/airlock/templates/workspaces.html
+++ b/airlock/templates/workspaces.html
@@ -5,19 +5,30 @@
 {% block content %}
   <div class="flex flex-col gap-4 max-w-3xl">
     {% airlock_header title="Workspaces for "|add:request.user.username %}
-
     {% for project, workspaces in projects.items %}
-      {% #card title=project.display_name %}
-        {% #list_group id="workspaces" %}
-          {% for workspace in workspaces %}
-            {% #list_group_item href=workspace.get_url %}{{ workspace.display_name }}{% /list_group_item %}
-          {% endfor %}
-        {% /list_group %}
-      {% /card %}
-    {% empty %}
-      {% #card title="Workspaces" container=True %}
-        {% list_group_empty title="No workspaces available" description="You do not have access to any workspaces" %}
-      {% /card %}
-    {% endfor %}
-  </div>
-{% endblock content %}
+      {% #card %}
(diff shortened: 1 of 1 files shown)


airlock/templates/workspaces.html
```diff
@@ -5,19 +5,30 @@
...
-      {% #card title="Workspaces" container=True %}
-        {% list_group_empty title="No workspaces available" description="You do not have access to any workspaces" %}
-      {% /card %}
-    {% endfor %}
-  </div>
-{% endblock content %}
+      {% #card %}
+        <details class="group" open>
+          <summary class="list-none more-dropdown__btn">
+            {% fragment as custom_button %}<span class="opacity-75 ml-1"><img height="16" width="16" class="icon inline-block group-open:hidden" src="/static/icons/keyboard_arrow_down.svg" alt="">
```
rebkwok, line 12: This is pedantic (sorry), but I think it would be better to increase the size of the up/down arrows so they're more in line with the size of the project name text.  Currently they're the same size as they are in the nav bar dropdowns, and I almost missed them altogether.
height/width =30 is about right, I think (you'll need to remove the `icon` class because it overrides the height/width)

image

airlock/templates/workspaces.html
```diff
@@ -5,19 +5,30 @@
...
-    {% empty %}
-      {% #card title="Workspaces" container=True %}
-        {% list_group_empty title="No workspaces available" description="You do not have access to any workspaces" %}
-      {% /card %}
-    {% endfor %}
-  </div>
-{% endblock content %}
+      {% #card %}
+        <details class="group" open>
+          <summary class="list-none more-dropdown__btn">
```
rebkwok, line 11: We don't want the `more-dropdown__btn` here, I think. It has extra button styling for the More button on file views and I don't think any of them are being applied here.

Providence-o: Thanks for catching this! It wasn't meant to be in the code

PR #837, opened 2025-02-25
Switch to using new db models and django.contrib.auth

bloodearnest:
The previous PR has created and populated the new User table, so we are now safe to switch over.

This is all contained within a new `users` django app.

This change is designed to be the minimal change to switch. The following are deliberately *not* part of this PR, and are expected to be done separately:

 - does not change the User object API, just the implementation
 - does not add a foreign key, at the db or ORM level, between local_db and the new User model.
 - does not add support for `copiloted_workspaces`, even though the api data contains it
 - does not add support for `fullname`, even though the api data contains it, UI continues to use the username.
 - does not change the BLL API, it still consumes full User models (we may or may not want to change this).


airlock/settings.py
```diff
@@ -101,14 +102,14 @@ def get_env_var(name):
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
-    # "django.contrib.auth.middleware.AuthenticationMiddleware",
+    "django.contrib.auth.middleware.AuthenticationMiddleware",
+    # "django.contrib.auth.middleware.LoginRequiredMiddleware",
```
rebkwok, line 106: Should we just remove the django LoginRequiredMiddleware? We're not planning to use it in future are we?

rebkwok (approved):

airlock/settings.py
bloodearnest: Yeah, I though we might do.

I will remove in the next PR though, to avoid changing this and having to re-approve, if that's ok

rebkwok (approved):

PR #838, opened 2025-02-25
Wait for iframe content to be present in functional tests

rebkwok:
I don't have much confidence in this, but checking the iframe is the most frequently seen of the intermittent functional test timeouts.  `hover()` waits for the element to be [visible, stable and receiving events](https://playwright.dev/docs/actionability), so I wonder if calling `hover()` on the iframe container before getting its text content will help. If not, it might be informative to see whether it times out at the `hover()` or still on checking visibility of the text content.

### DIFF
--- a/tests/functional/test_e2e.py
+++ b/tests/functional/test_e2e.py
@@ -132,6 +132,9 @@ def test_e2e_release_files(
 
     # Check the expected iframe content is visible
     iframe_locator = page.locator("#content-iframe")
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Worth a go!

rebkwok:
I'm going to close this one as I haven't seen the failure in a while, and I think other changes recently are more likely to have helped with it than this

PR #841, opened 2025-02-26
Allow files to be moved between groups

### LINKED ISSUE #340: Allow files to be moved between groups
This can be worked around by withdrawing a file and re-adding it.

madwort:
* add ability to move a file from one group in a request to another
* fixes #340 
* use existing withdraw/re-add functions
* creates new template for the modal based on the add file modal as it had a few differences, although I've since reverted some of those changes & if we use this to allow files to be switched between output/supporting then they may align more (& then need refactoring to be reused)
Screenshot from 2025-02-28 15-53-21
Screenshot from 2025-02-28 15-54-04
Screenshot from 2025-02-28 15-54-22


airlock/views/request.py
```diff
@@ -600,23 +612,66 @@
...
+                initial={"next_url": f"/requests/view/{release_request.id}/"},
+            )
 
-        display_multiple_messages(request, errors, "error")
-        display_multiple_messages(request, successes, "success")
+            filetype_formset = FileFormSet(
+                initial=[{"file": f} for f in files_to_add],
+            )
+            return TemplateResponse(
+                request,
```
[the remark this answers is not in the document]
madwort: hmm, it does arguably have a point. if we change it here then I think there are possibly a couple of other places in the codebase where we should change it to match.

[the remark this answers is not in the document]
madwort: I think this is sanitised by the template, so this is already safe

airlock/templates/file_browser/request/file.html
```diff
@@ -34,6 +34,29 @@
       </form>
     {% endif %}
 
+    {% if content_buttons.move_file_button %}
```
rebkwok, line 37: Should this be `content_buttons.move_file_button.show`?

airlock/templates/file_browser/request/file.html
```diff
@@ -34,6 +34,29 @@
...
+      {% if content_buttons.move_file_button.disabled %}
+        {% #button type="button" disabled=True tooltip=content_buttons.move_file_button.tooltip id="update-file-modal-button" %}
```
rebkwok, line 39: `id="change-filegroup-modal-button"`?

airlock/templates/file_browser/request/file.html
```diff
@@ -34,6 +34,29 @@
...
+          Move file to new group
+        {% /button %}
+      {% else %}
+        <form
+          method="POST"
+          hx-post="{{ content_buttons.move_file_button.url }}"
+          hx-target="#multiselect_modal"
+          hx-swap="outerHtml"
+        >
+          {% #button type="submit" name="action" value="update_files" variant="success" id="update-file-modal-button" %}
```
rebkwok, line 49: id="change-filegroup-modal-button"?

airlock/templates/change_file_group.html
```diff
@@ -0,0 +1,44 @@
+{# hide default button, as we are going to autoshow the modal anyway #}
+{% fragment as button %}
+{% endfragment %}
+<style>
+  div.filetype-radio div { display: inline; }
+  div.filetype-radio label { display: inline; }
+</style>
```
rebkwok, line 7: I don't think these are used (but might make sense to leave them, if we're going to reuse this template for changing filetype)

airlock/templates/file_browser/request/file.html
```diff
@@ -34,6 +34,29 @@
       </form>
     {% endif %}
 
+    {% if content_buttons.move_file_button %}
+      {% if content_buttons.move_file_button.disabled %}
+        {% #button type="button" disabled=True tooltip=content_buttons.move_file_button.tooltip id="update-file-modal-button" %}
+          Move file to new group
```
rebkwok, line 40: Suggest "Change filegroup"? When I first looked at the screenshots, this suggested to me that it was only for creating a new group

airlock/views/request.py
```diff
@@ -604,26 +615,72 @@ def request_multiselect(request, request_id: str):
...
         display_form_errors(request, multiform.errors)
-    else:
-        action = multiform.cleaned_data["action"]
-        if action != "withdraw_files":
-            raise Http404(f"Invalid action {action}")
+        url = multiform.cleaned_data["next_url"]
+        return HttpResponse(headers={"HX-Redirect": url})
+
+    action = multiform.cleaned_data["action"]
+    match action:
```
rebkwok, line 622: This might be nicer split out into functions for withdraw and update, like we do in the workspace version

airlock/views/request.py
```diff
@@ -717,6 +717,55 @@ def file_approve(request, request_id, path: str):
...
+        return form.cleaned_data["next_url"]
+
+    # default redirect in case of error
+    return release_request.get_url()
+
+
+@instrument(func_attributes={"release_request": "request_id"})
+@require_http_methods(["POST"])
+def file_move_group(request, request_id):
+    # TODO: finish this
```
rebkwok, line 731: leftover TODO?

airlock/views/request.py
madwort: good idea, will do

airlock/permissions.py
```diff
@@ -193,18 +193,28 @@ def user_can_add_file_to_request(
...
 
 
 def user_can_replace_file_in_request(
-    user: User, request: "ReleaseRequest", workspace: "Workspace", relpath: UrlPath
+    user: User,
+    request: "ReleaseRequest",
+    workspace: "Workspace",
+    relpath: UrlPath,
+    filegroup: str | None = None
 ):  # pragma: no cover; not currently used
```
inglesp, line 213: Is this still not used?

inglesp (dismissed):
Have only reviewed @rebkwok's changes.

inglesp (approved):

PR #842, opened 2025-02-26
Some more functional tests rework and a tracing workflow

rebkwok:
This reworks some playwright locators, especially those around the flakiest bits (generally the tree clicks and iframe content), by using the locators that are identified in the [Playwright Inspector](https://playwright.dev/python/docs/running-tests#debugging-tests) when running the tests with `PWDEBUG=1`

Also adds a GHA workflow to run the functional tests with tracing and upload the trace as an artifact if the test fails.  The workflow will run every hour, so if it fails, we can download the trace for the failing test(s) and inspect the page state at the time of the failure. Hopefully this will give us some insights into how we might fix the flakiness.

The trace is found with the workflow run, [this is an example](https://github.com/opensafely-core/airlock/actions/runs/13544691316) from a test run where it uploaded the traces even if tests passed (i.e. at the second-to-last commit of this PR). 

Also some more developer docs around debugging and tracing.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -171,6 +171,8 @@ or to run a single test, run e.g. `just test tests/unit/test_urls.py::test_urls`
 
 #### Debugging
 
+##### Headed mode
+
 Functional tests run headless by default. To see what's going on, they
 can be run in headed mode. The following command will run just the
 functional tests, in headed mode, slowed down by 500ms. See the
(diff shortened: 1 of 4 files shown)


DEVELOPERS.md
```diff
@@ -187,6 +189,34 @@ pytest drop into the debugger using the `--pdb` argument:
...
+for each test with:
+
+```
+just test ... --tracing on
+```
+
+This will record the trace and place it into the file named trace.zip in the `test-results`
+directory.
+
+You can load the trace using Playwrright's trace viewer, and see the state of the page
```
inglesp, line 214:
```suggestion
You can load the trace using Playwright's trace viewer, and see the state of the page
```

inglesp (dismissed):
NIce.  Where do these CI failures get reported?

rebkwok:
> NIce. Where do these CI failures get reported?

Currently we'd just need to check the workflows; they'll get reported as part of the bennettbot checks, but only the last one.  I'll see if I can get the slack notifcations to work and post into the team channel

rebkwok:
Have added a slack notification; output from test run [here](https://bennettoxford.slack.com/archives/CGF9TKZLG/p1740584183451949)

(The link goes to the relevant test run, but there's no artifact to download there because the tests didn't actually fail. It also only posted to tech-noise for testing, it'll actually post to team-rap for real test failures)

inglesp (approved):

PR #850, opened 2025-02-28
Switch from username string to User object in BLL/DAL

bloodearnest:
Basically, wherever we used a username string, we now have a db User object.
This is a consistent and more future-proof API, as it allows us natural access to full user data whenever we need it, moves us away from using username (which can change), and allows us to easily change the User model without altering the API again.

The local_db DAL implementation still just persists the user_id, however (which is currently just username, but we expect in future that it could be a different (e.g. Github) id, so we can support renames.

However, doing this when combined with our to_dict/from_dict serialisation barrier between the BLL and the DAL, creates a slightly strange interaction.  Everything in those dicts so far had been native python types, in part because it was kind of simulating a future JSON API with job-server.

But the BLL models need User objects.  So, if the DAL uses user_id to look upthe User, and passes it in the dict, it's not doing its intended purpose of simulating a JSON interface.  So, for this PR, I have made the BLL object's from_dict method responsible for converting a user_id passed from the DAL layer into a full User object. This is also a bit strange, as from_dict is now doing a DB query, which is not ideal. But I think its better than the DAL doing it.

I discussed this design change in detail with Dave, and we came to the conclusion that it is likely that the to_dict/from_dict interface is hoisted too high up, and in future, we will probably want the DAL to both receive and return full BLL objects directly, rather than dicts. This avoids unnecessarily exposing the details of a JSON dict to the BLL.  But that is a future piece of work.

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -68,13 +68,13 @@ class DataAccessLayerProtocol(Protocol):
     def create_release_request(
         self,
         workspace: str,
-        author: str,
+        author: User,
         status: RequestStatus,
         audit: AuditEvent,
     ):
         raise NotImplementedError()
 
-    def get_active_requests_for_workspace_by_user(self, workspace: str, username: str):
+    def get_active_requests_for_workspace_by_user(self, workspace: str, user: User):
         raise NotImplementedError()
 
     def get_requests_for_workspace(self, workspace: str):
@@ -83,7 +83,7 @@ class DataAccessLayerProtocol(Protocol):
     def get_released_files_for_workspace(self, workspace: str):
         raise NotImplementedError()
 
-    def get_requests_authored_by_user(self, username: str):
+    def get_requests_authored_by_user(self, user: User):
         raise NotImplementedError()
 
     def get_requests_by_status(self, *states: RequestStatus):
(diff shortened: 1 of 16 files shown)


airlock/business_logic.py
```diff
@@ -485,7 +482,7 @@ def check_status(
                     f"only an output checker can set status to {to_status.name}"
                 )
 
-            if user.username == release_request.author:
+            if user == release_request.author:
```
evansd, line 485: Do we need to define an `__eq__` method for `User` so that they compare by ID or do Django models naturally do the right thing here?

airlock/business_logic.py
```diff
@@ -868,9 +866,7 @@ def register_file_upload(
             user=user,
             path=relpath,
         )
-        self._dal.register_file_upload(
-            release_request.id, relpath, user.username, audit
-        )
+        self._dal.register_file_upload(release_request.id, relpath, audit)
```
evansd, line 869: Because this was one of the few methods where the number of arguments had changed I spent ages failing to understand how this code could possibly work and then I realised I was mixing up the DAL and BLL methods!

evansd (approved):
Thanks for slogging through this! Can't have been much fun, but really good to have all this on a solid footing now.

Just one question (mostly for my own info) on how equality between user objects works.

airlock/business_logic.py
bloodearnest: Yeah, I think I just caught an unneeded parameter in a DAL call that could be removed, because we already construct the AuditEvent in the BLL, and that's all this method needs.

In hindsight, I probably could have pulled out the removal of the username parameter from this DAL method in a separate commit

PR #855, opened 2025-03-03
Include copiloted workspaces in navigation

### LINKED ISSUE #278: Include co-piloted workspaces in requests index

rebkwok:
Fixes #278 

Adds a new page for copiloted workspaces, and a new nav item (only visible if the user has copiloted workspaces to view)

image

If a user doesn't have any copiloted workspaces, they don't see the nav option, but going to the page doesn't error:
image

Navigation of workspace files from the copiloted index page is as normal.


docs/screenshots/copiloted_workspace_file_add.png
bloodearnest, line 1: unfortunate hover position on this screen shot?

bloodearnest (approved):
This looks great, thanks.

I wonder if it is worth making the `create_*_user` functions have keyword only arguments?

That should help in future if we add a new parameter again?

docs/screenshots/copiloted_workspace_file_add.png
rebkwok: No, this one is deliberate

bloodearnest (approved):

PR #856, opened 2025-03-03
Add metadata and filter to 'all requests in workspace' view 

### LINKED ISSUE #836: Add descriptive metadata to 'workspaces' view
The workspaces view (which displays all the requests within a workspace) doesn't provide identifying metadata for users (which aids navigation). It would be helpful to make the following changes:

- Include the request names
- Add date created displayed in reverse chronological order
- Display the workspace the request belongs to
- Format the request status 

Current page: 

Image

The modified page should match this formatting:

Image

Providence-o:
Fixes #836 

This PR also fixes a line of code in the test for viewing all the requests in a workspace; changing `airlock_client.post` to `airlock_client.get`


airlock/views/request.py
```diff
@@ -629,13 +629,38 @@ def requests_for_workspace(request, workspace_name: str):
...
+        "PENDING",
+        "SUBMITTED",
+        "WITHDRAWN",
+        "APPROVED",
+        "REJECTED",
+        "PARTIALLY_REVIEWED",
+        "REVIEWED",
+        "RETURNED",
+        "RELEASED",
+    ]
```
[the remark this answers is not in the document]
rebkwok: Or `(status, status.description().title())` if you want the labels to be lower case

airlock/views/request.py
```diff
@@ -629,13 +629,38 @@ def requests_for_workspace(request, workspace_name: str):
...
+
+    filter_url = request.path
 
     return TemplateResponse(
         request,
         "requests_for_workspace.html",
         {
             "workspace": workspace_name,
             "requests_for_workspace": requests_for_workspace,
+            "request_filter": request_filter,
```
rebkwok, line 725: If you change to use the status description, you'll also need to change this so it matches the filter labels; you can retrieve the enum from the name with `RequestStatus[request_filter]` and call `description()` on that

airlock/templates/requests_for_workspace.html
```diff
@@ -4,10 +4,61 @@
...
     {% #list_group id="requests-workspace" %}
-      {% for request in requests_for_workspace %}
-        {% #list_group_item href=request.get_url %}{{ request.workspace }} ({{ request.author }}): {{ request.status.name }} {% /list_group_item %}
+      {% for request in requests_for_workspace|dictsortreversed:"created_at" %}
+        {% fragment as title %}{{ request.get_short_id }} by {{ request.author }}{% endfragment %}
+        {% fragment as status %}{% pill variant="info" text=request.status.name %}{% endfragment %}
+        {% #list_group_rich_item title=title url=request.get_url custom_status=status %}
+          <div class="flex flex-col gap-1 text-xs text-slate-600 sm:flex-row">
+            <div class="flex flex-row gap-1 w-full">
+              <dt>Workspace:</dt>
```
rebkwok, line 47: Do we need the workspace on this page? It's all request for one workspace, so you've already got the workspace in the page header.

tests/integration/views/test_request.py
```diff
@@ -1149,6 +1149,32 @@ def test_requests_for_workspace(airlock_client):
...
+
+    factories.create_request_at_status(
+        "test1",
+        author=author2,
+        status=RequestStatus.PENDING,
+    )
+
+    response = airlock_client.get("/requests/workspace/test1?status=APPROVED")
+
+    response.render()
```
rebkwok, line 1177: You shouldn't need to call `render()` explicitly if you're testing with `rendered_content`.

I'd suggest testing the context instead of the string content; i.e. test that `response.context_data["requests_for_workspace"]` contains the requests you expect - that'll be more likely to continue to pass if we change some of the template format later. At the moment you're asserting that "APPROVED" is in the response content, but I think it always will be. If the filter didn't work and returned all requests, including the pending one, this would still match. If it returned no requests, I think it would also still match, because "APPROVED" would be found in the dropdown options.

rebkwok (dismissed):
Thanks, LGTM.
One small improvement you could make is to limit the filter drop down to only statuses that have matching requests. So if there are only requests in pending and released requests for a particular workspace, we don't show the full list of possible statuses in the dropdown.

airlock/views/request.py
```diff
@@ -700,7 +700,15 @@ def requests_for_workspace(request, workspace_name: str):
...
-    status_list = [(status.name, status.description()) for status in RequestStatus]
+    status_choice = [(status.name, status.description()) for status in RequestStatus]
+
+    # limit filter to statuses with matching requests
+    status_list = {
+        status_name: status_description
+        for status_name, status_description in status_choice
+        for r in requests_for_workspace
+        if r.status.name == status_name
+    }
```
rebkwok, line 711: nit: you could munge `status_choice` into `status_list` with
```
status_list = {
        status.name: status.description()
        for status in RequestStatus
        for r in requests_for_workspace
        if r.status == status
    }
```
v minor nit: it's not a list anymore; maybe rename to the variable to `status_choices`?

tests/functional/test_e2e.py
```diff
@@ -633,3 +633,69 @@ def test_e2e_withdraw_request(page, live_server, dev_users):
...
+
+    # We now have a "View all release requests for workspace" button
+    find_and_click(page, page.locator("#requests-workspace-button"))
+
+    # Clicking on it takes us to the list of all requests in that workspace
+    expect(page).to_have_url(live_server.url + "/requests/workspace/test-workspace")
+    expect(page.locator("body")).to_contain_text("All requests in Workspace")
+
+    # Find the filter button, and verify the statuses there
+    # find_and_click(page, page.locator("#id-status"))
```
rebkwok, line 685: should this be commented out?

airlock/views/request.py
rebkwok: What do you think the filter behaviour should be when the list is filtered?  We have the clear filter that resets it, but should the filter still show the available options?  It's slightly surprising that when I filter to just pending, the dropdown now only shows pending as an option. You could make it show all available options by building the status choices with the version of `requests_for_workspace` before applying any filter.

rebkwok (approved):
Thanks for the comprehensive e2e test and detailed comments, really helpful for reviewing. 
A couple of minor nits, nothing blocking. 
I do have a question about what we want the filter to show when there's a filter applied - quite happy to leave that till a later PR if you want to get this one in now

airlock/views/request.py
Providence-o: Yes. I agree the filter should show all available options, so there won't be a need to clear the selection first before making a different choice

tests/functional/test_e2e.py
Providence-o: Yes and it's not meant to be there at all

airlock/views/request.py
```diff
@@ -693,21 +693,20 @@ def requests_for_workspace(request, workspace_name: str):
     requests_for_workspace = bll.get_requests_for_workspace(
         workspace_name, request.user
     )
+    requests_for_workspace_unfiltered = requests_for_workspace
```
rebkwok, line 696: Tiny nit: this line is unnecessary if you build the `status_choice` dict here, before you do the `if request_filter` block

rebkwok (approved):

PR #858, opened 2025-03-03
Use user fullnames on UI

### LINKED ISSUE #443: Show users' real names rather than GH handles
We currently just use the github username, which we know from job server is not always the best identifier, we probably want full name.

We can get that from the login API, but we only have that data for the current user. So to show the full name of *other* users, we'd probably need a formal persistent User model, to save that full user name somewhere to show. Unless we want to make an API call to job-server every time we need an full user name?

This would also allow us to have an airlock user page in future, which could show the audit log for a specific user, perhaps

bloodearnest:
The username is available as title tooltip.

This is achevied by implementing new `airlock_user` component. It's very simple, but we may want to do more in future (e.g. link to an view for that user, or provided a better toolip).

Fixes #443 at last

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -33,6 +33,10 @@ class User(AbstractBaseUser):
         return self.api_data["username"]
 
     @property
+    def fullname(self):
+        return self.api_data.get("fullname", self.username)
+
+    @property
     def workspaces(self):
         return self.api_data.get("workspaces", {})
 
--- a/airlock/templates/_components/header/request/header.html
+++ b/airlock/templates/_components/header/request/header.html
@@ -23,7 +23,7 @@
           <span class="sr-only">User</span>
           <img class="icon" src="/static/icons/person.svg" alt="">
         </dt>
-        <dd>{{ release_request.author }}</dd>
+        <dd>{% airlock_user user=release_request.author %}</dd>
       </div>
       <div>
         <dt>
--- a/airlock/templates/_components/user.html
+++ b/airlock/templates/_components/user.html
@@ -1,0 +1,1 @@
+<span class="user" title="{{ user.username }}">{{ user.fullname }}</span>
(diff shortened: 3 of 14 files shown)


users/models.py
```diff
@@ -32,6 +32,10 @@ def __str__(self):
     def username(self):
         return self.api_data["username"]
 
+    @property
+    def fullname(self):
+        return self.api_data.get("fullname", self.username)
```
inglesp, line 37: Why would `api_data` ever be missing `"fullname"`?

inglesp (approved):
Neat.

users/models.py
bloodearnest: In theory, it shouldn't. But it comes over the API from job-server, and login updates the db.  And we don't have any validation for this atm, so it could just get removed. Also, I am not 100% sure if job-server guarantees the users has set a fullname or not. 

Additionally, we currently use `api_data.get(,,, {})` for workspaces and the recently added 

So, if it didn't have it for any reason, not having a fallback would break any page that tries to render this user, e.g. activity in workspace, request, or group pages.

I know you prefer to shout loudly in these scenarios, whereas, if the failure UI or otherwise is not critical, I prefer to be defensive and provide fallbacks.  I think my intuitions here have been tuned on much larger scale systems with more stringent uptime goals.

Assuming we can clarify that job-server always provides a fullname, then we may want to break if `"fullname"` is not present in `api_data`.  However, in that case I think we would want the failure to occur much earlier, when we receive the data from the job-server API. We could validate it on login, and refuse the login if not correct. I don't think we can do a partial update of `api_data`, as we rely on it for permissions. We'd also need to validate the regular refresh of data from API, not 100% sure what the correct action should be if we get and invalid response there, again security concerns rear their head.

This API validation works is probably worth doing, I think? I can try add it to this PR, or we can file a ticket?

inglesp: Agree that we want failure to occur as early as possible.  But happy for that to come later.

inglesp (approved):

PR #861, opened 2025-03-03
Add row numbers and sticky headers to tables

### LINKED ISSUE #713: Improvements for table view
A couple of suggestions for making reviewing of tabular outputs easier:

- Row numbers. This allows me to see the total number of rows easily and I can reference a specific row if needed
- Sticky header. For large tabular outputs it can be easy to forget what you're looking at

evansd:
I gave up trying to figure out how to do the CSS the proper way and so just added a style element to the head.

Preview:
Screenshot from 2025-03-04 11-30-27

Closes #713

### DIFF
--- a/airlock/templates/file_browser/file_content/csv.html
+++ b/airlock/templates/file_browser/file_content/csv.html
@@ -12,6 +12,12 @@
     {% vite_hmr_client %}
     {% vite_asset "assets/src/scripts/main.js" %}
     {% vite_asset "assets/src/scripts/datatable.js" %}
+    <style>
+      #airlock-table td.datatable-row-number {
+        text-align: right;
+        color: rgb(186 189 197);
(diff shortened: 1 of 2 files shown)


airlock/templates/file_browser/file_content/csv.html
```diff
@@ -21,6 +27,7 @@
         {% #datatable column_filter searchable sortable %}
           <thead>
             <tr>
+              <th class="datatable-row-number" data-searchable="false" data-sortable="false"></th>
```
inglesp, line 30: Do we definitely not want this column to be sortable?  I've found it annoying elsewhere when I've not been able to get a table back into its initial state after sorting it.

inglesp (dismissed):

inglesp (approved):

Issue #863, opened 2025-03-04
Validate job-server API data


bloodearnest:
Currently, we do not validate the contents of the JSON response from job-server, in users/login_api.py.  We just stick it in the db.  This means that its possible keys might be not be present in the JSON, leading us to [code defensively](https://github.com/opensafely-core/airlock/blob/main/users/models.py#L36).

It would probably be better to validate and error early, and then our code can make stronger assumptions about the contents of api_data.  In a special case its possible that [`fullname` might be `""`](https://github.com/opensafely-core/job-server/blob/main/jobserver/models/user.py#L92), and we'd need to provide a fallback anyway, but based on value of "fullname", not presence of the key.

However, this means we need to decide what to do when validation fails.  We use the login api in two places, initial login, and then the periodic refresh, and the failure behaviour may want to be different in those two cases.  We can potentially fall back to any pre-existing data in the db, however, there are security implications there, privileges may have been removed in job-server, and we want to enforce that.

evansd:
There's a third option between "blow up loudly" and "silently handle the error" which would be to do our best to deal with the missing data but alert (somehow) ourselves to the issue.

In non-locked-down contexts I've often used Sentry's APIs for this but maybe there's something we can rig together with Honeycomb to get the same effect?

I wonder if the best option would be to have the login API handle missing data and both alert and generate reasonable fallback values if it has to. Then the rest of the code is allowed to assume that all the values are populated.

bloodearnest:
Production users do have empty fullname values currently:  https://bennettoxford.slack.com/archives/C069SADHP1Q/p1741081304419789

job-server issue to fix: https://github.com/opensafely-core/job-server/issues/4699

bloodearnest:
Yes, we can add whatever we like to otel telemetry, and add a trigger on it to be alerted, w/o failing the request.

I think we can fall back safely for some things, e.g. fullname. I don't think we can fallback for user_id/username?

For security, I think the fallback for workspaces/copiloted_workspaces/output_checker might need to be`{}/{}/False` respectively. At which point, the user cannot interact with airlock in meaningful way, so we may want to fail the login if we have to fall back too many things?

evansd:
> so we may want to fail the login if we have to fall back too many things?

Yeah, definitely. I'm not saying that the login should never fail; just that (as I think you originally suggested) it would be better to not fail on stuff where we can reasonably cope with it being missing.

PR #864, opened 2025-03-04
Better fullname fallback

bloodearnest:
A user's `fullname` in job-server can be an empty string, which is not ideal. This changes the fallback to username to trigger on both the absence of a `fullname` key *or* its value being Falsey, i.e. `""`

Partially related to #863

### DIFF
--- a/tests/users/test_models.py
+++ b/tests/users/test_models.py
@@ -1,0 +1,11 @@
+from tests import factories
+from users.models import User
+
+
+def test_user_model_fullname_fallback():
+    api_data = factories.create_api_user(username="username", fullname="fullname")
+    assert User.from_api_data(api_data).fullname == "fullname"
+    api_data["fullname"] = ""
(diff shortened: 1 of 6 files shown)


tests/users/test_models.py
```diff
@@ -0,0 +1,10 @@
+from tests import factories
+from users.models import User
+
+
+def test_user_model_fullname_fallback():
+    api_data = factories.create_api_user(username="username")
+    api_data["fullname"] = ""
+    User.from_api_data(api_data).fullname == "username"
+    api_data.pop("fullname")
+    User.from_api_data(api_data).fullname == "username"
```
rw251, line 10: Is this some pytest magic, or is this just missing the `assert`s?

bloodearnest: Ok, so in going bacvk to the fix the missing asserts, I added an extra one, which uncovered a whole other thing, which is fixed in a separate commit.

rw251 (approved):

PR #866, opened 2025-03-04
Allow filetype to be changed

### LINKED ISSUE #354: Allow author to change filetype more easily
For example, from supporting to output or vice versa.

Currently, this is possibly during PENDING state via withdraw/re-add, but I don't think it possible once the request has been submitted.

As well as being able to fix a mistake, I think would allow for a workflow where a file is switched to supporting file, rather than withdrawn, which I suspect is likely a useful option.

rebkwok:
Fixes #354 
Uses the same mechanism as changing file group to also allow changing file type.

The modal now looks like this:
image

(The filename display could be improved, but that's out of scope for this ticket. It's using the same template as the add-files modal now, so the same applies there)

Also improves the redirecting after changing group/type so it uses the input from the modal trigger - when/if we add this to the multiselect, it will redirect to the page it was launched from, either file or directory. If the group is changed, it redirects to the file/directory within the new group.


airlock/views/request.py
```diff
@@ -650,40 +651,53 @@ def multiselect_withdraw_files(request, multiform, release_request):
...
         workspace = bll.get_workspace(release_request.workspace, request.user)
-        relpath = release_request.get_request_file_from_urlpath(f).relpath
-        if permissions.user_can_change_request_file_group(
-            request.user, release_request, workspace, relpath
+        request_file = release_request.get_request_file_from_urlpath(f)
+        if filegroup is None:
+            filegroup = request_file.group
+        else:
+            # We should always be updating files from the same group
+            assert filegroup == request_file.group
```
inglesp, line 665: Would it be cleaner to have `filegroup = multiform.cleaned_data["selected"][0].group` outside the loop?  You then don't need to wrap the assertion in an `if` branch (which is something I'm surprised coverage hasn't moaned about).

inglesp (dismissed):

airlock/views/request.py
rebkwok: I didn't do this before, because you can't do `multiform.cleaned_data["selected"][0].group` directly, so it would mean repeating the `request_file = release_request.get_request_file_from_urlpath()` call.

It could instead set the filegroup on the first iteration and the assert the current iteration's request file's group matches. Is that better?
```
    for i, f in enumerate(multiform.cleaned_data["selected"]):
        workspace = bll.get_workspace(release_request.workspace, request.user)
        request_file = release_request.get_request_file_from_urlpath(f)
        if i == 0:
            filegroup = request_file.group
        # We should always be updating files from the same group
        assert request_file.group == filegroup
```

inglesp: Fine.

inglesp (approved):

PR #877, opened 2025-03-06
Show a different confirmation message for returning an approved request

### LINKED ISSUE #722: Fully approved requests can be sent back to the researcher without releasing anything
Once review requests are fully approved, as well as being able to release the files, output checkers also have the option to send the request back to the researcher. This is easily misunderstood  as being the correct button to press. If it is, then the files cannot be released until, presumably, the researcher has sent the request back (without any changes). [This request](https://github.com/ebmdatalab/opensafely-output-review/issues/903) is an example of a request where this happened.

I suggest removing the return request button once all outputs are approved. Also consider relabelling the release outputs button as something like "Release outputs and complete request"

madwort:
* fixes #722 
* if all files are approved, show a new message with a warning. if any files are not approved, show the current message.
Screenshot from 2025-03-07 10-45-19

### DIFF
--- a/airlock/views/request.py
+++ b/airlock/views/request.py
@@ -163,7 +163,7 @@ def _get_request_button_context(user, release_request):
             if not return_btn.disabled:
                 return_btn.tooltip = "Return request for changes/clarification"
                 if release_request.all_files_approved():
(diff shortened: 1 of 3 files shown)


madwort:
Oh, the documentation screenshots may need to be updated, just checking that

LFISHER7:
One small suggestion for the text - could include something like "If you want to release the files, use the release files button."

Otherwise looks good, thanks

Providence-o (dismissed):

Providence-o (approved):

PR #878, opened 2025-03-06
Enforce comments when changes are requested

### LINKED ISSUE #726: Enforce reviewer comments for file groups containing files with changes requested
You don't currently have to write a reviewer comment for a file group that contains a file you have requested changes for. This means a request can be returned with changes required, but no detail on what changes.

Reviewers should each have to write a comment before they submit their request in these cases. There should also be a public comment before the request is returned.

See [this Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1734450147206509) for more detail.

rebkwok:
Fixes #726 

Enforces comments on filegroups in order to:

1) Submit Review
If output checkers request changes, they must add a comment (public or private) on the relevant filegroup(s)

2) Return request after independent review
If output checkers request changes, they must add a public commenton the relevant filegroup(s).
Returning a request early (before independent review is complete) does not require comments.

3) Resubmit a returned request
The researcher must add comments to any filegroup with files with requested changes.
We check all files, not just output files, because a file can change type (e.g. a withdrawn file could have had changes requested in the previous turn, so we want a comment on it). We also check for output files with Incomplete decisions as these are files that have been newly added, withdrawn and re-added, moved between groups, or updated.

Buttons are disabled, and alerts have clickable groups to add comments:
Needs a comment before review can be submitted:
Screenshot from 2025-03-06 17-20-56

Needs a public comment before request can be returned:
Screenshot from 2025-03-06 17-21-43

Returned request needs comments before it can be resubmitted:
Screenshot from 2025-03-06 17-22-21


airlock/models.py
```diff
@@ -925,6 +925,44 @@ def all_files_reviewed_by_reviewer(self, reviewer: User) -> bool:
     def submitted_reviews_count(self):
         return len(self.submitted_reviews)
 
+    def filegroups_missing_comment_by_reviewer(self, reviewer) -> set[str]:
```
evansd, line 936: Just to say that I really appreciated the clarity of this function. It's the sort of logic that's potentially a bit fiddly but the naming of things and the comments make it incredibly easy to read and see that it's doing the right thing.

airlock/models.py
```diff
@@ -973,24 +973,50 @@ def all_filegroups_commented_by_reviewer(self, reviewer: User) -> bool:
...
+        in the current turn.
         """
         if self.all_files_approved():
             return []
-        if self.submitted_reviews_count() < 2:
-            # public comments are not enforced until independent reivew is complete
-            return []
-        submitted_reviewers_this_turn = self.submitted_reviews.keys()
+
+        match self.status:
```
evansd, line 992: Fancy!

evansd (approved):
I was slightly dreading reviewing this when I saw the size of the diff! But the commit messages and the clarity of the individual functions made it all very easy to follow along.

evansd (approved):

PR #880, opened 2025-03-07
Make sure screenshots stay in sync with docs

rebkwok:
I found a couple of screenshot files that weren't being used, or which were referenced incorrectly in the docs, so this adds a test to make sure they're in sync.


tests/integration/test_screenshots_in_docs.py
```diff
@@ -0,0 +1,31 @@
+import os
+import re
+from pathlib import Path
+
+from django.conf import settings
+
+
+SCREENSHOT_RE = re.compile(r"\b[\w-]+.png\b")
```
rw251, line 8: nit - I know this is unlikely to happen, but this regex wouldn't match files like `file.thing.png` or `file(1).png`.

rw251 (dismissed):

rw251:
Nice that this already caught a few inconsistencies.

tests/integration/test_screenshots_in_docs.py
rebkwok: Hmm yeah - I was checking for an initial `/` and thought `\b` might be better, but it wouldn't get dotted filenames, and I hadn't thought about brackets in filenames. Have updated the regex now and added some tests for it.

rw251 (approved):

PR #881, opened 2025-03-07
Display a loading indicator while the table is sorting

rw251:
Looks like this:

[Screencast from 07-03-25 08:13:00.webm](https://github.com/user-attachments/assets/c6e17a38-5604-4f8b-94f1-48f9eb43beaa)


rebkwok (approved):
Looks good :+1: 
Would the same thing work for filtering as well? Filtering itself doesn't seem too bad because it does it as you type, but clearing a filter has the same problem as sorting - you click on the X and nothing happens for a second or two

rw251:
That's a good point. I hadn't spotted it because the typing is pretty quick, and the "x" seems to be browser specific - you get in chrome but not in firefox with a `<input type="search">` field. I'm sure you could do something very similar. Might be a bit trickier given the behaviour is slightly different in that you may want to debounce the filter so that the search only triggers after a small pause in the user typing. I'll add it to the original issue as this PR only fixes the sorting part, and not quite the loading part.

PR #884, opened 2025-03-07
Make the add-or-change and update modals wider

rebkwok:
Makes longer filenames look nicer in the modals

Before:
Screenshot from 2025-03-07 10-33-07
Screenshot from 2025-03-07 10-33-17

After:
Screenshot from 2025-03-07 10-32-21
Screenshot from 2025-03-07 10-32-33


inglesp (approved):
Much nicer!

PR #885, opened 2025-03-07
Make tree-view expand to display long file names and nested directories

rw251:
Looks like this:

[Screencast from 07-03-25 12:04:35.webm](https://github.com/user-attachments/assets/8ce8bbb5-6ec8-4fac-889e-e8f044f4c024)


inglesp (approved):
Good to see the extra sample data.

PR #886, opened 2025-03-07
Change wording of the request button that takes you to the workspace

### LINKED ISSUE #869: Back to workspace button on request overview page
When on the request overview page as a reviewer, there is a back to workspace button in the top right. If I'm reviewing, that's not where I came from

madwort:
* fixes #869


Providence-o (approved):

Providence-o (commented):
It might be worth removing the icon since it no longer represents the wording.

madwort:
We still have the left arrow for "Request overview", so I think it makes sense to keep it for now - arguably we could remove it in both places, but we can merge this & think about that later.

PR #887, opened 2025-03-07
Request changes and reset votes for all files in a group

### LINKED ISSUE #735: "Batch" responses to similar groups of outputs
I was reviewing a very large request today, with files grouped into various groups (i.e, one group was 45 figure files). Every file in one the group was missing the underlying data. I was able to add a comment to the group as a whole but I had to assign the "request additional info" status for each file (this process was made even more painful by the fact that I had to manually scroll down to find each file every time I assigned the review status because every time I assigned a review status the screen would refresh, the file would have to reload, and the screen would revert to the top of the list of files).

It would be very useful to assign a review status to a group of files...but possibly only sensible for the "request additional info" and "unsure". This may have already been thought of but thought I would share incase it hasn't.

rebkwok:
Fixes #735 

Output-checkers can now request changes and reset votes on all files in a file group.

Requesting changes sets the vote on any _unreviewed_ file; any file that already been approved is left as is (so is any changes-requested file, so we don't end up with superfluous audit logs).

Resetting votes resets all files, approved or changes-requested and is only visible before reviews are submitted (you can't reset a vote after submitting, but you can change it).

Modals for each button explain what you're about to do.

New buttons:
Screenshot from 2025-03-07 17-25-08

Request changes confirmation modal:
Screenshot from 2025-03-07 17-25-24

Reset votes confirmation modal:
Screenshot from 2025-03-07 17-25-31

Plus a couple of drive-by fixes for docs favicon and functional test locators


airlock/views/request.py
```diff
@@ -1099,6 +1120,134 @@ def group_comment_visibility_public(request, request_id, group):
...
+        )
+    except (exceptions.RequestPermissionDenied, exceptions.RequestReviewDenied) as exc:
+        raise PermissionDenied(str(exc))
+
+    # Track the number of files we update so we can report it to the user
+    approved_files = 0
+    changes_requested_already = 0
+    changes_requested = 0
+    errors = []
+    for output_file in filegroup.output_files:
```
bloodearnest, line 1145: What do you think about moving this logic and loop into a bll method .e.g. `group_request_changes(...)` or similar?

It feels like its slightly on the edge of too much logic for a view to me? And it would push the permissions check into the BLL, which I think is the general pattern where possible, and speaks to the first couple of lines of the big block comment (although concurrency is still an issue). 

However, needing approved/changed file counts and errors does make that all a bit awkward :thinking:

airlock/views/request.py
```diff
@@ -1099,6 +1120,134 @@ def group_comment_visibility_public(request, request_id, group):
...
+        permissions.check_user_can_currently_review_request(
+            request.user, release_request
+        )
+        permissions.check_user_can_reset_review(request.user, release_request)
+    except (exceptions.RequestPermissionDenied, exceptions.RequestReviewDenied) as exc:
+        raise PermissionDenied(str(exc))
+
+    reset = 0
+    errors = []
+    for output_file in filegroup.output_files:
```
bloodearnest, line 1215: Same question about pushing down into BLL here too, I think

bloodearnest (approved):
This is a nice UX change, cleanly done.

I've left a question about whether some logic might be worth pushing down into the BLL. If you agree with that idea, am totally happy for this to land as is, ready for demoing, and we can do a follow up PR if we want to push it down?

rebkwok:
@bloodearnest I did originally think of pushing this down to the bll (and potentially having an extra audit event for group_request_changes), but I didn't, because it felt more like repeating bll operations than something that needed to be its own thing. However, I think that was before I added the collecting counts and errors. They could be bll methods that return the various counts and errors instead, and I don't think anything much would need to change, other than the location of the code.

I'll have a look at this as a later PR

PR #900, opened 2025-03-13
Use the file uploader in screenshot tests

### LINKED ISSUE #892: Investigate CI "database table is locked" errors
E.g. https://github.com/opensafely-core/airlock/actions/runs/13771326696/job/38510379689#step:8:732

This has happened a few times recently, and I think is usually at the [same point](https://github.com/opensafely-core/airlock/actions/runs/13771326696/job/38510379689#step:8:673)

I suspect it's because we manually call bll.register_file_upload() in the test, but the file uploader is also running in the test server, and it calls bll.register_file_upload_attempt() to update the upload attempt count on the same file. I think we can avoid this by mocking the job-server urls so that all the uploads are registered by the file uploader. It might need some fiddliness to get the screenshots to happen at the right time.

rebkwok:
This point in the test has seen the occasional "db locked" error, which could possibly be due to the htmx polling interfering with the manual bll calls. Calling the run_file_uploader management command makes for simpler code in the test and more closely mimics the way the uploads work.

Tentatively maybe fixes #892

### DIFF
--- a/tests/functional/test_docs_screenshots.py
+++ b/tests/functional/test_docs_screenshots.py
@@ -1,6 +1,8 @@
 import re
+from unittest.mock import Mock
 
 import pytest
+from django.core.management import call_command
 from playwright.sync_api import expect
 
 from airlock.business_logic import bll
@@ -73,7 +75,7 @@ def move_mouse_from(page):
 
 
 def test_screenshot_from_creation_to_release(
-    page, live_server, context, release_files_stubber
+    page, live_server, context, upload_files_stubber
 ):
     author, user_dicts = get_user_data()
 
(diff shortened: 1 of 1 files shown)


tests/functional/test_docs_screenshots.py
```diff
@@ -438,20 +440,22 @@ def do_review_and_approve(username):
...
+
     take_screenshot(page, "request_approved_upload_in_progress.png")
 
-    # Progress the release request to all uploads complete and released
-    release_request = factories.refresh_release_request(release_request)
-    for relpath in all_output_relpaths[1:]:
-        bll.register_file_upload(release_request, relpath, checker_user)
-    bll.set_status(release_request, RequestStatus.RELEASED, checker_user)
+    # upload by running the file uploader for one iteration
+    call_command("run_file_uploader", run_fn=Mock(side_effect=[True, False]))
```
evansd, line 452: Nifty!

evansd (approved):

PR #902, opened 2025-03-13
Making the csv display a lot more responsive

rw251:
Currently we use the simple-datatables library to display csv files. Because this requires rendering the entire table it can become a bit laggy, particularly when sorting and searching (searching isn't too bad except as it's incremental as you type, but is not great when you clear the selection). The sort of files we display in output checking are likely to be pretty small. But even so, users have noticed the slowness, and so we recently added better visual indications that something was happening when the table was loading and sorting.

This draft PR is an attempt to display the csv files in a much more responsive way. It uses a library called clusterize.js which solves the rendering problem of large tables by only actually creating DOM nodes for a few hundred rows at a time and updating them on scrolling and sorting. This PR currently displays csv files with the old method, unless they contain the string `faster` in which case they use the new way so you can compare the behaviour side by side.  `just load-example-data` will set up `example-workspace` with `rows.csv` and `rows-faster.csv` - identical files except the latter will display with the new method. There are also `rows-LARGE.csv` and `rows-LARGE-faster.csv` which are again identical and are simply csvs 5x larger than the `rows.csv` to really demonstrate the difference.

Pros:
- This will work for csv files of millions of rows - we don't need that but it's nice that this will never hit problems
- The loading, sorting, and searching are all a lot more responsive than before

Cons:
- Relies on the clusterize.js library which hasn't been updated for a while, and in fact I needed to monkey patch a callback. However partly the reason for not being updated is that it works well and is stable (I've used in previous projects with no issues) so maybe not a problem.
- clusterize.js only deals with the "display large numbers of rows" problem and so you need to implement things like "search" and "sort" yourself. That is done in this PR but if we needed to add something like pagination it is no longer a one line change - though we could always revert to simple-datatables for paginated tables.
- The column widths change depending on the rows currently in the DOM. This PR code keeps these widths updated, and does sensible things like ensuring when you sort a column, the column you click on doesn't move. However it's possible that some people might not like it.

### DIFF
--- a/airlock/templates/file_browser/file_content/csv.html
+++ b/airlock/templates/file_browser/file_content/csv.html
@@ -29,23 +29,23 @@
       }
 
       /* Ensure the correct sort icon displays */
-      .csv-table-wrapper .clusterize-table-sorter .datatable-icon--ascending,
-      .csv-table-wrapper .clusterize-table-sorter .datatable-icon--descending,
-      .csv-table-wrapper .clusterize-table-sorter .datatable-icon--sorting{
+      .clusterize-table-wrapper .clusterize-table-sorter .datatable-icon--ascending,
+      .clusterize-table-wrapper .clusterize-table-sorter .datatable-icon--descending,
+      .clusterize-table-wrapper .clusterize-table-sorter .datatable-icon--sorting{
         display: none;
       }
-      .csv-table-wrapper .sort-ascending .datatable-icon--ascending {
+      .clusterize-table-wrapper .sort-ascending .datatable-icon--ascending {
         display: block;
       }
-      .csv-table-wrapper .sort-descending .datatable-icon--descending {
+      .clusterize-table-wrapper .sort-descending .datatable-icon--descending {
         display: block;
       }
-      .csv-table-wrapper .table-sorting .datatable-icon--sorting {
+      .clusterize-table-wrapper .table-sorting .datatable-icon--sorting {
         display: block;
       }
-      .csv-table-wrapper .sort-ascending .datatable-icon--no-sort,
-      .csv-table-wrapper .sort-descending .datatable-icon--no-sort,
-      .csv-table-wrapper .table-sorting .datatable-icon--no-sort  {
+      .clusterize-table-wrapper .sort-ascending .datatable-icon--no-sort,
+      .clusterize-table-wrapper .sort-descending .datatable-icon--no-sort,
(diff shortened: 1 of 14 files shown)


tests/functional/test_csv_viewer.py
```diff
@@ -99,12 +111,162 @@ def test_csv_renders_all_text(live_server, browser, csv_file):
...
+    )
+
+    page.goto(
+        live_server.url + workspace.get_contents_url(UrlPath("outputs/file1.csv"))
+    )
+
+    page.locator("tbody").hover()
+    for i in range(num_rows):
+        row_locator = page.get_by_text(f"value{i}", exact=True)
+        if not row_locator.is_visible():  # pragma: no cover (if we ever revert to non-virtualized tables then this is never True)
```
rebkwok, line 264: Could this use `pragma: no branch` instead? I think that would mean that it'd pass coverage while this `if` is always True, but fail if it's never True (in which case, we'd probably want to know, so we can fix the test)

rebkwok (approved):
I have a couple of comments on naming things and coverage-skipping, nothing blocking. Tried it out locally and everything looks good, and it's much nicer now the columns stay the same width.

tests/functional/test_csv_viewer.py
rw251: Ah I didn't know that - thanks for the tip.

airlock/templates/file_browser/file_content/csv.html
```diff
@@ -207,7 +207,7 @@
             </table>
           </div>
         </div>
-        {% vite_asset "assets/src/scripts/csv-datatable.js" %}
+        {% vite_asset "assets/src/scripts/clusterize-datatable.js" %}
```
rebkwok, line 210: This is fine, although it was actually the "datatable" bit that I found confunsingly named, rather than the "csv" bit. But it's clear now that it's a clusterize datatable we're talking about rather than a datatables datatable, and if we're planning at somepoint to change the directory table to use this, it makes sense that it isn't labelled with csv.

rebkwok (approved):

airlock/templates/file_browser/file_content/csv.html
rw251: Ha - ok. Yes I was making it (1) not csv-specific, and (2) explicitly clusterize. But you're right about the datatable bit, probably clusterize-table would have been better. I may implement it for the file-browser if I have time this week and can change it then.

PR #918, opened 2025-03-21
Add hypothesis tests to check that randomly generated csv files display correctly in airlock

rw251:
Currently to display a csv file in airlock we:
- read the file from disk
- parse it with `csv.reader()`
- render it to a HTML table via a Django template
- enhance the table with a client side javascript library
If there was a bug somewhere in this flow, then in theory you could have a csv file that displayed differently in airlock. This is an attempt to write tests to confirm that this will never happen.

### DIFF
--- a/tests/functional/csv_generator.py
+++ b/tests/functional/csv_generator.py
@@ -1,0 +1,123 @@
+"""
+A Hypothesis strategy for generating valid CSV files
+"""
+
+import functools
+import string
+from csv import writer
+from io import StringIO
+
+from hypothesis.strategies import (
+    composite,
+    floats,
+    integers,
+    lists,
+    sampled_from,
+    text,
+)
+
+
+def _records_to_csv(rows, lineterminator):
+    """
+    Convert the results into a csv string
+    """
+    f = StringIO()
+    w = writer(f, lineterminator=lineterminator)
+    for row in rows:
+        w.writerow(row)
+
+    return f.getvalue()
+
+
+def _escape_csv_field(field) -> str:
+    field = str(field)
+
+    # Carriage return characters (\r) become newline (\n) characters in the html
+    # This is fine, but annoying for the tests which then fail. So easier to just
+    # exclude them as a possible csv character
+    while "\r" in field:
+        field = field.replace("\r", "_")
+    return str(field)
+
+
+@composite
+def csv_row(draw, columns):
+    """
+    Strategy to produce a single csv row
+
+    Args:
+      columns: a list of hypothesis strategies, one per column
+
+    Returns:
+      tuple: the csv row
+    """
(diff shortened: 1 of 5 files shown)


tests/functional/csv_generator.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _escape_csv_field(field) -> str:
+    field = str(field)
+
+    # Carriage return characters (\r) become newline (\n) characters in the html
+    # This is fine, but annoying for the tests which then fail. So easier to just
+    # exclude them as a possible csv character
+    while "\r" in field:
+        field = field.replace("\r", "_")
```
inglesp, line 39: I think the loop is unnecessary:

```
>>> "a\rb\rc".replace("\r", "_")
'a_b_c'
```

tests/functional/test_csv_viewer.py
```diff
@@ -0,0 +1,108 @@
...
+          }
+        """,
+        csv_list,
+    )
+
+    assert len(failures) == 0, print(repr(failures))
+
+    # In theory we should tidy up by closing the context. But there is a bug which
+    # causes this test to hang when run with coverage turned on. So we check to see
+    # if coverage is on, and if not we tidy up.
```
inglesp, line 106: o_O

inglesp: Is there an issue on the playwright or coverage repo we can link to?

tests/functional/test_csv_viewer.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    page.goto(
+        live_server.url + workspace.get_contents_url(UrlPath("outputs/file1.csv"))
+    )
+
+    reader = csv.reader(StringIO(csv_file), delimiter=",")
+    csv_list = []
+    for row in reader:  # each row is a list
+        csv_list.append(row)
```
inglesp, line 63: You could do `csv_list = list(reader)`.

inglesp (approved):
Neato.

tests/functional/csv_generator.py
rw251: I was thinking that the field might contain `\\rr`, in which case the first replace would still leave `\r`.

rw251: ah - but that was when I was replacing it with the empty string. Replacing with the underscore almost certainly means it's redundant.

tests/functional/test_csv_viewer.py
rw251: I had found one for hypothesis (https://github.com/HypothesisWorks/hypothesis/issues/4052) which I'll link to.

evansd: Oh,looks like we might need to upgrade our Python version.

inglesp (approved):

PR #926, opened 2025-03-26
Give copilot user a better name in dev_users.json

inglesp:


rebkwok (approved):
I'm pretty sure this was a copy/paste error and this is what it was supposed to be in the first place :)

Issue #768, opened 2025-02-10
Merge 'workspaces' and 'requests' views


Providence-o:
Depends on the completion of:
- #756 
- #757 

Following the user feedback gathered, we will be creating a merged page showing the workspace and request views for both researchers and output checkers. See the [slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1738761781588549) for further discussion. 

The page will look like the below image, with the following key points:
- the embedded workspace and request links will retain their current functionality. 
- a more appropriate (navigation tab)name other than **requests** should be given to this new view 
- we need to test (functionally and with the users) how having zero active requests will be handled

![Image](https://github.com/user-attachments/assets/fb35f10a-65f4-4219-9d6f-81e06a9b8442)

PR #932, opened 2025-04-01
Merge request and workspace views

Providence-o:
Fixes: #768

New merged tab:
Screenshot from 2025-04-01 15-47-14


rebkwok (commented):
This was quite a lot of work! I've left a bunch of comments, but I think it mostly looks good. I think the main issue that needs addressed is the copiloted workspaces view - which might need a decision on whether requests should be shown there or not.

When I was clicking about locally I found it annoying that everything starts collapsed, and to see requests I have to open the project and each workspace. That might be worth asking users about. If you're a user with one project and only one or two workspaces, it could be really irritating to always have to expand everything, and they'd prefer that it starts open and they can collapse what they don't want to see. Users with lots of projects/workspaces might prefer everything to be collapsed (or projects open but workspaces collapsed?), but I'd guess that there's not so many of them, and they're probably internal users. Also, copiloted workspaces might be different - if one copilot has lots of copilot projects, they might prefer things to be collapsed.

airlock/templates/requests_for_workspace.html
```diff
@@ -30,25 +30,15 @@
...
-                  {{ request.created_at|date:"H:i" }}
-                </time>
-              </dd>
-            </div>
-          </div>
-        {% /list_group_rich_item %}
+        {% #airlock_list_group_rich_item title=title url=request.get_url status_text=request.status.description %}
+          <div class="group-hover:pointer-events-none ml-auto sm:flex-row -mt-4 -mb-6 -mr-5">
+            {% airlock_list_group_rich_item date=request.created_at %}
+              </div>
```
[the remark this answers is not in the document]
Providence-o: The div tag is applied to override the formatting of the date and doesn't affect the other contents of       `airlock_list_group_rich_item`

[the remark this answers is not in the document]
rebkwok: Ah, sorry, I was misled by the indentation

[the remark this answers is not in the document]
rebkwok: Odd that the linter didn't pick that up; I guess it's confused by it wrapping a component

rebkwok:
Sorry @Providence-o, had another thought on this - it's going to need some documentation updated. The docs currently tell users to go to /requests which doesn't exist anymore.

airlock/nav.py
```diff
@@ -37,18 +37,20 @@ def menu(request):
...
 
     if request.user.is_authenticated:
         if request.user.copiloted_workspaces:
             copiloted_workspaces_menu = NavItem(
-                name="Copiloted Workspaces", url_name="copiloted_workspace_index"
+                name="Copiloted Workspaces & Requests",
+                url_name="copiloted_workspace_index",
             )
-            items.insert(1, copiloted_workspaces_menu)
+            items.insert(nav_index, copiloted_workspaces_menu)
```
rebkwok, line 48: This isn't doing anything different yet.
In order to get the behaviour we want (i.e. copiloted workspaces then reviews), you need to increment `nav_index` here, so if there are copiloted workspaces, reviews will be inserted at index 2.

airlock/templates/workspaces.html
```diff
@@ -1,28 +1,52 @@
...
+                      {% #airlock_list_dropdown custom_button=button %}
+                        <div>
+                          {{ workspace.display_name }}
+                        </div>
+                      {% /airlock_list_dropdown %}
+                    {% /airlock_list_group %}
+                  </summary>
+                  <div>
+                    <a class="header__link px-6 sm:px-12" href="{{ workspace.get_url }}">Go to {{ workspace.display_name }}</a>
+                  </div>
```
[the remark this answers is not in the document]
Providence-o: That was my first thought too but after working with Thomas, we settled on this approach (he shared [this link](https://infinum.com/blog/website-navigation-dropdown-menus/) on design decisions). We could get feedback from users after merging?

Providence-o:
> Sorry @Providence-o, had another thought on this - it's going to need some documentation updated. The docs currently tell users to go to /requests which doesn't exist anymore.

The docs will be updated in a new PR

airlock/nav.py
rebkwok: If a copilot is also an output-checker, they will see both tabs (and I think most of them are). Our test copilot isn't an output-checker. You could update the `dev_users.json` to make it an output-checker for testing

airlock/nav.py
```diff
@@ -34,22 +34,23 @@ def iter_nav(items, request):
...
+        NavItem(name="Workspaces & Requests", url_name="workspace_index"),
         NavItem(name="Docs", url_name="docs_home"),
     ]
+    nav_index = 1
 
     if request.user.is_authenticated:
         if request.user.copiloted_workspaces:
             copiloted_workspaces_menu = NavItem(
-                name="Copiloted Workspaces", url_name="copiloted_workspace_index"
+                name="Copiloted Workspaces & Requests",
```
rebkwok, line 45: This is now just workspaces for Copilots, isn't it?

airlock/templates/workspaces.html
```diff
@@ -1,28 +1,65 @@
 {% extends "base.html" %}
 
-{% block metatitle %}{{ workspace_type|title }} for {% airlock_user user=request.user %} |  Airlock{% endblock metatitle %}
+{% block metatitle %}{{ workspace_type|title }} & Requests for {% airlock_user user=request.user %} |  Airlock{% endblock metatitle %}
```
rebkwok, line 3: The "& Requests" bit is only valid for non-copiloted workspaces

airlock/nav.py
Providence-o: Yes it is, thank you

airlock/nav.py
Providence-o: It works out that swapping the conditions fixes the order problem without incrementing nav_index

airlock/views/workspace.py
```diff
@@ -52,13 +52,15 @@ def workspace_index(request):
     workspaces = bll.get_workspaces_for_user(request.user)
     projects = dict(grouped_workspaces(workspaces))
     authored_requests = bll.get_requests_authored_by_user(request.user)
+    workspace_header = "Workspaces & Requests for "
```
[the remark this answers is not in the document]
rebkwok: No, I think this is the best solution

rebkwok (approved):

PR #938, opened 2025-04-09
Use a filesystem db in tests

rebkwok:
Django by default uses an in-memory database during tests. We can't set WAL mode on an in-memory database, so the functional tests that have htmx polling and also run the file uploader occasionally hit a "database table is locked" error when the polling tries to read at the same time as the file uploader writes. We can use a filesystem db instead by setting a test db name in the DATABASES settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ assets/out
 
 # Hypothesis outputs
 .hypothesis
+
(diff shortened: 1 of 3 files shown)


evansd (dismissed):
:crossed_fingers:

PR #946, opened 2025-05-06
Show a different decision status to researchers when there are conflicting reviews

### LINKED ISSUE #914: Don't show that files are conflicted when request is returned to researcher
It is possible to return a request to a researcher with conflicted files, ie when one output checker has approved and the other has requested changes.  This is potentially confusing for the researcher, and it's not explained anywhere in the docs what they should do.

Image

We should show that researcher that the output checkers have requested changes.  The output checkers should still see that the file is conflicted.

(The alternative is to insist that the approving output checker changes their vote, but this runs into problems with holidays.)

Providence-o:
Fixes #914

### DIFF
--- a/tests/functional/test_request_pages.py
+++ b/tests/functional/test_request_pages.py
@@ -466,15 +466,24 @@ def test_file_vote_buttons(
     changes_requested_selected_disabled,
     decision_tooltip,
 ):
-    checker = factories.create_api_user(
-        username="checker", workspaces=_workspace_dict(), output_checker=True
-    )
+    user_data = {
+        "author": dict(
+            username="researcher", workspaces=_workspace_dict(), output_checker=True
+        ),
+        "checker": dict(
+            username="checker", workspaces=_workspace_dict(), output_checker=True
+        ),
+    }
+
+    author = factories.create_airlock_user(**user_data["author"])
+    checkers = [
+        factories.create_airlock_user(**user_data["checker"]),
+        factories.get_default_output_checkers()[0],
+    ]
 
     release_request = factories.create_request_at_status(
         "workspace",
-        author=factories.create_airlock_user(
-            username="researcher", workspaces=_workspace_dict(), output_checker=False
-        ),
+        author=author,
         status=status,
         withdrawn_after=RequestStatus.PENDING,
         files=[
@@ -483,15 +492,12 @@ def test_file_vote_buttons(
                 path="file1.txt",
                 approved=current_vote == RequestFileVote.APPROVED,
                 changes_requested=current_vote == RequestFileVote.CHANGES_REQUESTED,
-                checkers=[
-                    factories.create_airlock_user(**checker),
-                    factories.get_default_output_checkers()[0],
-                ],
+                checkers=checkers,
             ),
         ],
     )
 
-    login_as_user(live_server, context, checker)
+    login_as_user(live_server, context, user_data["checker"])
     page.goto(live_server.url + release_request.get_url("group/file1.txt"))
 
     approve_button = page.locator("#file-approve-button")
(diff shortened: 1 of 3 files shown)


rebkwok (commented):
I think perhaps it would be simpler to leave the decision (i.e. the return value from `get_decision()` as it is; we want to show non output-checkers a different value, but the decision itself is still _really_ "conflicted". That means you don't need to pass user around to all the places that need to call `get_decision()`; the only place that it would need to change is in the view/templates.

The view/template uses `path_item.request_status.decision` to get the `decision` that gets displayed in templates - we can make this be something other than the result of get_decision. 

`path_item.request_status` is set by calling ReleaseRequest.get_request_file_status [here](https://github.com/opensafely-core/airlock/blob/main/airlock/models.py#L721), which calls `get_decision` and already has access to the current user - in the COMPLETE or AUTHOR phases, if get_decision is conflicted, and the current user is an output checker but not the author, we can make it return changes requested instead.

tests/unit/test_models.py
```diff
@@ -1061,3 +1062,71 @@ def test_coderepo_from_workspace(bll):
...
+    release_request = factories.refresh_release_request(release_request)
+
+    # check file decision displayed to output checkers
+    release_request_file = release_request.get_request_file_status(
+        UrlPath("group/path/file1.txt"), checker1
+    )
+    assert (
+        release_request_file
+        and release_request_file.decision == RequestFileDecision.CONFLICTED
+    )
```
rebkwok, line 1110: The asserts in this test shouldn't need to assert release_request_file (just `assert release_request_file.decision == ...`). If the release_request_file is None, it'd error with something like "None has no attribute 'decision'", so if the assert passes, release_request_file can't be None

tests/unit/test_models.py
```diff
@@ -1061,3 +1062,71 @@ def test_coderepo_from_workspace(bll):
...
     )
+
+
+def test_displayed_decision_for_conflicting_file_reviews(bll):
+    path = UrlPath("path/file1.txt")
+
+    author = factories.create_airlock_user(
+        username="author", workspaces=["workspace"], output_checker=True
+    )
+
```
rebkwok, line 1073: It'd be worth adding another user here, who has access to the workspace, but isn't an output checker, and checking that they also see changes_requested for a conflicted file (after return)

tests/unit/test_models.py
```diff
@@ -1061,3 +1062,71 @@ def test_coderepo_from_workspace(bll):
...
+    bll.return_request(release_request, checker2)
+    release_request = factories.refresh_release_request(release_request)
+
+    release_request_file = release_request.get_request_file_status(
+        UrlPath("group/path/file1.txt"), author
+    )
+    assert (
+        release_request_file
+        and release_request_file.decision == RequestFileDecision.CHANGES_REQUESTED
+    )
```
rebkwok, line 1132: Maybe also test that the output-checkers see CONFLICTED, as expected

tests/functional/test_request_pages.py
```diff
@@ -1302,3 +1302,75 @@ def test_request_action_required_alert(
     page.locator(".tree__file").filter(has_text="file1").click()
     expect(content_alert).to_be_visible()
     expect(content_alert_link).to_be_visible()
+
+
+def test_conflicted_file_decision(live_server, context, page, bll):
```
rebkwok, line 1307: This test is fine, but it does duplicate all of the test setup from the unit test, and the only thing it's testing that is different is the html content in the decision tooltip. We could just add an extra bit to the `test_file_vote_buttons` test in the same file, which has a test already for the conflicted case [here](https://github.com/opensafely-core/airlock/blob/main/tests/functional/test_request_pages.py#L528)

So, at the end of that `if` block, logout as the output checker, login as the author, do `decision_locator.hover()` again, and check that it shows different text to the author.

rebkwok (commented):
Just some suggestions for minor improvements to tests

tests/unit/test_models.py
Providence-o: Without asserting `release_request_file`, the test errors with "None has no attribute 'decision'"  even when release_request_file is not None because i think it also checks for instances where RequestFileStatus is none based on our method definition: 
```
def get_request_file_status(
        self, relpath: UrlPath, user: User
    ) -> RequestFileStatus | None:
```

rebkwok (approved):

PR #956, opened 2025-06-06
Management command for creating a release request for a user/workspacce

rebkwok:
Intended for occasional use only, to create a release request for a user and workspace, and automatically add all files in specified directories within the workspace.

The request is created with groups derived from the directory names.

Files that have already been added or released are skipped.

### DIFF
--- a/airlock/management/commands/create_release_request.py
+++ b/airlock/management/commands/create_release_request.py
@@ -34,8 +34,6 @@ class Command(BaseCommand):
     """
 
     def add_arguments(self, parser):
-        # In production, we want this check to run forever. Using a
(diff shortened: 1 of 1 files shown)


airlock/management/commands/create_release_request.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class Command(BaseCommand):
+    """
+    Automatically create a release request
+    """
+
+    def add_arguments(self, parser):
+        # In production, we want this check to run forever. Using a
+        # function means that we can test it on a finite number of loops.
```
madwort, line 38: not sure I understand this comment

madwort (dismissed):

PR #966, opened 2025-07-04
Amend gunicorn config

madwort:
* this port variable doesn't work, but 8000 happened to be the default


bloodearnest (approved):
we already EXPOSE 8000 in the airlock dockerfile :+1:

PR #969, opened 2025-07-08
Handle Gunicorn timeouts properly

bloodearnest:
We are seeing timeout tracebacks in the airlock logs regularly on TPP.

Due to the way timeouts work with gunicorn by default, we have no idea what url
is timing out, or any other useful information to diagnose it.

There's a known fix to this, so I've implmeneted it. The fix is simple, but the
testing for it is complex, as it involves running gunicorn servers in the
background.

This should hopefully help us diagnose the issue in airlock.

We may also want to port the fix to the controller too.


tests/functional/test_gunicorn.py
```diff
@@ -0,0 +1,104 @@
...
+        while time.time() - start_time < timeout:
+            # check if process died
+            if process.poll() is not None:
+                stdout, _ = process.communicate()
+                print(stdout)
+                raise AssertionError("gunicorn failed to start correctly")
+
+            # check it is up
+            try:
+                client.get("http://localhost/{check_url.lstrip('/')}")
```
rebkwok, line 62: Should this be an f-string?

rebkwok (approved):
One comment on a missing `f` string. Otherwise this makes sense to me (at least, the actual code does, I'm not entirely sure I've got a complete handle on the details of the tests)

tests/functional/test_gunicorn.py
bloodearnest: huh, it 100% should, and that exposes another bug, that I'm not checking the response

We probably don't care what the response code is, just that there is one!

rebkwok (approved):

PR #975, opened 2025-07-16
More gracefully handle connection timeouts with gunicorn

bloodearnest:
We discovered that chrome is eagerly connecting to guniocrn, but not
sending a request through. Normally, there would be nginx or similar
infront of gunicorn, but not in airlocks case.

When this happens, the gunicorn worker timesout, and puts and ugly
stacktrace in the logs, with nothing we can do about it.

This change implements a bit of a hack to allow us to detect this case,
and instead of exiting with an exception traceback, we can log one line
and then exit cleanly.


gunicorn.conf.py
```diff
@@ -21,11 +25,34 @@ def post_fork(server, worker):
...
+# it means we just got a connection with no data, so we exit w/o error. This
+# happens with airlock specifically as we have no proxy infront of it.
+#
+# If not, we raise RequestTimeout, so that our django stack will handle it as
+# an error and emit logs and telemetry.
 def worker_abort(worker):
+    if not _http_request:
+        worker.log.info(f"No request sent timeout, exiting (pid: {worker.pid})")
+        # do not raise SystemExit, just quit
+        os._exit(0)
```
evansd, line 54: Is there a change in behaviour in exiting 0 vs 1?

evansd (approved):
Seems like a reasonable workaround in the face of an annoying problem :shipit:

gunicorn.conf.py
bloodearnest: Yes and no :)

See the last paragraph of this commit from the initial introduction of raising RequestTimeout: https://github.com/opensafely-core/airlock/commit/b61eece0948bbda7c731d97c47eeaaa2e850c285

Basically, because we raise [at this point](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L203), then the worker actually drops out of this loop [at this point](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/sync.py#L61) (as self.worker is now False), and exits with a 0.

So, yes, it's a change from the default gunicorn behaviour, but no this commit is not a change, we already made that change in the earlier commit.

bloodearnest: It would actually be quite complex to *keep* the behaviour of exiting with a 1 when raising an exception.

Of course, we could choose to exit with ever exit code we want here. Because there's no error in this case, there's nothing to see, so I figure 0 was best?

PR #996, opened 2025-10-07
Update to PostgreSQL v17

StevenMaude:
The Digital Ocean production database upgraded to v17.

We're not sure why yet (opensafely-core/job-server#5339).

The proposed fix is to upgrade local setups to match production (opensafely-core/job-server#5338).

### DIFF
--- a/job-server/docker-compose.yml
+++ b/job-server/docker-compose.yml
@@ -2,7 +2,7 @@
 # configuring the production build
 services:
   db:
-    image: "postgres:16"
+    image: "postgres:17"
     environment:
       POSTGRES_USER: user
       POSTGRES_PASSWORD: pass


StevenMaude:
question: One thing I noticed this time: there is a postgresql-client install within the Airlock Docker setup: https://github.com/opensafely-core/airlock/blob/2f0c41a73ff3337cb1c7290c39028fac0abc5373/docker/dependencies.txt#L3

However, [that was added a while ago](https://github.com/opensafely-core/airlock/commit/9d9a1b4f182f6abb67c96c9159e5e52800715ecc), when there was a PostgreSQL server in the Airlock Docker setup. That no longer exists, so I'm not sure whether the client's used now?

Does that need upgrading too? If so, I may have to add the PostgreSQL PPA there to get the v17 client.

bloodearnest (approved):
To be clear, the only reason we have postgres in airlock at is our local testing setup that spins up a local job-server.

I don't know why postgresql-client is installed, I suspect a c/p error - we don't use psql anywhere AFAICT

StevenMaude:
Thanks for the clarification, Simon.

I'll merge this then and open an issue to discuss the client being still installed (as that's not really in my scope to fix here).

PR #1015, opened 2025-11-04
Don't error in management command if nothing to do

rebkwok:
Report but don't error in the `create_release_request ` management command if all files have already been released, or if a release request is already under review.

Return the ID, completed status and info message so we can add it to the telemetry in the runjobs job.

The `runjobs weekly` command will run as a systemd timer; while it's going to run for jobs that we expect to be _run_ (i.e. run via job-server) weekly, we're going to run it daily (see https://github.com/opensafely-core/backend-server/pull/312) so that it picks up any new release requests that need to be created soon after the job(s) have run. These changes mean that we won't raise errors if a release request is already in progress (i.e. we're running the command after a request has been created, but before output checkers have reviewed and released) or if all files are already released (i.e. the job hasn't been re-run yet).

### DIFF
--- a/airlock/jobs/daily/create_regular_release_requests.py
+++ b/airlock/jobs/daily/create_regular_release_requests.py
@@ -2,7 +2,7 @@ import json
 import logging
 
 from django.conf import settings
-from django_extensions.management.jobs import WeeklyJob
+from django_extensions.management.jobs import DailyJob
 from opentelemetry import trace
 
 from airlock import actions
@@ -16,7 +16,7 @@ CONFIG_PATH = settings.WORK_DIR / "regular_release_requests.json"
 class ConfigValidationError(Exception): ...
 
 
-class Job(WeeklyJob):
+class Job(DailyJob):
     help = "Create release requests for regularly run jobs."
 
     def execute(self):
--- a/tests/integration/jobs/test_daily_create_regular_release_requests.py
+++ b/tests/integration/jobs/test_daily_create_regular_release_requests.py
@@ -6,7 +6,7 @@ import pytest
 from django.core.management import call_command
 
 from airlock.enums import RequestStatus
-from airlock.jobs.weekly.create_regular_release_requests import (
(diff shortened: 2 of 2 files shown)


airlock/management/commands/create_release_request.py
```diff
@@ -65,6 +66,19 @@ def add_arguments(self, parser):
     def handle(self, username, workspace_name, **options):
         user = User.objects.get(user_id=username)
         request = bll.get_or_create_current_request(workspace_name, user)
+        if request.is_under_review():
+            self.stdout.write(
+                f"A release request for workspace '{workspace_name}' is already under review for user '{username}'"
+            )
+            # Management commands expect to write their output to std out, so return it as a (json)string
+            return json.dumps(
```
bloodearnest: What happens with this return value if we run this command manually from the cli?

Does it get printed? Or just swallowed? Does it exit with 0 or 1?

airlock/management/commands/create_release_request.py
```diff
@@ -65,6 +66,19 @@ def add_arguments(self, parser):
     def handle(self, username, workspace_name, **options):
         user = User.objects.get(user_id=username)
         request = bll.get_or_create_current_request(workspace_name, user)
+        if request.is_under_review():
+            self.stdout.write(
+                f"A release request for workspace '{workspace_name}' is already under review for user '{username}'"
+            )
+            # Management commands expect to write their output to std out, so return it as a (json)string
```
bloodearnest: Huh, I thought you had to print it yourself, and returning anything was only for internal use and testing.

bloodearnest (approved):
Left a question about the semantics of returning json from a management command. 

It feels a bit inconsistent that we are running a "weekly" task daily. Should this actually be a daily task?

Conceptually, to my mind, the submission of the job in job-server is the weekly task, but the picking up of any new files is daily?

airlock/management/commands/create_release_request.py
bloodearnest: TIL :D

PR #1057, opened 2026-01-12
Automated release request config validation

rebkwok:
- Add expected type for the new `supporting_files`  in automated release requests (currently it fails because it assumes they're the default str type)
- Explicitly map all the possible config keys to expected types so this doesn't happen again (and add a test to ensure we don't forget)
- Make config validation runnable as a separate management command - lets us check the config is OK when we add to it or update it, without having to run the actual release request job for real.

### DIFF
--- a/airlock/jobs/daily/create_regular_release_requests.py
+++ b/airlock/jobs/daily/create_regular_release_requests.py
@@ -95,7 +95,9 @@ def validate_config_data(release_request_data):
     )
 
     if missing_required_keys:
-        errors.append("keys missing in config: " + ",".join(missing_required_keys))
+        errors.append(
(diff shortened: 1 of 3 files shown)


tests/integration/jobs/test_daily_create_regular_release_requests.py
```diff
@@ -50,12 +53,28 @@ def test_get_config_data_file_does_not_exist():
         assert get_config_data() == []
 
 
+def test_config_types():
+    signature = inspect.signature(create_release_request_action)
+    config_keys = set(signature.parameters) - {"kwargs"}
+    assert config_keys == set(CONFIG_TYPES)
```
evansd: I love tests like this :+1:

PR #1063, opened 2026-01-20
Retrieve author from manifest file for automated release requests

rebkwok:
This [job-runner PR](https://github.com/opensafely-core/job-runner/pull/1310) added the user who created a job request to the manifest data for each output (I have checked in Airlock for recently run jobs and verified that the user is indeed included in the manifest). So we can now retrieve it and use it as the author for the automated release requests.

If no username is passed to create_release_request, we look up the most recent output in the manifest.json and use that user to create the release request.
    
Note that in order to run the job, the user must have access to the workspace in job-server. However, they may not have logged into Airlock before, and/or their workspace permissions in Airlock may not be up to date. We therefore call the authorise endpoint on job-server, and create/update the user with the data received. This doesn't authenticate the user in Airlock, so if they were to try to access Airlock in the browser, they'd still need to log in.
    
We may have multiple release requests to be created for the same user (e.g. it's likely that the same  user would rerun all 3 database reports at the same time). If we're attempting to create another release request with the same user within 60s, we skip calling the API again. (60s is probably overly conservative, but automated release requests typically only involve a small number of files and take a very short time to run).

### DIFF
--- a/airlock/actions/create_release_request.py
+++ b/airlock/actions/create_release_request.py
@@ -1,13 +1,12 @@
 import logging
-import time
 from dataclasses import dataclass, field
 
 from airlock import permissions, policies
 from airlock.business_logic import bll
 from airlock.enums import RequestFileType, WorkspaceFileStatus
-from airlock.exceptions import FileNotFound, ManifestFileError
+from airlock.exceptions import APIException, FileNotFound, ManifestFileError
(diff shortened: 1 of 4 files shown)


bloodearnest (dismissed):
LGTM. One possible suggestion

PR #1066, opened 2026-01-20
Save session every request.

### LINKED ISSUE #1065: Change session expiry to match intent/docs
The goal was that users would be logged out after 8 weeks of inactivity.

However, users have report that they are logged every 8 weeks.

Turns out we needed to enable that, its not the default. 

https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1751270418475739

Setting [SESSION_SAVE_EVERY_REQUEST=True](https://docs.djangoproject.com/en/5.2/ref/settings/#session-save-every-request) should do it.

bloodearnest:
This will update the timestamp, causing the session to only expire after
8 weeks of *inactivity*, not just every 8 weeks, which is what we
originally intended and documented.

Fixes #1065


evansd:
Ah yeah, I'd forgotten about all the audit writes.

PR #1076, opened 2026-01-28
Prevent wrapping in clusterize table

### LINKED ISSUE #1060: Unexpected formatting of a CSV file in a release request
@andrewscolm reports that a large amount of white space is displayed between rows of a CSV file (`df_pf_consulatations_breakdown_updated.csv`) in [this release request][2]. For more information, see [this Slack thread][1].

[1]: https://bennettoxford.slack.com/archives/C09039RP0SH/p1768474938521359
[2]: https://github.com/ebmdatalab/opensafely-output-review/issues/1009

rebkwok:
Fixes #1060 
See also [thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1769440482315959)

Clusterize.js needs rows to be of the same/approximately the same height 
Wrapping can break the cluster updating - [this issue ](https://github.com/NeXTs/Clusterize.js/issues/62)seems related and the suggested fix is to prevent wrapping 

The wrapping behaviour in the table seems to be a bit odd and inconsistent; I can reproduce the issue with a dummy data version of reported problematic file, but it only appears on the first load of the file (and the rows display as not wrapped). On a refresh, the text in the rows is wrapped and the scrolling behaviour works fine. BUT in L4 I did see the problematic scrolling behaviour consistently with this file.  In tests, I can't seem to get a table that displayed with wrapped text, even with very long cell content.

Update: thanks to Richard, I've also added a regression test. The exact cause is still undetermined, but this should prevent us inadvertently re-introducing it in future if someone suggests that wide table columns should wrap

### DIFF
--- a/tests/functional/test_csv_viewer.py
+++ b/tests/functional/test_csv_viewer.py
@@ -310,3 +310,53 @@ def test_csv_summarize(live_server, page, context):
 
     modal_button.click()
     expect(summary_table).to_be_visible()
+
+
+def test_csv_with_wrapping(live_server, page, context):
+    # This is a regression test for some odd behaviour seen with CSV row values that wrapped.
+    # With some (not easy to determine) combination of additional columns, clusterize seems to
+    # calculate the expected height of a cluster incorrectly.
+    # Visually, the table would appear unwrapped. When scrolled past the 200 row cluster, the
+    # table content would disappear and we see a big vertical gap.
+    # Possibly clusterize gets the height and then the table re-renders wrapped/unwrapped so
+    # the calculations are now wrong.
+    # The fix was to prevent wrapping in the table altogether. This test fails consistently
+    # before the fix and passes consistently afterwards. Note that in headed mode, it also
+    # fails, but if given a long enough slowmo (100ms at the time of writing) it passes.
+    # https://github.com/opensafely-core/airlock/pull/1076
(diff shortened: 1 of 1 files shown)


tests/functional/test_csv_viewer.py
```diff
@@ -310,3 +310,53 @@ def test_csv_summarize(live_server, page, context):
...
+def test_csv_with_wrapping(live_server, page, context):
+    # This is a regression test for some odd behaviour seen with CSV row values that wrapped.
+    # With some (not easy to determine) combination of additional columns, clusterize seems to
+    # calculate the expected height of a cluster incorrectly.
+    # Visually, the table would appear unwrapped. When scrolled past the 200 row cluster, the
+    # table content would disappear and we see a big vertical gap.
+    # Possibly clusterize gets the height and then the table re-renders wrapped/unwrapped so
+    # the calculations are now wrong.
+    # The fix was to prevent wrapping in the table altogether. This test fails consistently
+    # before the fix and passes consistently afterwards. Note that in headed mode, it also
```
alarthast: Interestingly, I ran this in headed mode and have seen it pass with and without the fix (with no slowmo). Running a couple of times more and I observed it failing so it's quite flaky 😕 

I wonder if it's best to just say that this should be run in headless mode only?

PR #1109, opened 2026-03-05
Switch to dependabot for version updates

rebkwok:
This switches to using the same dependabot configuration as we've been using in BennettBot for a few weeks now.

This should prevent the failing dependabot security updates which conflict with the explicit exclude-newer timestamps that we were putting in the pyproject.toml/uv.lock. Hopefully it will also prevent the duplicate security alerts (currently we have 8 security alerts for 2 django vulnerabilities - 2 each for pip/uv for each vulnerability. Renaming and excluding the requirements.uvmirror file should stop it running pip; not sure why we have the second duplicate, but possibly due to the first one being unable to resolve due to the timestamps?)

1) Use dependabot with uv for both security updates and version updates. We use dependency groups to get one PR per dev/prod group, the 7 day cooldown for version updates, and we specify `dependency-type: "all"`, which should get us transitive dependencies.

2) Keep the uvmirror requirements file (renaming it to avoid dependabot running pip updates on it).

3) Add a uvmirror-update workflow. This checks for changes to the uvmirror file and creates a commit to add them if it finds any. Dependabot doesn't have write permission, so for dependabot PRs (which is where we expect the uvmirror to be out of date), it fails and links to the page to manually rerun the workflow.

The uvmirror file has been useful with the update-dependencies workflow, for reviewing PRs and being able to more easily see the diff and where transitive dependencies are coming from. It may not be required when we have dependabot PRs which pull in the release notes, and it may just become annoying that every dependabot PR fails and requires re-running the uvmirror-update and waiting for the tests to run again.  I think it's worth keeping it for a few weeks to see.

### DIFF
--- a/justfile
+++ b/justfile
@@ -121,7 +121,7 @@ uvmirror file="requirements.uvmirror":
 # This is the default input command to update-dependencies action
 # https://github.com/bennettoxford/update-dependencies-action
 
-# Bump the timestamp cutoff to midnight UTC 7 days ago and upgrade all dependencies
+# Upgrade all dependencies
 update-dependencies: upgrade-all
 
 format *args:


alarthast:
Oh, I forgot to say:

> not sure why we have the second duplicate

I think this is just because we have two `django`s for Python 3.12+ and not. The two pairs of security updates target different `django` versions.

PR #1149, opened 2026-04-13
Allow CSV table columns to be hidden and unhidden

rebkwok:
Add hideable columns to the CSV clusterize table viewer

Users can now hide columns in the CSV viewer by clicking the hide icon
(a `-` in a circle) in each column header.

Hidden columns are shown as buttons above the table, with a show icon
that is the opposite of the hide icon (i.e. a `+` in a circle) . Clicking
a button unhides just that column. Hidden column buttons are shown in
the same order as they appear in the table (and reordered if necessary
when a new one is hidden)

Hidden column state is tracked in a module-level Set, so it persists
across sorts and searches

Demo: 
[Screencast from 2026-04-11 15-30-20.webm](https://github.com/user-attachments/assets/9e51591f-9595-40b1-a14d-13e106f810e6)

Closes #1147


evansd (approved):
🤖 🚀

PR #1187, opened 2026-05-12
Fix slow all-workspaces page by using lightweight WorkspaceListing objects

rebkwok:
The all-workspaces page was unusably slow in production, because there are so many workspaces, and it was calling get_workspace() for every directory, triggering manifest.json reads, hashing, file tree construction, and DB queries for every workspace, most of which it did't need to do.

Add a WorkspaceListing dataclass that just has the get_url()/display_name() interface the template needs, and refactor get_all_workspaces() to only scan the WORKSPACE_DIR and check for directories and for the existence of a manifest file.

Also simplify _build_workspace_list() to always record ManifestFileError telemetry unconditionally, since it is now only ever called with a user's own workspaces and copiloted workspaces rather than arbitrary filesystem directories.

### DIFF
--- a/airlock/business_logic.py
+++ b/airlock/business_logic.py
@@ -30,6 +30,7 @@ from airlock.models import (
     ReleaseRequest,
     RequestFile,
     Workspace,
+    WorkspaceListing,
 )
 from airlock.notifications import send_notification_event
 from airlock.types import UrlPath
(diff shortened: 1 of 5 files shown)


evansd (approved):
Nice 👍

PR #1196, opened 2026-05-19
Preserve tree scroll position after file review actions

### LINKED ISSUE #906: Maintain file tree position when submitting a file review
If there are lots of files in a request, you have to scroll down the file tree to get to some of the outputs. Once you make a review decision for a file, you're sent back to the top of the tree and have to scroll again to get to the next file. This is a little bit annoying. 

An alternative workflow would be to provide github inspired next/prev buttons to allow me to quickly move to the next file. You might only want to allow this navigation within file groups.

Providence-o:
Closes: #906 

When reviewing files in a large request, clicking approve/request changes (and undoing these actions) would redirect back to the same file but reset the file tree scroll position to the top. 

This fix saves the tree scroll bar position before the form submits and restores it after the page reloads.

There is no test added for this change because the scroll bar position restoration is dependent on a hard-coded delay, making it prone to flakiness.

### DIFF
--- a/airlock/static/assets/file_browser/index.js
+++ b/airlock/static/assets/file_browser/index.js
@@ -191,20 +191,47 @@ document.body.addEventListener("submit", (event) => {
   if (tree) sessionStorage.setItem("treeScrollTop", tree.scrollTop);
 });
 
-// Restore scroll position on page load
-document.addEventListener("DOMContentLoaded", () => {
+// Restore scroll position on page load.
+//
+// scrollTop is a property on any scrollable element. It's the number of
+// pixels the content is scrolled down from the top. We can read it and
+// write it - but because it's a position, the fact that we set it doesn't
+// mean that it definitely has the value we set.
+//
+// The valid range is [0, scrollHeight − clientHeight]. If the layout hasn't
+// fully settled when we assign, then scrollHeight can momentarily be <=
+// clientHeight, so it's actually set to 0 and nothing ever re-applies it.
+// Reading scrollTop back after assignment tells us whether the value stuck.
+//
+// So: assign, read back, and if it didn't stick, retry on the next animation
+// frame. Give up after ~30 frames (~0.5s at 60fps) so we don't loop forever if
+// the saved position is genuinely unreachable (e.g. the destination page has
+// less tree content than the source).
+//
+// Note on very large trees: the budget is 30 animation frames, not 500ms of
+// wall-clock time. If layout is expensive (tens of thousands of files) and the
+// frame rate drops, rAF fires less often, so we still get 30 real chances to
+// re-apply. The number of settle events we're waiting for (initial layout,
+// fonts, the resizer's 100ms debounce) is fixed regardless of tree size, so
+// this should scale.
+window.addEventListener("load", () => {
   const saved = sessionStorage.getItem("treeScrollTop");
+  if (!saved) return;
+  sessionStorage.removeItem("treeScrollTop");
+  // We only need to care about scroll positions if we're on a page with a tree
   const treeContainer = document.getElementById("tree-container");
+  if (!treeContainer) return;
 
-
-  if (saved) {
-    sessionStorage.removeItem("treeScrollTop");
-    // Wait for browser scroll restoration to complete before applying saved position.
-    // The browser may reset scrollTop after DOMContentLoaded, so we delay slightly.
-    setTimeout(() => {
-      treeContainer.scrollTop = parseInt(saved, 10);
+  const target = parseInt(saved, 10);
+  let attempts = 30;
+  const restoreScrollPosition = () => {
+    treeContainer.scrollTop = target;
+    if (treeContainer.scrollTop === target || attempts-- <= 0) {
       document.getElementById("scroll-restore-style")?.remove();
-    }, 125);
-  }
+      return;
+    }
+    requestAnimationFrame(restoreScrollPosition);
+  };
+  restoreScrollPosition();
 });
 
--- a/airlock/static/assets/file_browser/index.js
+++ b/airlock/static/assets/file_browser/index.js
@@ -1,3 +1,11 @@
+// Only disable browser scroll restoration when there is a position to restore
+if (sessionStorage.getItem("treeScrollTop")) {
+  if ('scrollRestoration' in history) {
+    history.scrollRestoration = 'manual';
+    document.head.insertAdjacentHTML("beforeend", "<style id='scroll-restore-style'>#tree-container{visibility:hidden}</style>");
+  }
+}
+
 // keep the selected class up to date in the tree on the client side
 function setTreeSelection(tree, event) {
   // target here is the hx-get link that has been clicked on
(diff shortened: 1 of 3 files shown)


airlock/static/assets/file_browser/index.js
```diff
@@ -1,3 +1,7 @@
+// Only disable browser scroll restoration when there is a position to restore
+if (sessionStorage.getItem("treeScrollTop") && 'scrollRestoration' in history) {
+  history.scrollRestoration = 'manual';
+}
```
rebkwok, line 9: There's currently a slightly disconcerting jump in the tree when you click an approve/reject etc button - the tree appears scrolled to the top and then jumps to the remembered position. If we modify this slightly we can hide the tree here, and then unhide it in the restore code below. That means that there's a brief period where the tree isn't visible at all as the page loads, but I think it's a better UX than the jumping.

```
if (sessionStorage.getItem("treeScrollTop")) {
  if ('scrollRestoration' in history) {
    history.scrollRestoration = 'manual';
    document.head.insertAdjacentHTML("beforeend", "<style id='scroll-restore-style'>#tree-container{visibility:hidden}</style>");
  }
}
```

airlock/static/assets/file_browser/index.js
```diff
@@ -155,4 +159,32 @@ if (document.readyState !== "loading") {
...
+    sessionStorage.removeItem("treeScrollTop");
+    treeContainer.style.visibility = "hidden";
+    
+    // Wait for browser scroll restoration to complete before applying saved position.
+    // The browser may reset scrollTop after DOMContentLoaded, so we delay slightly.
+    setTimeout(() => {
+      treeContainer.scrollTop = parseInt(saved);
+      treeContainer.style.visibility = "visible";
+    }, 125);
+  }
```
rebkwok: To unhide the tree now, we'd change this to:
```
if (saved) {
    sessionStorage.removeItem("treeScrollTop");
    // Wait for browser scroll restoration to complete before applying saved position.
    // The browser may reset scrollTop after DOMContentLoaded, so we delay slightly.
    setTimeout(() => {
      treeContainer.scrollTop = parseInt(saved, 10);
      document.getElementById("scroll-restore-style")?.remove();
    }, 125);
  }
```

airlock/static/assets/file_browser/index.js
```diff
@@ -155,4 +159,32 @@ if (document.readyState !== "loading") {
...
+});
+
+// Restore scroll position on page load
+document.addEventListener("DOMContentLoaded", () => {
+  const saved = sessionStorage.getItem("treeScrollTop");
+  const treeContainer = document.getElementById("tree-container");
+
+  if (saved) {
+    sessionStorage.removeItem("treeScrollTop");
+    treeContainer.style.visibility = "hidden";
```
rebkwok: I think this is intended to do what I've suggested, hiding the tree altogether first and then unhiding it after the scroll restoration. But when I test it locally, I see the tree at the top position flash briefly.  
Claude says this is because DOMContentLoaded fires too late. 

> Here's the timing:
> 
>   When the browser encounters <script src="index.js">, it pauses HTML parsing and
>   fetches the file over the network. During that fetch, the browser may paint 
>   whatever it has already parsed — including #tree-container, visible at the wrong
>   scroll position. That's the first paint, and it's done before the script even runs.
>      
>   By the time the script executes, fires its DOMContentLoaded listener, and sets
>   visibility: hidden, the user has already seen that first frame.
> 
>   The style injection at the top of index.js works because it runs synchronously
>   during script execution — before the browser gets its next paint opportunity. The
>   browser batches paints; it won't paint again until the current JS task finishes. So
>    the tree is hidden before the next frame, and stays hidden through the scroll
>   restoration until the timeout reveals it at the right position.

airlock/static/assets/file_browser/index.js
```diff
@@ -155,4 +159,32 @@ if (document.readyState !== "loading") {
...
-document.body.addEventListener("clusterize-table-updated", renderCheckboxStatus);
\ No newline at end of file
+document.body.addEventListener("clusterize-table-updated", renderCheckboxStatus);
+
+// Save scroll position before approve/request_changes form submits
+document.body.addEventListener("submit", (event) => {
+  const form = event.target.closest("form");
+  if (!form) return;
+  const action = form.getAttribute("action") || "";
+  if (action.includes("/approve/") || action.includes("/request_changes/") || action.includes("/reset_review/")) {
```
rebkwok: Do we _need_ to check for the action? Could we cjust check that there's a tree present and assume we want to do this? We've already confirmed that there's a form that could be submitted. i.e.

```
const tree = document.getElementById("tree-container");                    
  if (tree) sessionStorage.setItem("treeScrollTop", tree.scrollTop); 
```

That has the added bonus of maintaining the scroll position if you're in a workspace with lots of groups, and you submit the context/controls or comments forms, and if we ever change/add actions for that form, it'll continue to work.

rebkwok (commented):
I apologise for getting Claude to review this (but it is JS, in my defence).  I think the approach is good, some suggestions that I think make the UX slightly better and the code less brittle.

I think this should be testable with a playwright test (I've had some success with getting Claude to write those)

rebkwok:
Since I've got Claude looking at this anyway, here's a playwright test it wrote for me:

```

@pytest.mark.parametrize(
    "button_id",
    ["#file-approve-button", "#file-request-changes-button"],
)
def test_tree_scroll_preserved_after_file_review_action(
    live_server, context, page, button_id
):
    login_as_user(
        live_server,
        context,
        user_dict=factories.create_api_user(
            username="checker", output_checker=True
        ),
    )
    release_request = factories.create_request_at_status(
        "workspace",
        status=RequestStatus.SUBMITTED,
        files=[
            factories.request_file(group="group", path=f"file{i:02d}.txt")
            for i in range(30)
        ],
    )

    page.goto(live_server.url + release_request.get_url("group/file00.txt"))

    assert page.evaluate("document.getElementById('tree-container').scrollTop") == 0
    page.evaluate("document.getElementById('tree-container').scrollTop = 200")
    scroll_before = page.evaluate("document.getElementById('tree-container').scrollTop")
    assert scroll_before > 0

    page.locator(button_id).click()
    page.wait_for_load_state("load")
    # Tree is hidden during scroll restoration; wait for it to become visible
    expect(page.locator("#tree-container")).to_be_visible()

    scroll_after = page.evaluate("document.getElementById('tree-container').scrollTop")
    assert scroll_after == scroll_before
```

tests/functional/test_request_pages.py
```diff
@@ -1398,3 +1398,57 @@ def test_request_action_required_alert(
...
+    tree_container = page.locator("#tree-container")
+
+    # The resizer script (assets/src/scripts/resizer.js) shrinks
+    # #tree-container to fit the viewport asynchronously via a debounced
+    # ResizeObserver; until it runs the container isn't overflow-scrollable
+    # and scroll_into_view_if_needed would scroll the window instead.
+    page.wait_for_function(
+        "() => { const c = document.getElementById('tree-container');"
+        "  return c && c.scrollHeight > c.clientHeight; }"
+    )
```
alarthast, line 1434: I trust that the comment is true and have focused only on the `page.wait_for_function` code.

alarthast (dismissed):
I am happy for this to be merged as is - just made some suggestions in https://github.com/opensafely-core/airlock/pull/1232 since it took me a while to understand what each bit did and I thought it might be worth peppering in some more comments and extracting a function that does the restoration.

alarthast (approved):

PR #1206, opened 2026-05-28
Invoke packages as a python module in just recipes

rebkwok:
`uv sync` writes wrapper scripts to .venv/bin for packages that declare a console_scripts entrypoint, which
includes a shebang pointing at the absolute path to the venv. If the absolute venv path doesn't exist (e.g. because uv sync was run on a host machine and the code is mounted into a vagrant VM), `uv run mkdocs/pytest/ruff` etc fails. Avoid that by invoking packages as python scripts with `uv run python -m` to bypass the bin scripts altogether.

### DIFF
--- a/justfile
+++ b/justfile
@@ -134,18 +134,18 @@ uvmirror file="requirements.uvmirror":
     uv export --format requirements-txt --frozen --no-hashes --all-groups --all-extras > {{ file }}
 
 format *args:
-    uv run ruff format --diff --quiet {{ args }} .
-    uv run djhtml --tabwidth 2 --check airlock/
+    uv run python -m ruff format --diff --quiet {{ args }} .
+    uv run python -m djhtml --tabwidth 2 --check airlock/
 
 lint *args:
-    uv run ruff check {{ args }} .
+    uv run python -m ruff check {{ args }} .
 
 lint-actions:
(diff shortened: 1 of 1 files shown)


alarthast (approved):
I've tested that the commands work, and I think I didn't notice much speed difference, so I'm happy to approve.

I however couldn't manage to reproduce the original issue that this fixes - I imagine this to be due to different VM setups.

Just to note however that Claude claims that adding `UV_PROJECT_ENVIRONMENT=.venv`  to the `.env` file might get relative path resolution working (I couldn't really verify this claim either), so it might be worth giving that a brief go before merging this.

rebkwok:
Thanks! I'll give that a try, would definitely be a simpler fix

rebkwok:
OK, setting UV_PROJECT_ENVIRONMENT to `.venv`  doesn't work (either in .env or in the justfile). uv understands the relative paths but still writes absolute paths to the bin scripts, so if I `just clean` and `just docs-build` on my host machine and then try to `just docs-build` in my vagrant VM, where the path is `/projects/airlock/.venv`, it errors with

```
error: Failed to spawn: `mkdocs`
  Caused by: No such file or directory (os error 2)
```

I can fix it in the VM, by running `just clean` first, but then it points at the vm's venv, so the same thing happens on the host.

Issue #961, opened 2025-06-18
Allow previously released files to be included as supporting files


rebkwok:
Previously released files (within the same workspace) cannot be added to a new release request, by design. However, we prevent them being added as both output files _and_ supporting files.

In some cases, a user may want to create a release request for an output that is entirely dependent on data in a previous release request - e.g. the request mentioned in this [slack thread](https://bennettoxford.slack.com/archives/C01UJLWNZJT/p1750235678099699?thread_ts=1750235262.265119&cid=C01UJLWNZJT) is for a report based on previously released data, which the user is unable to add to the request as supporting files, so they would need to find/link to the previous release request, or the output-checker needs to go looking for the relevant files in the workspace.

Given that supporting files are by definition not released, we should consider allowing previously released files to be added as supporting files only.

suzannehamilton:
This is a UI change as well as a permissions change, but we could probably grey out the "select for release" boxes.

rebkwok:
@suzannehamilton I know it was months ago, but which "select for release" box are you referring to? Airlock doesn't have any select-for-release boxes that I can think of.  Users add files to a release request as either "output" or "supporting" file type. If the release request is approved and released, **all** the output files in the request are released. Supporting files are never released.

suzannehamilton:
@rebkwok  Given the time of the comment (10am on a Wednesday), that looks like a note that I added during triage, so it might be a misinterpretation of what someone said on the call. So ignore it if it doesn't seem helpful!

PR #1207, opened 2026-05-28
Fix select / deselect all behaviour when released files are present

alarthast:
When implementing #961, we conducted a poll for the desired behaviour of the select-all checkbox when there are previously released files in a directory. Based on the poll results, we then decided to leave previously released files unselected upon clicking select all.

When the checkbox logic was implemented in ec69771, the deselect all behaviour became counter-intuitive, in that the checkbox does nothing when all unreleased files are selected but not all released files are. This odd behaviour was reported by a user as a bug and subsequently discussed in [this Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1779878129565449). This PR re-introduces more user-friendly deselection logic covered by parametrised functional tests.

### DIFF
--- a/tests/functional/test_workspace_pages.py
+++ b/tests/functional/test_workspace_pages.py
@@ -690,6 +690,97 @@ def test_select_all(live_server, page, context):
     expect(page.locator("input.selectall")).not_to_be_checked()
 
 
+@pytest.mark.parametrize(
+    # `initial_state` and `final_state` here are tuples of booleans representing
+    # the file checkboxes where True=checked, False=unchecked, in the order of
+    # (released_file_1, released_file_2, unreleased_file_1, unreleased_file_2)
+    "initial_state,final_state",
+    [
+        # Not all unreleased checked - select all unreleased
+        ((False, False, False, False), (False, False, True, True)),
+        ((True, False, False, False), (True, False, True, True)),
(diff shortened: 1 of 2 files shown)


tests/functional/test_workspace_pages.py
```diff
@@ -690,6 +690,94 @@ def test_select_all(live_server, page, context):
     expect(page.locator("input.selectall")).not_to_be_checked()
 
 
+@pytest.mark.parametrize(
+    "initial_state,final_state",
+    [
+        # Not all unreleased checked - select all unreleased
```
rebkwok, line 699: I'm finding these test params hard to understand - an extra comment to explain what the Trues/Falses relate to might be enough to help (i.e. the first 2 are the checked state of released files, the second 2 are unreleased).

rebkwok (dismissed):

tests/functional/test_workspace_pages.py
alarthast: Thank you - that makes sense; have added a comment here:

https://github.com/opensafely-core/airlock/blob/d427ffa64054011692f0bcbb1d77c8928ba7379e/tests/functional/test_workspace_pages.py#L694-L696

rebkwok (approved):

PR #1208, opened 2026-05-29
Rename DATABASE_URL env var to AIRLOCK_DATABASE_URL

rebkwok:
For consistency with othe env var settings, which are prefixed with AIRLOCK_, and to avoid confusion with the ehrql db env variables in the backend.

### DIFF
--- a/airlock/settings.py
+++ b/airlock/settings.py
@@ -164,7 +164,7 @@ WSGI_APPLICATION = "airlock.wsgi.application"
 
 # Database
 # https://docs.djangoproject.com/en/5.0/ref/settings/#databases
-database_url = os.environ.get("DATABASE_URL")
+database_url = os.environ.get("AIRLOCK_DATABASE_URL")
 if not database_url:
     # This is our current production config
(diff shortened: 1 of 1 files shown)


suzannehamilton (approved):
For my own understanding, is this one of the env vars that we set manually in a file on the server? I can't see any values or placeholders in [backend-server](https://github.com/opensafely-core/backend-server/tree/565bfc71f4ae3230261730eee0c65bb3eec1b9af/config).

rebkwok:
@suzannehamilton Airlock currently doesn't actually use a `DATABASE_URL`, so there's no setting in backend server. Dave added the minimum changes we needed in order to test out using RDS, which essentially meant adding [this else block](https://github.com/opensafely-core/airlock/blob/main/airlock/settings.py#L211) in the settings. If there's no environment variable, it falls back to the current sqlite config. 

In my RDS test setup, I'm adding the RDS DATABASE_URL to the `/home/opensafely/config/02_secrets.env`

PR #1210, opened 2026-05-29
Allow out of date action outputs to be shown/hidden in the workspace view

### LINKED ISSUE #1106: Add a way for users to toggle visibility of out of date outputs and actions in a workspace
A workspace currently shows the outputs produced by the last run of each action. 

Outputs can go "out of date" by:
1)  re-running an action that produces outputs at different filepaths from the previous run
2) re-running an action which fails and produces only a subset of outputs, or no outputs
3) re-naming or removing an action in a project.yaml

1 & 2 marks the output as `out_of_date_output` in the manifest.json, and it is not displayed in Airlock
3 marks the output as `out_of_date_action` and it is displayed in Airlock, but with a warning message.

Users may want to hide outputs from old actions that they no longer care about. They may want to see outputs that were produced by previous runs of an action, but not by the last run (especially in the case of failed jobs).

It would be useful to provide some sort of toggle to show/hide out of date outputs and actions. It's probably also useful to be able to toggle these two types of out-of-date output independently.

The Workspace model builds a list of [outputs to display from the manifest.json](https://github.com/opensafely-core/airlock/blob/main/airlock/models.py#L276). It currently excludes outputs where `out_of_date_output` is true in the manifest.json. The `Workspace.from_directory()` [method](https://github.com/opensafely-core/airlock/blob/main/airlock/models.py#L180) which is called in the view code to build the Workspace could take some additional arguments to tell it whether to include/exclude the out of date outputs.

rebkwok:
Outputs marked `out_of_date_action` in `metadata/manifest.json` are produced by actions that no longer exist in the project. Previously these were always shown in the workspace file tree, styled with a greyed-out warning icon. For workspaces with lots of out of date files this clutters up the workspace and makes it slow to render.

Closes #1106 

This PR hides them by default and adds a notice strip at the top of the file tree with a `show`/`hide` link to reveal them. Visibility state is remembered per-workspace in the Django session, so toggling one workspace doesn't affect others and the choice persists while navigating.

- `Workspace.from_directory` grows an `include_out_of_date_action_outputs` flag. When `False`, the manifest parse skips those outputs entirely. A separate `out_of_date_action_count` property counts them directly from the manifest so the notice strip can show the count even when they are hidden.
- `workspace_view` reads a per-workspace boolean from `request.session` (key `show_out_of_date_action_outputs`) to determine whether to show/hide out of date files
- A new `workspace_toggle_out_of_date_action` endpoint flips the session value and re-renders the tree and content panel
- To preserve manually-expanded tree folders across the swap, the toggle `POST` sends expanded path info via `hx-vals`, 

Note:
- we don't include the toggle for the release request tree as out-of-date action info isn't relevant there
- we don't do anything different for out of date outputs (i.e. previous outputs from a current action). Those remain excluded entirely. We _might_ want to include them and allow them to be (separately) toggled as well, but there's a wider discussion to be had around that, as they can be confusing to users.

User feedback on the UI [here](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1779874748306969); there was a preference for the wording of the link option, but not the highlighting when the files were hidden, so the final implementation is:

Out of date files hidden:
Screenshot from 2026-05-29 11-06-39

Out of date files shown:
Screenshot from 2026-05-29 11-06-56


tests/integration/views/test_workspace.py
```diff
@@ -647,13 +656,156 @@ def test_workspace_view_out_of_date_file(airlock_client):
...
 
+def test_workspace_view_out_of_date_action_hidden_by_default(airlock_client):
+    airlock_client.login(output_checker=True)
+    factories.write_workspace_file("workspace", "file.txt", "test")
+    factories.write_workspace_file("workspace", "current.txt", "current")
+    workspace = factories.create_workspace("workspace")
+    manifest = workspace.manifest
+    manifest["outputs"]["file.txt"]["out_of_date_action"] = True
+    (workspace.root() / workspace.manifest_path()).write_text(json.dumps(manifest))
+    factories.refresh_workspace("workspace")
```
alarthast:
```suggestion
```

I think `get_workspace` is being called by the GET request beneath this line, so this `refresh_workspace` call would be redundant.

tests/integration/views/test_workspace.py
```diff
@@ -647,13 +656,156 @@ def test_workspace_view_out_of_date_file(airlock_client):
...
+
+
+def test_workspace_view_out_of_date_action_visible_when_session_on(airlock_client):
+    airlock_client.login(output_checker=True)
+    factories.write_workspace_file("workspace", "file.txt", "test")
+    workspace = factories.create_workspace("workspace")
+    manifest = workspace.manifest
+    manifest["outputs"]["file.txt"]["out_of_date_action"] = True
+    (workspace.root() / workspace.manifest_path()).write_text(json.dumps(manifest))
+    factories.refresh_workspace("workspace")
```
alarthast:
```suggestion
```

Same as above

airlock/templates/file_browser/tree.html
```diff
@@ -1,7 +1,7 @@
 {% for child in path.children %}
   <li class="tree tree__item">
     {% if child.is_directory %}
-      <details class="tree__folder group" {% if child.expanded %}open{% endif %}>
+      <details class="tree__folder group" {% if child.expanded %}open{% endif %} data-path="{{ child.relpath }}">
```
alarthast, line 4: I'm not sure I can locate where `data-path` is being used - is there a special use for it?

alarthast: Oh, I think I get it - it's `dataset.path` in the JS (https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes)

docs/how-tos/view-workspace-files.md
```diff
@@ -36,3 +36,18 @@ From the file view, the `More` dropdown also allows you to [view the file in alt
...
 
 ![More dropdown](../screenshots/more_dropdown_el.png)
+
+## Outputs from out-of-date actions
+
+Some outputs may have been produced by an action that no longer exists in the `project.yaml`.
+These are hidden from the file tree by default to keep your workspace uncluttered.
+A notice at the top of the tree shows how many such outputs exist.
+
+![Out-of-date outputs hidden](../screenshots/workspace_out_of_date_hidden.png)
```
alarthast: Perhaps this is lingering from a previous wording? 
Since it is alt-text and hence user-facing, it might be good to use the clumsy but unambiguous wording.
I think there is probably no need to change the screenshot file name for now as it is internal and devs know what we mean here. 

```suggestion
![Outputs from out-of-date actions hidden](../screenshots/workspace_out_of_date_hidden.png)
```

docs/how-tos/view-workspace-files.md
```diff
@@ -36,3 +36,18 @@ From the file view, the `More` dropdown also allows you to [view the file in alt
...
+
+Click **show** to reveal them. Out-of-date outputs appear with a warning icon and
+greyed-out styling to distinguish them from current outputs.
+
+![Out-of-date outputs shown](../screenshots/workspace_out_of_date_shown.png)
```
alarthast:
```suggestion
![Outputs from out-of-date actions shown](../screenshots/workspace_out_of_date_shown.png)
```

docs/how-tos/view-workspace-files.md
```diff
@@ -36,3 +36,18 @@ From the file view, the `More` dropdown also allows you to [view the file in alt
...
+
+## Outputs from out-of-date actions
+
+Some outputs may have been produced by an action that no longer exists in the `project.yaml`.
+These are hidden from the file tree by default to keep your workspace uncluttered.
+A notice at the top of the tree shows how many such outputs exist.
+
+![Out-of-date outputs hidden](../screenshots/workspace_out_of_date_hidden.png)
+
+Click **show** to reveal them. Out-of-date outputs appear with a warning icon and
```
alarthast:
```suggestion
Click **show** to reveal them. Outputs from out-of-date actions appear with a warning icon and
```

alarthast (dismissed):

alarthast (approved):

PR #1211, opened 2026-06-01
Parallelise tests

rebkwok:
This adds pytest-xdist to the dev requirements and makes one bugfix and very minimal test changes to run the tests in parallel.

Locally (with 12 physical CPU cores), `just test-all` runs in ~1m 45s with `-n auto` and in ~6m with `-n 0`. In CI the improvement is less dramatic because we only have 2 CPUs available in the GHA workflow, but it's still running the full CI workflow in about 50% of its usual time (<5m vs ~8m)

Bugfix:
airlock/lib/git.py was importing the airlock settings module directly, instead of doing it the django way `from django.conf import settings`. This meant that the test fixture we have that's supposed to ensure isolated file storage for each test wasn't actually working for GIT_REPO_DIR, which didn't matter for serial tests, but does for parallel ones.

Test fixes:
Changed the way a setting is mocked in the screenshot tests to avoid template cache conflicts (again, only shows up in parallel tests ). Plus a couple of drive-by fixes to remove some redundant waits.

### DIFF
--- a/airlock/lib/git.py
+++ b/airlock/lib/git.py
@@ -14,7 +14,8 @@ from urllib.parse import urlparse, urlunparse
 # from jobrunner import config
 # from jobrunner.lib.string_utils import project_name_from_url
 # from jobrunner.lib.subprocess_utils import subprocess_run
-from airlock import settings as config
+from django.conf import settings as config
+
 from airlock.types import UrlPath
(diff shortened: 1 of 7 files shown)


evansd (approved):
🚀

PR #1218, opened 2026-06-10
Move `build-docs` jobs into main CI workflow

alarthast:
Instead of having the docs deploy as a separate workflow that runs when CI on main passes, have it as a job on the CI workflow that only runs on `main` that `needs` all the other jobs to succeed before it runs.

This removes the use of the `workflow_run` trigger and resolves a `zizmor` complaint.


.github/workflows/main.yml
```diff
@@ -176,3 +176,21 @@ jobs:
            echo "${{ secrets.GITHUB_TOKEN }}" | docker login "$REGISTRY" -u "${GITHUB_ACTOR}" --password-stdin
            docker tag "$IMAGE_NAME" "$PUBLIC_IMAGE_NAME:latest"
            docker push "$PUBLIC_IMAGE_NAME:latest"
+
+  build-docs:
+    runs-on: ubuntu-latest
+    needs: [check, test, docker-test-dev, docker-test-prod, docker-publish]
```
alarthast, line 182: Note: I've included `docker-publish` here to preserve the original behaviour of only running this when everything in the main workflow has succeeded.

alarthast: (Oh, maybe it can only `need` docker-publish since that itself needs the others. Oops.)
```suggestion
    needs: [docker-publish]
```

Issue #1204, opened 2026-05-27
Timeouts when adding lots of files to a request


rebkwok:
We've seen a few 504 errors such as [this one](https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/result/4Yj7QDrdyLL/trace/HZ4NR3ewggN?fields[]=s_name&fields[]=s_serviceName&span=b93fe7286057fdd0)

The user (Em) has reported timeouts when trying to add lots of files to a request, and they've had to add them in batches of 15 or so.

Locally, I can add lots of files to a request, but it does take a long time (no timeout, but it's not running with gunicorn).

We could limit the number of files that can be added at a time; annoying, and we'd need to do something about the select-all option that implies you can add everything in a directory. Maybe we can hand this off to the file-uploader process to do in the background?

suzannehamilton:
We think this might be because Airlock copies and hashes the files, rather than DB performance, but we should check. Output files must be small because there's a file size limit, but supporting files might be larger.

PR #1224, opened 2026-06-17
Improve workspace performance

rebkwok:
Some workspaces are very large (10s of 1000s of files of more) and are slow to load. We've seen 504 timeouts more frequently lately on workspace views.

Locally, a workspace with ~10K outputs and a 5MB manifest file was taking ~190ms to load (i.e. to do `bll.get_workspace()`)whether or not you actually rendered the file browser. Every file browser navigation does that again, even for htmx requests.

 It turns out to be mostly related to work we did upfront and then often didn't use. We parsed the whole manifest JSON on every load even for views (e.g. workspace index views) that only need the workspace's name and project metadata. We also built the workspace's path parent-child map (used for the tree) every time, even for views that don't need the tree.

This PR makes `Workspace.manifest` lazy: we read and store the raw text only initially. `workspace_child_map` and `workspace_files` become lazy too: built from a stored list of valid paths the first time anything actually needs the tree. 

One of the most expensive things turns out to be constructing the file path maps using UrlPaths. We now do the same thing with plain strings internally, and only convert them to `UrlPath`s in the file-browser api when we need to render them.

Performance measured on a locally generated large workspace with a 5.3MB manifest, 3000 log files in the metadata dir, and ~10,400 outputs across 24 folders, n=10:

  | Path | Main | Branch | Speedup (avg) | Speedup (min) |
  |---|---:|---:|---:|---:|
  | Index path (`bll.get_workspace`, no tree access) | avg 334.6ms, min 279.7ms | avg 45.4ms, min 32.5ms| **7.4×** | **8.6×** |
  | View path (`bll.get_workspace` + tree access) | avg 338.4ms, min 287.9ms | avg 55.2ms, min 46.5ms |**6.1×** | **6.2×** |

(Note: I also tried creating a WorkspaceMetadata db model and storing the manifest and filepaths there as a cache, only invalidating when the manifest_hash changed, but the db read actually made things very marginally worse, and all the improvements are in the cached properties and UrlPath -> str changes. The one thing it did gain is the ability to store project and workspace metadata that we currently only get from the user's api_data, which means it can be shown to output-checkers and readonly users who don't have explicit workspace access)

Probably doesn't entire sort #1204, but should make it less frequent; adding files has to do quite a lot of work reading, copying and writing to the DB anyway, but at least now it will need to spend less time loading the workspace again when it posts.

### DIFF
--- a/airlock/models.py
+++ b/airlock/models.py
@@ -204,7 +204,7 @@ class Workspace:
         return dict(json.loads(self.manifest_text))
 
     @cached_property
-    def _tree_map(self) -> tuple[dict[UrlPath, set[UrlPath]], set[UrlPath]]:
+    def _tree_map(self) -> tuple[dict[str, set[str]], set[str]]:
         """Build (workspace_child_map, workspace_files) on first access.
 
         This is the most expensive thing for large workspaces, and is only
@@ -216,11 +216,11 @@ class Workspace:
         )
 
     @property
-    def workspace_child_map(self) -> dict[UrlPath, set[UrlPath]]:
+    def workspace_child_map(self) -> dict[str, set[str]]:
         return self._tree_map[0]
 
     @property
-    def workspace_files(self) -> set[UrlPath]:
+    def workspace_files(self) -> set[str]:
         return self._tree_map[1]
 
     @classmethod
@@ -297,24 +297,27 @@ class Workspace:
         We are only really interested in file paths - those include any parent
         directories we need for the tree for free.
         """
-        root = settings.WORKSPACE_DIR / name
-        paths = set()
+        # Note: these are strings rather than UrlPaths to match get_workspace_child_map
+        # which is much more efficiently built with strings and only converted to
+        # UrlPath in the filebrowser when needed
+        root = str(settings.WORKSPACE_DIR / name) + "/"
+        paths: set[str] = set()
 
         def scan(current: str) -> int:
             children = 0
 
             for entry in os.scandir(current):
                 children += 1
-                path = UrlPath(entry.path).relative_to(root)
-
                 if entry.is_dir():
                     scan(entry.path)
                 else:
-                    paths.add(path)
+                    # Strip the workspace-root prefix for a workspace-relative
+                    # path.
+                    paths.add(entry.path[len(root) :])
 
             return children
 
-        scan(str(root / "metadata"))
+        scan(root + "metadata")
 
         return paths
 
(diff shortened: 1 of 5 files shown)


airlock/models.py
```diff
@@ -195,6 +195,12 @@ class Workspace:
     manifest_hash: str
     out_of_date_action_count: int
 
+    @cached_property
+    def manifest(self) -> dict[str, Any]:
+        """Lazily parsed manifest dict. The workspace-index hot path doesn't
+        touch this and so avoids the JSON parse entirely."""
+        return dict(json.loads(self.manifest_text))
```
evansd: Why do we need `dict()` here?

airlock/models.py
```diff
@@ -212,19 +218,17 @@ def from_directory(
         if not manifest_path.exists():
             raise exceptions.ManifestFileError(f"{manifest_path} does not exist")
 
+        # Read once, then both parse and hash from the same bytes
+        data = manifest_path.read_bytes()
+        manifest_hash = hashlib.sha256(data).hexdigest()
+        manifest_text = data.decode()
         try:
-            manifest = json.loads(manifest_path.read_text())
+            manifest = json.loads(manifest_text)
```
evansd: Don't we now end up parsing the manifest twice: once here to validate it and then again in the `manifest()` property? I guess given the performance numbers this can't be having too much of an impact but it still feels a bit weird if the whole aim was to do this lazily.

evansd (commented):
This looks like a really significant improvement. Thanks!

I want a give a bit more time to make sure I understand the final commit but just dropping in a couple of small questions I had initially.

evansd (dismissed):
OK, I'm happy I understand what this is doing now. Looks like a really good improvement. Would still be interested in the answers to my earlier questions, but neither are blockers.

airlock/models.py
rebkwok: It was just to appease mypy

airlock/models.py
rebkwok: Hmm. I think this might be a leftover from a previous implementation. I'll have a look

airlock/models.py
evansd: Ah I see, it's because `json.loads()` can return stuff other than dicts.

You might also be able to keep it happy with an assert e.g.
```py
manifest = json.loads(self.manifest_text)
assert isinstance(manifest, dict)
return manifest
```

The `dict()` call is going to construct a new dictionary object which we don't need (though it will be so fast I doubt we'll notice) but if we ever did get a value which wasn't a dict the errors might be less obvious.

rebkwok: It's gone now anyway

evansd (approved):

PR #1227, opened 2026-07-08
Use persist-credentials:true for uvmirror-update workflow

rebkwok:
We've discussed that the security risk is acceptable for this, and persist-credentials is required in order to commit the uvmirror changes.

Note that we only run the update-uvmirror job (where actions/checkout is used with persist-credentials:true) at all if the previous check-uvmirror job (which uses actions/checkout with persist-credentials:false) finds changes to the uvmirror file.


alarthast (approved):
Thank you for the note in the commit message - I found it helpful 😄

PR #1233, opened 2026-07-17
Show a loading spinner while tree scroll position is being restored

rebkwok:
We hide the tree until the scroll position can be restored, with no feedback to the user. Now we show a spinner while it's restoring, similar to the spinner pattern used in login.html.


alarthast (approved):
Ran locally and checked that it works as expected. My JS is limited but I think the code makes sense as well.

PR #1234, opened 2026-07-17
Fit images to iframe using a new ImageRenderer

### LINKED ISSUE #1183: Displayed images don't fit within the viewer
Most images are larger than the viewer, so you have to scroll to view to whole image. I end up having to open them in a new tab instead. Could they be resized so they fit in the viewer, and then the view in new tab could be used if this is too small?

rebkwok:
Adds an ImageRenderer for jpg/jpeg/png files.

The new renderer converts the image to base64 data URI in an HTML template with max-w-full / max-h-full CSS so they display contained within the iframe. Previously, images were served as raw FileResponse and rendered at their natural pixel size and large images overflowed.

SVGs are left as raw FileResponse since they scale naturally via viewBox.

Closes #1183


alarthast (approved):
It's the first time I am coming across data URIs - I am happy that the change is consistent with https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data.

Checked that the FT passes with the change
Image
and fails without the change
Image
.

